Issue #2, opened 2022-11-23
Set up CI


slifty:
We will want to run tests + linting on PR and also appropriate deployment hooks on merge to main.

(I don't think this can be taken on until #1 is finished.)

PR #7, opened 2023-02-24
Initial set up

### LINKED ISSUE #1: Set up this repository with development environment
🎉 🎉  Let's get this party started! 🎉 🎉 

(I'm trying to capture some of the thoughts from a kickoff conversation with @reefdog here!)

This will be a TypeScript project; the frontend framework is TBD but may likely be Svelte.

This issue is intended to capture the need for:

1. Setting up the dev environment with default documentation and tooling (e.g. running the sveltekit app installer).
2. Making modifications to things like linting rules, code coverage expectations, etc based on this specific project.
3. Documenting the intended code structure.

jasonaowen:
Set up the initial React application, and set up CI.

I am using Create React App (aka `react-scripts`), which has its advantages and disadvantages. I am fine to eject if needed, but for now I think the convenience is worthwhile.

Still to do: configure ESLint with the rules we want, and figure out our deployment plan; the former will require some changes to the template code, and I wanted to get CI going first, and the latter will require a conversation with the team.

Resolves #1 Set up this repository with development environment
Issue #2 Set up CI / CD


.github/workflows/lint.yml
```diff
@@ -0,0 +1,16 @@
...
+    branches:
+      - main
+jobs:
+  run_lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
```
reefdog, line 14: Rather than explicitly stating our Node version here, and risking drift from our other tooling, what do you think about using a `.node-version` file? And then:

```suggestion
          node-version-file: ".node-version"
```

.github/workflows/test.yml
```diff
@@ -0,0 +1,16 @@
...
+    branches:
+      - main
+jobs:
+  run_tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 18
```
reefdog, line 14: Same comment as above, re: `.node-version` file.

reefdog:
@jasonaowen Besides doing a basic code scan, could you put some testing instructions in the PR description so I know what to expect while evaluating this? (I can infer, but it's good practice.)

I assume at least `npm i`, `npm test`, and `npm start`… anything else?

reefdog (approved):
This installs and runs just as expected, and the commits were sensibly organized with useful descriptions, thank you!

I left a non-blocking suggestion about the use of a `.node-version` file in the repo, that I leave up to your judgment. No re-review necessary whether you choose to take the suggestion or not.

.github/workflows/lint.yml
slifty: Oh wow this is a neat trick!  Side note: fully agree with use of `.node-version` since that becomes an artifact that others can use (e.g. they can symlink it to an .nvmrc).

jasonaowen:
> @jasonaowen Besides my doing a basic code scan, could you put some testing instructions in the PR description so I know what to expect while evaluating this? (I can infer, but it's good practice.)
> 
> I assume at least `npm i`, `npm test`, and `npm start`… anything else?

Yes, of course, thank you for the reminder - always better to be explicit, and I'll be more conscious of this in future PRs!

I'd say the things documented in the README, which also include `npm run build`, and perhaps the ability to load the resulting built application (perhaps by way of [serve](https://www.npmjs.com/package/serve), or `python3 -m http.server`, or whatever ad-hoc dev server is convenient for you).

PR #24, opened 2023-03-01
Configure linter and logging

### LINKED ISSUE #1: Set up this repository with development environment

### LINKED ISSUE #2: Set up CI

jasonaowen:
Adopt Airbnb's ESLint configuration. This includes both their [base rules (including React rules)](https://www.npmjs.com/package/eslint-config-airbnb) and their [TypeScript rules](https://www.npmjs.com/package/eslint-config-airbnb-typescript). Also use the full [suite of recommended rules from the typescript-eslint plugin](https://typescript-eslint.io/linting/configs).

Add and configure [Pino](https://www.npmjs.com/package/pino) to log messages to the browser console - `INFO` and above during development, `ERROR` and above in production.

Clean up CI a bit, and fix several linter errors that show up with the stricter configuration.

See the individual commits for more detail.

To test, make sure that the appropriate logging happens during dev and production. In dev (`npm run start`), you should see the [performance metrics in the console](https://create-react-app.dev/docs/measuring-performance/); in prod (`npm run build` and load the resulting static assets), you should not. To test the error logging, you might try renaming the `root` HTML element in `public/index.html`.

Resolves #1 Set up this repository with development environment
Resolves #2 Set up CI

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -10,5 +10,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version-file: ".node-version"
-      - run: npm ci
-      - run: npm run lint
+      - name: Install dependencies
+        run: npm ci
+      - name: Run ESLint
+        run: npm run lint
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -2,8 +2,6 @@ name: Lint
 on:
   pull_request:
   push:
-    branches:
-      - main
 jobs:
   run_lint:
     runs-on: ubuntu-latest
--- a/src/logger.ts
+++ b/src/logger.ts
@@ -1,0 +1,17 @@
+import pino from 'pino';
+import type { Logger } from 'pino';
+
+const level = process.env.REACT_APP_LOG_LEVEL
+  ?? (process.env.NODE_ENV === 'production' ? 'error' : 'info');
+
+const logger = pino({
+  name: 'data-viewer',
+  level,
+  browser: {
+    asObject: true,
+  },
+});
+
+const getLogger = (source: string): Logger => logger.child({ source });
+export { getLogger };
+export type { Logger };
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,5 +10,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version-file: ".node-version"
-      - run: npm ci
-      - run: npm run build
+      - name: Install dependencies
+        run: npm ci
+      - name: Build
+        run: npm run build
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -13,4 +13,6 @@ jobs:
       - name: Install dependencies
         run: npm ci
       - name: Build
+        env:
+          DISABLE_ESLINT_PLUGIN: true
         run: npm run build
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -10,5 +10,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version-file: ".node-version"
-      - run: npm ci
-      - run: npm test
+      - name: Install dependencies
+        run: npm ci
+      - name: Run tests
+        run: npm test
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,6 +20,8 @@
         "@types/node": "^16.18.12",
         "@types/react": "^18.0.28",
         "@types/react-dom": "^18.0.11",
+        "eslint-config-airbnb": "^19.0.4",
+        "eslint-config-airbnb-typescript": "^17.0.0",
         "react-scripts": "5.0.1",
         "typescript": "^4.9.5"
       }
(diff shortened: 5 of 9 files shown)


src/logger.ts
```diff
@@ -0,0 +1,20 @@
+import pino from 'pino';
+import type { Logger } from 'pino';
+
+const level = process.env.REACT_APP_LOG_LEVEL
+  ?? process.env.NODE_ENV === 'production' ? 'error' : 'debug';
```
bickelj, line 5: Since node won't be in production, I assume this refers to "level in a build artifact intended for production" versus "level in a build artifact for intended for development".

Does `REACT_APP_LOG_LEVEL` drive log levels for dependencies too, or only the code in our data viewer app? If it drives more, perhaps it would be better to start with a  `DATA_VIEWER_LOG_LEVEL` if it exists and fall back to `REACT_APP_LOG_LEVEL`. That way we could amplify (or mute) our app and its dependencies more easily.

Another minor note, the description says "INFO and above during development, ERROR and above in production." but the setting here appears different.

One more minor consideration, if the default level is ERROR in production, that only gives two levels to work with by default in production. It might make sense to default to WARN or INFO in production to allow one or two more levels.

src/logger.ts
```diff
@@ -0,0 +1,20 @@
...
+
+const logger = pino({
+  name: 'data-viewer',
+  level,
+  browser: {
+    asObject: true,
+  },
+  transport: {
+    target: 'pino-pretty',
```
bickelj, line 14: On the server side, we would do `node ... | npx pino-pretty` but I suppose that's not available in the browser, therefore it makes sense to configure this transport straightaway, yes? I cannot quickly tell if this is what the pino folks recommend.

bickelj (commented):
This looks great. I was able to clone, run all the `npm run` commands successfully (although I did not fully `eject` because I do not know what that means).

One potential issue is some old dependencies found when I ran `npm ci`:
> 6 high severity vulnerabilities

I had some minor comments as well, take them or leave them. Overall, great, but others should probably take a look too.

.github/workflows/lint.yml
```diff
@@ -2,8 +2,6 @@ name: Lint
 on:
   pull_request:
   push:
```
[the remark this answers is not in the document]
jasonaowen: Yeah, the challenge is ensuring coverage with third-party PRs... I believe `push` covers everything in *this* repo, and it's handy to have the ability to get CI on branches before opening a PR, but that wouldn't cover a PR from a GitHub-forked repo. Maybe that's being overly optimistic, though?

.github/workflows/lint.yml
```diff
@@ -12,5 +10,7 @@ jobs:
       - uses: actions/setup-node@v3
         with:
           node-version-file: ".node-version"
```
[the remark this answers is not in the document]
jasonaowen: Credit to @reefdog for this suggestion in [the previous pull request](https://github.com/PhilanthropyDataCommons/data-viewer/pull/7#discussion_r1117785046)!

> I prefer exact versions in the codebase for maximum reproducibility and traceability.

I completely agree when it comes to dependencies, but for Node.js versions I tend to come down on only pinning to a major version: they have a frequent release cadence, and have (at least in my experience) done an excellent job of following semver - I don't think I've ever had a broken build due to a new Node.js version!

src/logger.ts
jasonaowen: Ah, there are some React conventions here! The docs are really good: [Adding Custom Environment Variables](https://create-react-app.dev/docs/adding-custom-environment-variables/).

> Since node won't be in production, I assume this refers to "level in a build artifact intended for production" versus "level in a build artifact for intended for development".

Yep, exactly! The relevant bit about `NODE_ENV`:

> There is also a built-in environment variable called `NODE_ENV`. You can read it from `process.env.NODE_ENV`. When you run `npm start`, it is always equal to `'development'`, when you run `npm test` it is always equal to `'test'`, and when you run `npm run build` to make a production bundle, it is always equal to `'production'`.

So we can depend on that being set appropriately!

> Does `REACT_APP_LOG_LEVEL` drive log levels for dependencies too, or only the code in our data viewer app?

Right now it's just for our app; if and when I add dependencies that have configurable logging (which is relatively rare on the front-end, as you kind of allude to later), I'll try to wire them up to use the same configuration.

I'm not super familiar with Pino, and probably would not have reached for it but for its use in the back-end; I would imagine there's some degree of filtering available, but for now my plan is to just deal with that if and when we get there. 

> ... perhaps it would be better to start with a `DATA_VIEWER_LOG_LEVEL`

The other React-ism bit is that in order for a custom build-time environment variable to make it into the code, it must begin with `REACT_APP_` - this is a security measure against accidentally exposing environment variables from the build machine.

Honestly, I think this is primarily going to be helpful for development, and front-end logs are unlikely to ever be looked at or sent elsewhere.

> Another minor note, the description says "INFO and above during development, ERROR and above in production." but the setting here appears different.

Oops, yes, good catch! Thank you, I'll fix that.

> One more minor consideration, if the default level is ERROR in production, that only gives two levels to work with by default in production. It might make sense to default to WARN or INFO in production to allow one or two more levels.

You might have stronger opinions than I do about log levels! My impression is that `WARN` is for recoverable errors and deprecations and such, and that end users won't want or need to see that - is that right? Remember, this is getting logged to the browser console, so the only time folks will see this is if they open up the browser dev tools, which I hope they don't feel a need to do on a regular basis. If and when we wire up to eg DataDog or New Relic or Sentry or whatever, we would definitely consider what level we want to send at that time!

src/logger.ts
jasonaowen: Oops, excellent catch! I did not mean to commit this. :flushed: 

Unfortunately, `pino-pretty` does not work in the browser. I'll remove this - thanks!

.github/workflows/lint.yml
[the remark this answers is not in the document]
jasonaowen: Aha, I looked into this more closely, and I think it's worthwhile to have both!

The [`push` trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push) will build the commit in the branch, as you would expect. However, the [`pull_request` trigger](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request) will build against a new, transient merge commit between the branch and the target branch - `main`, in the example of this pull request. That distinction isn't very important on low-traffic repos, but can be very helpful when multiple people are working in parallel!

src/logger.ts
jasonaowen: Fixed!

src/logger.ts
jasonaowen: There was also a bug on this line :sweat: which should be fixed now.

Try:

`REACT_APP_LOG_LEVEL=error npm start`, which should not show you the web-vitals log messages, or `REACT_APP_LOG_LEVEL=trace npm run build` and loading the built artifacts, which should!

jasonaowen:
> This looks great. I was able to clone, run all the `npm run` commands successfully (although I did not fully `eject` because I do not know what that means).

Thanks for testing! [Ejecting](https://create-react-app.dev/docs/available-scripts/#npm-run-eject) drops the react-scripts dependency and inlines all the things it was taking care of for us. You can try it if you want; just `git reset --hard && npm ci` after! :smiley: 

> One potential issue is some old dependencies found when I ran `npm ci`:
> 
> > 6 high severity vulnerabilities

Yeah, it's not ideal, but I also don't think it's a problem. I addressed this in the commit message for 9a118b2fd5e036ef811d422d7c31df1008a70b06, which you might find interesting.

> I had some minor comments as well, take them or leave them. Overall, great, but others should probably take a look too.

Thanks for the review, @bickelj!

.github/workflows/lint.yml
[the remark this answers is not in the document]
bickelj: Right you are! https://github.com/PhilanthropyDataCommons/data-viewer/actions/runs/4308420520/jobs/7514684247#step:2:137 shows the transient merge. Very useful, indeed.

.github/workflows/lint.yml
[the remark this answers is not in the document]
bickelj: I would say node is still a dependency here while I acknowledge it is a more fundamental dependency rather than what we typically call dependencies. A key difference between this project and the `service` project is that it is only a build-time dependency and not a runtime dependency. That difference plus deference to the people doing the bulk of the work in a repo leads me to not object to the change here while I would still object to undoing the specificity in the `service` project.

src/logger.ts
bickelj: Thanks for the consideration and explanations, it all makes sense to me. I tend to prefer verbosity (plus grep) while most folks I think prefer concision. It does come down to preference for most of these decisions. It's more important that the levels are used consistently in the code of a given package than which levels to set by default.

bickelj: By the way, I found `pino` relatively familiar and straightforward to set up in the `service` project but if it is a pain or unfamiliar here I would not object to going with what works for you.

Issue #12, opened 2023-02-28
Create basic interface components


reefdog:
Given the UI mockups created in #5 and the Storybook installation added in #27, I would like to deliver a basic library of components.

- Navbar
- Buttons
- Panels of varying widths (full, ⅔, ⅓, sidebar)

Storybook allows us to demonstrate combinations of these elements in stories, so there may be some lightweight examples of that, but I'm not going overboard on that. The real implementation of them in combination will come when making Real Pages.

reefdog:
I don't think it makes sense to keep this issue open any longer, as we've moved on to specific pages (e.g. #14, #19). So, I'm going to go ahead and close it. I may continue to tag commits to it as I create new generic components.

PR #25, opened 2023-03-04
Add logo and placeholder text

jasonaowen:
Add the PDC logo in a new `<Header>` component, along with some placeholder text to more strongly link this as-yet still fairly generic React app. This gives me a chance to test our continuous deployment setup, starts to lay some visual foundations for the application, and will give me a spot to display the current user's info.

Screenshot of Data Viewer app

I like using the spinning React logo as a modern "under construction" logo. 🚧 It won't be around for much longer, so.

I'm going to mark this as a draft because it's currently using a PNG logo, whereas we want an SVG logo. I believe it should be otherwise ready to review. As always, see the individual commits for more detail.

Issue https://github.com/PhilanthropyDataCommons/data-viewer/issues/9 Set up CD
Issue https://github.com/PhilanthropyDataCommons/data-viewer/issues/12 Implement basic interface elements


src/App.test.tsx
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
-import App from './App';
+import { App } from './App';
```
reefdog, line 3: No opinion on default vs. named imports necessarily, but I don't quite understand the stated reason (so they can't be renamed). Can't one just as easily do `import { App as AnythingElse } from './App'`?

reefdog (approved):
I think this is good enough to merge! This is a great test of the CD and some basic lint improvements.

The SVG logo replacement can come in a future change, as it will probably take a while to get it from the designer. We shouldn't store out image assets at the root of `/src`, but I recognize this was following the stub pattern. The real style/asset organization can come while delivering #12.

src/App.test.tsx
jasonaowen: They can, but that syntax is much more explicit, and protects against typos:

```ts
// thing.ts
export default Thing;
export { Thing };

// file 2
import Typo from './thing'; // works! 😬
import { Typo } from './thing'; // import error 🚫
import { Thing as Typo } from './thing'; // intentional renaming, and searching for `Thing` will find this use
```

Also, another reason I ought to have called out is that we want to be consistent with the back-end repo: https://github.com/PhilanthropyDataCommons/service/blob/bc44723f7d5d34b4eb04c5b0a60a71b8b024ada8/.eslintrc.json#L17-L18

reefdog: Ahhhh, 💡. Got it. Yes!

jasonaowen:
Thanks, @reefdog!

jasonaowen:
> The SVG logo replacement can come in a future change, as it will probably take a while to get it from the designer.

My hope had been that the high-quality SVG version would be quickly forthcoming, and in general I do not love adding binary assets to git repos when I can avoid it, which is why I had initially marked this as a draft. I agree with @reefdog, though, that we can fix it later, and speed is more important now than correctness.

I ran [pngcrush](https://pmt.sourceforge.io/pngcrush/) on the logo to reduce it from 292k to 140k, so I'll force push this branch with that replacement and then merge after CI is done.

> We shouldn't store out image assets at the root of /src, but I recognize this was following the stub pattern. The real style/asset organization can come while delivering https://github.com/PhilanthropyDataCommons/data-viewer/issues/12.

I generally agree, but did this because it follows the pattern for SVG files, which get compiled into the final product, with the hope that we'd be able to quickly replace the PNG. I'll also be happy to fix this later!

PR #28, opened 2023-03-06
Add and configure Storybook

### LINKED ISSUE #27: Set up Storybook for documenting front-end components
[Storybook](https://storybook.js.org) is a well-regarded platform (add-on? thingy?) for describing your UI components. It gives us a formal way of documenting and demonstrating our components and their various states. This is useful both for ourselves / the developer creating the components, and for others / developers using them.

reefdog:
This PR adds [Storybook](https://storybook.js.org/) so we can manage, document, and test UI components. It also makes three small prerequisite changes to support Storybook. (I'm open to splitting these off to another PR.)

Some implementation notes:
- We are using Storybook v7, currently in beta, because it resolves lots of issues we had with the v6 installation (vulnerable dependencies, inscrutable Webpack errors, etc.).
- I'm including all the auto-generated sample components and stories (everything inside `/src/stories/`) because they're handy examples to follow. I expect to quickly replace them with our own.
  - I had to modify the samples to resolve ESLint violations.
- I disabled [telemetry](https://storybook.js.org/docs/react/configure/telemetry).

**Testing:**
1. `npm i` to install new packages
2. `npm run storybook` to fire up Storybook
3. Open http://localhost:6006/ in your browser
4. ✅ See the "Welcome to Storybook" sample page

Closes #27


package.json
```diff
@@ -6,27 +6,41 @@
...
     "@types/jest": "^27.5.2",
     "@types/node": "^16.18.12",
     "@types/react": "^18.0.28",
     "@types/react-dom": "^18.0.11",
+    "babel-plugin-named-exports-order": "^0.0.2",
     "eslint-config-airbnb": "^19.0.4",
     "eslint-config-airbnb-typescript": "^17.0.0",
-    "react-scripts": "5.0.1",
-    "typescript": "^4.9.5"
+    "prop-types": "^15.8.1",
```
jasonaowen, line 31: Do we need `prop-types`? I don't see any references to it in this diff.

We do already have it as a transitive dependency, fwiw:

```sh
$ npm list prop-types
data-viewer@0.1.0
└─┬ eslint-config-airbnb@19.0.4
  └─┬ eslint-plugin-react@7.32.2
    └── prop-types@15.8.1
```

```suggestion
```

package.json
```diff
@@ -6,27 +6,41 @@
...
+    "storybook": "^7.0.0-beta.62",
+    "typescript": "^4.9.5",
+    "webpack": "^5.75.0"
```
jasonaowen, line 34: `webpack` is currently a transitive dependency of `react-scripts`; do we need to depend on it directly?

```suggestion
```

jasonaowen (approved):
This looks good to me! I was able to both run and build Storybook. Thanks for following up on the linter stuff.

These are fantastically well-structured commits & commit messages - thank you!! A joy to review.

Do you have any concerns about publishing the storybook output on the internet? I feel like it's probably worthwhile to have an interface folks can poke at, without requiring them to spin up the project. If so, may I suggest:

```diff
diff --git a/package.json b/package.json
index 378166e..2bc0a97 100644
--- a/package.json
+++ b/package.json
@@ -35,12 +35,13 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "npm run build:app && npm run build:storybook",
+    "build:app": "react-scripts build",
+    "build:storybook": "storybook build -o build/storybook",
+    "storybook": "storybook dev -p 6006",
     "test": "react-scripts test",
     "lint": "eslint src/",
-    "eject": "react-scripts eject",
-    "storybook": "storybook dev -p 6006",
-    "build-storybook": "storybook build"
+    "eject": "react-scripts eject"
   },
   "eslintConfig": {
     "extends": [
```

If you like, I'd be happy to make that change and update this PR - but I don't want to make that kind of change to someone else's PR without permission. :)


Aside from that, a few minor suggestions & questions below.

package.json
```diff
@@ -18,7 +19,6 @@
     "@types/react-dom": "^18.0.11",
     "eslint-config-airbnb": "^19.0.4",
     "eslint-config-airbnb-typescript": "^17.0.0",
-    "react-scripts": "5.0.1",
```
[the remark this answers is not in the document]
reefdog: > Oh, I forgot to ask - does Storybook stay mad if this is in devDependencies? Or only during the install process?

😮 You're right! This only bit during the install. It runs fine now. I'll move it back, excellent catch.

> > Arguably, more of our packages should be moved here, but meanwhile we should at least move this one.
> 
> What else are you thinking? Or, are you saying we shouldn't be using `devDependencies` at all in a front-end web app like this? I'm receptive to either!

This was mostly a reference to our conversation. I prefer to organize them like you said, separating packages used during development (`devDependencies`) from those intended to be used in the final build (`dependencies`). But I'm unsure there's any practical benefit to that, and evidently it can cause a few speedbumps.

I don't think we need to change anything now, but I remain receptive to jettisoning `devDependencies` if it ever causes problems.

For now, since I'm going to remove this commit from the branch, this editorializing will be removed anyway, lol.

package.json
reefdog: Ah, I'll bet you're right! I simply committed the changes that `storybook init` created, apologies for being less attentive. I'll experiment with removing this (and the explicit webpack dependency) and see if it all builds and runs correctly, then will rewrite the commit without it.

reefdog:
@jasonaowen Thanks for the excellent review! I just made several of your suggested changes and force-pushed, including removing the `react-scripts` location change and removing the explicit dependencies on `prop-types` and `webpack`.

I wrote up my thoughts about including the Storybook build in Zulip, but I'll copy them here (with some tweaks):

> Do you have any concerns about publishing the storybook output on the internet? I feel like it's probably worthwhile to have an interface folks can poke at, without requiring them to spin up the project.

I don't have privacy or security concerns, but I'm not sure I see sufficient value either. The components will be pretty isolated (by design) and I suppose will give people an idea of the aesthetic, but they're not really intended to be useful to anyone except developers using them.

The primary risk would be the production Storybook deployment being treated as canonical, yet possibly being out of date with `main`. Now, if we're CDing `main` to the production URL then I suppose that wouldn't happen, but I assumed we'd be doing tagged releases for production deploys. (I personally prefer that extra safety step to the potentially accidental magic of CDing `main`.)

Overall I don't feel too strongly about it, and if you do see value then by all means! I just think -- especially early on while things are moving quickly -- I'd ask you / other devs to be sure they were looking at `main`'s Storybook rather than the production URL (unless those are tightly coupled).

I did really like your suggested syntax for the `build:storybook` command, as well as building into the `/build/` directory, so I stole that line. I'm just realizing, though, that maybe it's a problem to build into it if we _don't_ intend to deploy it (even though it's a separate / optional command)?

jasonaowen:
> The primary risk would be the production Storybook deployment being treated as canonical, yet possibly being out of date with `main`. Now, if we're CDing `main` to the production URL then I suppose that wouldn't happen, but I assumed we'd be doing tagged releases for production deploys. (I personally prefer that extra safety step to the potentially accidental magic of CDing `main`.)

That is the current configuration - Netlify deploys on every merge to `main`! Can you say more about your preference? I tend to believe that deploys should be frequent, automatic, and low-friction, but maybe some additional gates would be helpful in this situation.

> I did really like your suggested syntax for the `build:storybook` command, as well as building into the `/build/` directory, so I stole that line. I'm just realizing, though, that maybe it's a problem to build into it if we don't intend to deploy it (even though it's a separate / optional command)?

I think this is still fine if we decide not to deploy it; Netlify just runs `npm run build` and hosts the `/build/` directory, so if the `build` script does not include the `build:storybook` script, Storybook won't run. The only other time we'd run the `build:storybook` script would be in CI, to make sure that it is buildable, but then throw away the output.

PR #29, opened 2023-03-07
Add routing

jasonaowen:
Install [react-router](https://reactrouter.com/en/main), and set up a few routes: the existing placeholder content, and a 404 error page. Hopefully, this will be enough to set the pattern for @reefdog to work on adding more pages if needed, but we'll figure that out as we go!

Note that this has merge conflicts with #28; once that PR is merged, I'll rebase and resolve them.

Issue https://github.com/PhilanthropyDataCommons/data-viewer/issues/10 Scaffold routing and page construction

### DIFF
--- a/public/_redirects
+++ b/public/_redirects
@@ -1,0 +1,3 @@
+# https://docs.netlify.com/routing/redirects/
+
+/*  /index.html  200
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,6 +11,7 @@
         "pino": "^8.11.0",
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
+        "react-router-dom": "^6.8.2",
         "web-vitals": "^2.1.4"
       },
       "devDependencies": {
@@ -3896,6 +3897,14 @@
         "webpack-plugin-serve": {
           "optional": true
         }
+      }
+    },
+    "node_modules/@remix-run/router": {
+      "version": "1.3.3",
+      "resolved": "https://registry.npmjs.org/@remix-run/router/-/router-1.3.3.tgz",
+      "integrity": "sha512-YRHie1yQEj0kqqCTCJEfHqYSSNlZQ696QJG+MMiW4mxSl9I0ojz/eRhJS4fs88Z5i6D1SmoF9d3K99/QOhI8/w==",
+      "engines": {
+        "node": ">=14"
       }
     },
     "node_modules/@rollup/plugin-babel": {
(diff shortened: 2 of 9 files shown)


jasonaowen:
Okay, that involved some thrashing to properly configure Netlify, but you should now be able to go to https://deploy-preview-29--philanthropy-data-commons-viewer.netlify.app/test-not-found or so and see the new 404 page. Likewise, you should be able to `npm start` and navigate to http://localhost:3000/not-found to see it.

public/_redirects
```diff
@@ -0,0 +1,3 @@
+# https://docs.netlify.com/routing/redirects/
+
+/*  /index.html  200
```
reefdog, line 3: A consequence of this is that our 404 page returns an HTTP 200 status code. (Netlify [specifically mention this](https://docs.netlify.com/routing/redirects/rewrites-proxies/#history-pushstate-and-single-page-apps).) This seems inherent in a SPA architecture, but I sure wish we could respond with true 404s. I'm kind of surprised react-router doesn't provide this, but quick searching doesn't provide any answers.

Principles aside, a practical reason to respond with a true 404 is to signal search engines that the URL isn't found / shouldn't be indexed. Even though we're currently private, I think we should assume some day we may serve _some_ public pages and prevent `/anything-i-dream-up-and-link-to-or-submit-to-an-index-maliciously` from being indexed.

Could you add a [`noindex` meta tag](https://developers.google.com/search/docs/crawling-indexing/block-indexing) to the 404 page? If doing _this_ will require some churn to figure out how to write to the page `<head>` from an individual page component, then feel free to create a task and punt.

reefdog (approved):
This is great! I recognize you'll need to rebase after we merge in #28 but overall appreciate the work and approach here.

I made one request re: handling our 404 page, but there's an escape valve of "create task and punt" if you'd like, so I'm marking this as approved since that decision is yours to make.

public/_redirects
jasonaowen: We do have some options! This was just the quick approach. We could add an error page and specifically enumerate the valid paths in the `_redirects` file, or we could add a `robots.txt` file.

Adding an error page would be basically a copy of `index.html` that has a different title (which, note to self, we need to change the title from "React App"). Then, we'd update `_redirects` to send known paths to `index.html` with a 200, and unknown paths to `error.html` with a 404.

I'm kind of inclined, though, to just add a `robots.txt` file that forbids all bots from crawling the site - there are no intended public pages, and users must be signed in for everything, so there's not really anything worth indexing. The drawback to the previous approach is that every time we add, rename, or remove a route, we also need to update `_redirects`, which is an ongoing and easy-to-miss development cost - I worry one day we'll forget, and then have a bug where "I can't share my customized Application view" or whatever.

jasonaowen: Updated `robots.txt`! I'd be interested to hear if you think the `_redirects` work is also worthwhile, @reefdog, as the two approaches are not mutually exclusive.

reefdog: 1. I agree that we shouldn't list paths in `_redirect`. That sounds like brittle redundant work.
2. Adding a `robots.txt` is just fine and we should do it! But the specific thing I'm worried about (making sure our 404 page itself — which could be served to an infinite variation of paths — isn't ever indexed) is just implicitly protected by `robots.txt`. I'd rather make it explicitly protected by adding a `noindex` to the page itself as well. I know we've got no plans to open the site up soon, but if we even decide to host a vanilla unauthenticated marketing page, then we'll also be serving a 404 page, and should be sure to `noindex` it.

I don't want to bike shed this one so feel free to override and merge in! But I would at least want to create a task so we had it in history that if we release the blanket robots denial, we should add this.

jasonaowen: Should we also add `noindex` to the main page? If so, that's easily done, and yeah I will be happy to do it. If not, if that tag should only be on the error page, that becomes more difficult, because I am not sure how to have a separate error page without a complete enumeration of the valid routes. Maybe there's an approach I missed!

jasonaowen: Added #31 to capture this and moving on! Thanks, @reefdog.

reefdog: Perfect, thanks!

PR #32, opened 2023-03-07
Organize components and pages

### LINKED ISSUE #10: Scaffold routing and page construction
We need to establish patterns for:

- How we create new routes
- Any rules/patterns we want to establish around routes (RESTfulness, state preservation, etc.)
- Where (in the project/filesystem) pages and components live

Some of this should probably be a side effect of Doing Other Work, so the real scope of this issue is to put a few stub items in the filesystem that demonstrate the patterns we want to follow.

reefdog:
This PR demonstrates an organization strategy for components and pages, putting them into `/src/components` and `/src/pages` directories respectively. It also puts our images into a `/src/images` directory to minimize the number of files in the root of `/src`.

Our own app assets are at the root of each respective directory, while the Storybook examples have been further organized into `example/` directories (e.g., `/src/components/example/*`).

**Testing:**
- `npm start`
- ✅ Verify http://localhost:3000 works as before
- `npm run storybook`
- ✅ Verify http://localhost:6006 works as before

Closes #10


jasonaowen (approved):
This seems fine! I tested the deployed version, and ran `npm run build:storybook` locally, and everything worked.

I haven't followed this pattern before of separating routable components from other components, and maybe I'll have opinions about this as we get deeper, but this is a great starting point.

Thanks, @reefdog!

PR #33, opened 2023-03-09
Move example stories into `example/` directory

reefdog:
This PR includes a small miss from #32.

I didn’t understand the importance of story placement, and left the example stories in the root `stories/` directory. Now, writing stories of my own for #12, I want these examples to live in `stories/example/`. So, this PR moves them there.

Affects #10 (which is already closed)


jasonaowen (approved):
Good catch!

PR #34, opened 2023-03-09
Create Button component and story

reefdog:
This PR adds our `Button` component as designed in the Figma UI, along with all its various possible states. It also adds a Storybook story to demonstrate those states and allow playing with the various arguments.

We'll be adding more components, but I wanted to deliver this PR quickly to demonstrate patterns and get any overall feedback.

It includes some related global work, such as including global CSS in Storybook previews and affecting global styles and CSS variables. This work is arguably beyond this PR, but was a prerequisite to it and still relevant to the attached issue.

It also adds the [Heroicons](https://heroicons.com/) React library, which we use for icons (such as those in buttons).

**Testing:**
- `npm i` to get `@heroicons/react`
- `npm run storybook`
- Hit http://localhost:6006/?path=/docs/stories-button--docs and play with the button!

Affects #12

### DIFF
--- a/src/components/Button.tsx
+++ b/src/components/Button.tsx
@@ -1,0 +1,69 @@
+import React from 'react';
+import './button.css';
+
+interface ButtonProps {
+  /**
+   * Contents of a `<Button>` can be a mix of strings and JSX elements,
+   * and will be flex-aligned.
+   */
+  children: JSX.Element | JSX.Element[] | string;
+  color?: 'gray' | 'blue' | 'red';
+  /**
+   * Makes the button dark with light text.
+   */
+  inverted?: boolean;
+  /**
+   * Adds a visible border to the button.
+   */
+  bordered?: boolean;
+  /**
+   * Adds a notification disc to the button.
+   */
+  notification?: boolean;
+  disabled?: boolean;
+  /**
+   * Optional click handler
+   */
+  onClick?: () => void;
+}
+
+/**
+ * Primary UI component for user interaction.
+ */
+export const Button = ({
+  children,
+  color = 'gray',
+  inverted = false,
+  bordered = true,
+  notification = false,
+  disabled = false,
+  onClick = (() => true),
+}: ButtonProps) => {
+  const buttonClassNames = [
+    'button',
+    `button--color-${color}`,
+  ];
+
+  if (inverted) {
+    buttonClassNames.push('button--inverted');
+  }
+
+  if (bordered) {
+    buttonClassNames.push('button--bordered');
+  }
+
+  if (notification) {
+    buttonClassNames.push('button--notification');
+  }
+
+  return (
+    <button
+      type="button"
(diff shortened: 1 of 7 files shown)


reefdog:
Whoops, accidentally closed this by trying to rename the branch out from under it. Reverted all that all (should be) well again.

src/index.css
```diff
@@ -19,3 +19,9 @@ body {
 code {
   font-family: var(--font-family--monospace);
 }
+
+*,
+*:before,
+*:after {
+  box-sizing: border-box;
```
jasonaowen, line 34: Neat, TIL!

src/components/button.css
```diff
@@ -0,0 +1,99 @@
+:root {
+  --button--focus-color: blue;
+  --button--font-size: 0.9rem;
+  --button--height: 44px;
+  --button--icon-size: 24px;
```
jasonaowen, line 5: Should these be in relative units, too?

src/components/Button.tsx
```diff
@@ -0,0 +1,69 @@
...
+   */
+  inverted?: boolean;
+  /**
+   * Adds a visible border to the button.
+   */
+  bordered?: boolean;
+  /**
+   * Adds a notification disc to the button.
+   */
+  notification?: boolean;
```
jasonaowen, line 22: Interesting! Do you have any intended uses for notifications?

If not, might be better to leave it out - less complexity, and less code to maintain!

jasonaowen (approved):
Awesome! Storybooks are really cool. I look forward to seeing this live!

src/components/button.css
reefdog: They… should!

src/components/Button.tsx
reefdog: I do indeed have a use, although "notification" was a clumsy word for the semantic meaning of indicating the button is "in use":

button-example

(Note that I changed the design during implementation, because I kind of prefer the corner layout since it doesn't affect the size of the button. Open to change again if necessary.)

Semantically, "enabled" or "active" are both more accurate. My concern about both of _those_ words is that they already have semantic meaning to buttons: "enabled" is the opposite of "disabled", which is a native `<button>` property, and "active" is already a `<button>` property (at least in CSS, e.g. `button:active`) that means "currently being clicked". 😣

Open to suggested terminology changes here! Maybe just… `inUse` given that's what I naturally said above?

reefdog: Actually I'll stop painting the bike shed. We can rename it if we stumble over it in the future!

jasonaowen: "Notification" seems like a fine term! Although I did appreciate the "blip" working title you mentioned. :smile: 

I was less concerned with the name and more concerned about including unused functionality, but you have a plan so I'm no longer concerned - thanks for the explanation!

Issue #6, opened 2023-01-10
Proof of concept authentication via Keycloak


reefdog:
We need to build a skeletal, proof of concept authentication via Keycloak. The necessary workflow:

1. From the Data Viewer, click a Login link
2. Be taken to Keycloak to authenticate
3. Upon successful authentication, be sent back to Data Viewer
4. Data Viewer recognizes who you are and maintains a session
5. ~~Data Viewer is able to make API calls to the service layer using a token~~

Each of those must work for this task to be complete.

**Reference links:**

- [Auth.js + SvelteKit](https://authjs.dev/reference/sveltekit/modules/main)
- [Auth.js's documentation on Keycloak as a provider](https://authjs.dev/reference/oauth-providers/keycloak)

<details>
<summary><b>Stretch goals…</b></summary>

These niceties will probably be spun off to their own tasks since they're polish:

- Ensure the front end can gracefully handle the API's rejection of a token. (Verify it against Keycloak? Renew it and try again? Prompt to re-authenticate?)
- Password reset ← Update: This will almost certainly be handled entirely by the Keycloak installation.
- Account modification (email, password, etc.) ← Update: This will almost certainly be handled entirely by the Keycloak installation.
- ~Self-registration~ ← Update: This isn't even on the radar, scrap as a stretch goal.

</details>

jasonaowen:
The [lack of a dedicated back-end](https://github.com/PhilanthropyDataCommons/data-viewer/issues/4) leads to a few security implications that are worth mentioning:
- as a public client, we'll use the [Authorization Code flow](https://oauth.net/2/grant-types/authorization-code/) with [PKCE](https://oauth.net/2/pkce/) (see also [Auth0's documentation](https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow-with-proof-key-for-code-exchange-pkce), which I find to be more approachable)
- as a public client, the client credentials alone should not be granted any access in Keycloak; in particular, the [Client Credentials flow](https://oauth.net/2/grant-types/client-credentials/) should be disabled for this client, and this operational concern should be documented in the setup instructions of this repo
[...]

PR #35, opened 2023-03-10
Log in through OpenID Connect

jasonaowen:
Use the [`@axa-fr/react-oidc` library](https://www.npmjs.com/package/@axa-fr/react-oidc) to allow logging in via OpenID Connect. Add a `<User>` component to the `<Header>` that will show either a login button or the current user's name.

Note that the login button will probably not last long-term, but until we start showing any real data, it's a useful placeholder. It also looks good during demos, rather than the automatic login I expect we'll want. The `<User>` component is also unstyled for the moment; this commit is more focused on the functionality.

This adds a fair bit of required configuration. Add an `.env.example` file documenting several environment variables, both new and existing, and both optional and required.

The library supports service-worker-only communication between the app and the API, which is intended to improve security; at least for the moment, do not opt in, as we still need to figure out how to handle authentication in PR deploy previews. This may be something we want eventually, at which point we should commit the currently-ignored service worker & configuration files.

I did work far enough forward to confirm that I could make authenticated requests to the API, but am not including that yet both for the sake of scope and because it depends on unmerged service-side changes.

Issue #6 Proof of concept authentication via Keycloak

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,43 @@
 # data-viewer
 A UI for interacting with the data stored in the PDC service
 
+## Dependencies
+
+The application depends on an
+Open ID Connect ("OIDC")
+identity provider ("IdP")
+for authentication.
+It should be agnostic about what kind of IdP is used.
+To use the data-viewer application,
+you will need to be able to log in.
+
+## Configuration
+
+### Identity Provider
+
+The IdP needs a client for the data-viewer.
+The IdP should be configured for the client to use the
+[Authorization Code flow](https://oauth.net/2/grant-types/authorization-code/).
+This client is a public client,
+and so should *not* have a secret,
+and should not be able to log in with a client credentials grant.
+
+Set the client root URL to the app's URL
+(or `http://localhost:${PORT}` in development),
+and set the redirect URI to `/authentication/callback`.
+
+### Development
+
+Copy the `.env.example` file to `.env.local`,
+and replace the placeholder values with your specifics.
+Your new `.env.local` file will not be picked up by git.
+
+### Production
+
+The environment variables documented in `.env.example`
+will need to be configured in your deployment system;
+the values must be present at build time.
+
 ## Available Scripts
 
 In the project directory, you can run:
--- a/.env.example
+++ b/.env.example
@@ -1,0 +1,19 @@
+# Built-in Create React App environment variables
+# See https://create-react-app.dev/docs/advanced-configuration
+# for full list and documentation
+#DISABLE_ESLINT_PLUGIN=false
+#PORT=3000
+
+# Show logs in the browser console
+# See https://github.com/pinojs/pino/blob/49a29bb9d7895e791e7166fadfd7fd379f8db504/docs/api.md#loggerlevels-object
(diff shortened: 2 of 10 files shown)


jasonaowen:
Preview of the login button:
Screen shot showing login button

The signed-in view:
Screen shot showing my name

I configured the relevant environment variables in Netlify, so it works both in the deploy preview and should work in production. I manually added the deploy preview as an allowed origin to Keycloak; that approach isn't sustainable, but I think is an important step to show this PR works! I'll experiment with more flexible options later.

To test locally, configure your environment by following the (new) instructions in the README, running `npm start`, and confirming you can log in. You might also try *not* configuring your environment, and confirming it blows up with an error in the console log!

To test in the deploy preview, you'll need an account in Keycloak - contact me or @bickelj.

README.md
```diff
@@ -1,6 +1,43 @@
...
 
+## Dependencies
+
+The application depends on an
+Open ID Connect ("OIDC")
+identity provider ("IdP")
+for authentication.
+It should be agnostic about what kind of IdP is used.
+To use the data-viewer application,
+you will need to be able to log in.
```
reefdog, line 12: What's with the short line lengths in this document? I'm not criticizing, genuinely curious if this is meaningful!

reefdog:
This is very cool! At least, from the outside. I'm not going to be able to kick its tires until morning.

I have a few PRs with some components you may find useful for this, though I don't mind if this gets merged in first and I have to update mine. In particular, though, the just-opened #36 has a navbar that would be useful here.

What we'll likely want to do is move the `User` component from `Header` into my new [`Navbar`](https://github.com/PhilanthropyDataCommons/data-viewer/pull/36/files#diff-a0439d8a2cb6dee715c41a795bb174a2a65ad61704e88691fe5ab2b2bd6f798e), nested inside a new `<li>` element, and outputting either one of my new [`Button`](https://github.com/PhilanthropyDataCommons/data-viewer/pull/34/files#diff-1e4ab16772826f4d218576ae2acae4e6ab1c99196422e2eecd23db06ff869357) components (not the example one) or just text, as you are now.

bickelj (approved):
I don't know React, so all I can say about the code is "it looks good to me." The library looks to be actively maintained. It doesn't look like a lot of boilerplate was needed.

I was able to run all the `npm run` commands successfully (except eject) locally.

Using the preview site I was able to log in successfully via the login button, close that tab, open the preview site again, click login and was logged in without doing the username/passphrase dance again. I found my JSON Web Token in session storage.

README.md
jasonaowen: This is a practice I picked up a while ago: [Semantic Linefeeds](https://rhodesmill.org/brandon/2012/one-sentence-per-line/)! Markdown will collapse subsequent lines into a single paragraph, so we can add line breaks as we please. Long lines make it hard to review changes, so line breaks are useful. The idea, then, is to treat markdown as source code, and make it easier to edit in a similar way: have as few reasons to modify a line as possible!

reefdog (approved):
This looks and works great! I'll adapt my PRs to it.

Issue #5, opened 2022-12-12
Mock up UI layouts


reefdog:
In advance of our UX research sessions, we need to mock up the major prospective interface and functionality of the Data Viewer, including:

- Proposals table
- Proposals detail page
- Search and filtering functionality
- Site boilerplate (navigation, etc.)

reefdog:
This is currently in progress.

- [Figma document](https://www.figma.com/file/9QpV41Ibm6gFXFj8wmwfc0/PDC-Data-Viewer?node-id=0%3A1&t=vvkimwk28gAkNVf0-1)
- [Figma prototype](https://www.figma.com/proto/9QpV41Ibm6gFXFj8wmwfc0/PDC-Data-Viewer?page-id=0%3A1&node-id=290%3A1900&viewport=-711%2C360%2C0.2&scaling=scale-down&starting-point-node-id=290%3A1900)

reefdog:
I owed @jim-mcgowan a few screenshots, and figured I may as well put them here for future reference as well.

<img width="1512" alt="PDC Data Viewer 01 - Proposals Table" src="https://user-images.githubusercontent.com/4731/220751445-2701db4c-16d0-4e3b-83ce-df2b575f6f57.png">
<img width="1512" alt="PDC Data Viewer 02 - Proposal" src="https://user-images.githubusercontent.com/4731/220751450-d328b240-42f7-41fe-a1dd-887f3318edd9.png">
<img width="1512" alt="PDC Data Viewer 03 - Proposal History" src="https://user-images.githubusercontent.com/4731/220751454-1eb63a34-11e4-48ae-b8fe-34b85325af9e.png">
<img width="1512" alt="PDC Data Viewer 04 - Proposal Shortcodes" src="https://user-images.githubusercontent.com/4731/220751457-f69b77ec-25c4-4f6c-adae-1cf4777db443.png">
<img width="1512" alt="PDC Data Viewer 05 - Proposal Grid View" src="https://user-images.githubusercontent.com/4731/220751459-7e215098-020f-4feb-b643-a5bd1638cf94.png">
<img width="1512" alt="PDC Data Viewer 06 - Proposal with Applicant" src="https://user-images.githubusercontent.com/4731/220751461-105dd5bd-4f1d-45a9-be7f-6514d6452dba.png">
[...]

PR #36, opened 2023-03-10
Implement grid-based layout and improved header/navbar

reefdog:
_Note: This PR ladders off #34, which branch is set as the base._

This PR implements our grid-based interface. The header sits atop the page while the main area takes up the rest of the viewport height. (Future changes will add inner-scrolling panels within the main area.) It adds consistent spacing around and within elements, and applies some base typography and coloring.

It also adds a Navbar class with an example link (for "Home", which we don't really want) that will be replaced as we build real pages.

Affects #12


jasonaowen:
Oh interesting, there's no deploy preview - looks like that's because the merge base is not `main`! I don't see a way to change that, and the docs don't really cover it, and probably it's fine anyway; I was just surprised.

src/App.css
```diff
@@ -1,38 +1,69 @@
...
-}
+.App-header {
+  display: flex;
+  justify-content: space-between;
+  align-items: center;
 
-.React-logo {
-  height: 50vh;
-  width: 50vw;
+  padding: var(--spacing--gap);
```
jasonaowen, line 15: There ends up being kind of a lot of empty space at the top of the page. Now that I look more closely, this is in keeping with the mockups you rendered in #5; I forgot I had been looking at scaled images! Maybe the goal is to have a counterbalance against the density that's coming later? I only noticed because my previous screenshots fit in 440x440px, and with this change that's no longer enough vertical space!

jasonaowen (approved):
Code-wise, this looks good to me! Two small questions below.

As you suggested, @reefdog, let's check in on how to proceed with merging!

Screen shot of main page with the new header and background

src/components/Header.tsx
```diff
@@ -3,7 +3,9 @@ import logo from '../images/pdc-logo.png';
 
 const Header = () => (
   <header className="App-header">
-    <img src={logo} className="App-logo" alt="logo" />
+    <a href="/">
+      <img src={logo} height="60" alt="Philanthropy Data Commons logo" />
```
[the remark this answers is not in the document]
reefdog: > Why set the height here?

Old habit, signaling to the browser how to lay out the image before it's loaded (i.e. before its dimensions are known) and before CSS has been applied. But, in our case, it's unlikely to matter and more likely to be out of date, so, I'll remove!

> Should it have a unit?

[Not in HTML!](https://www.w3schools.com/tags/att_img_height.asp)

src/App.css
reefdog: All that extra space comes from the amount of invisible padding in the spinning react image:

image

Being a temporary stand-in anyway, I'm going to not worry about it. We can create a version of the logo without the padding if we want!

jasonaowen: Ah sorry, I meant the space above the PDC logo / home button! With these changes the PDC logo is larger and has more padding above it:

before:
image

after:
image

reefdog: Oh, yes! It is quite a bit roomier than before, but in keeping with the mockup.

We can adjust this down if we'd like at smaller viewports, though we're intentionally unconcerned with the mobile experience. I do see how this would be nicer for screenshotting, though.

reefdog:
@jasonaowen I'd love your eyes on this one last time after the merge and after the changes I made to your `User` component to fit into the new navbar!

jasonaowen:
Great, looking now!

src/index.css
```diff
@@ -12,11 +14,16 @@
     sans-serif;
   --font-family--monospace: source-code-pro, Menlo, Monaco, Consolas, 'Courier New', monospace;
 
+  --spacing--gap: 10px;
```
jasonaowen, line 17: Should I keep banging the "relative units" drum? I admit my current understanding is basically "always use relative units" - which is an ironically absolutist view - so if you have a better mindset you can share I'd appreciate it!

jasonaowen (commented):
Oh hm, GitHub correctly updated the merge base, but Netlify didn't notice. :confused: Let me see if I can poke it; otherwise I'll spin it up locally!

src/App.css
```diff
@@ -1,38 +1,92 @@
...
+}
+
+.App-navbar__item {
+  all: unset;
+
   display: flex;
-  flex-direction: row;
-  justify-content: space-between;
-  color: black;
+  align-items: top;
```
jasonaowen, line 43: My browser dev tools does not like `top` as a value for [`align-items`](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items); did you mean `start` or `flex-start`?

src/components/Navbar.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+import { HomeIcon } from '@heroicons/react/24/solid';
+import { User } from './User';
+
+const Navbar = () => (
+  <nav className="App-navbar">
+    <ul>
+      <li>
+        <NavLink to="/" className={({ isActive }) => `App-navbar__item ${isActive ? 'App-navbar__item--active' : ''}`}>
```
jasonaowen, line 10: This is long enough to wrap in my browser; mind if we wrap it?

```suggestion
        <NavLink
          to="/"
          className={({ isActive }) => (
            `App-navbar__item ${isActive ? 'App-navbar__item--active' : ''}`
          )}
        >
```

Alternatively, [a `<NavLink>` has a default `active` class](https://reactrouter.com/en/main/components/nav-link#default-active-class); is writing a rule for `.App-navbar__item a.active` a terrible idea?

src/components/User.tsx
```diff
@@ -10,22 +11,34 @@ const User = () => {
...
+          <UserIcon />
+          Loading…
+        </div>
+      );
     case OidcUserStatus.Unauthenticated:
       return (
         <button
           type="button"
-          className="btn btn-primary"
+          className="App-navbar__item"
```
jasonaowen, line 24: Do we want to use the new `<Button>` component here?

src/index.css
reefdog: It's good to bang that drum! But in this case, I don't _want_ our spacing gap to be relative to font size. (At least, not _this_ spacing gap. We may want others, such as between typographic blocks, to be relative.) If a user has increased their default font size for accessibility reasons, I don't want the gap around our panels to be any larger, and in fact doing so would be counterproductive as it would gobble up valuable real-estate.

[Josh Comeau has an excellent article](https://www.joshwcomeau.com/css/surprising-truth-about-pixels-and-accessibility/) that formed my thinking about when to use fixed versus relative units. I don't always make the correct choice (see: the button styling), but in this case and a few others, the choice was intentional.

That said, _please_ do continue to call attention to them when you see them and if you suspect it's a time when relative units are preferable!

src/components/Navbar.tsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import { NavLink } from 'react-router-dom';
+import { HomeIcon } from '@heroicons/react/24/solid';
+import { User } from './User';
+
+const Navbar = () => (
```
jasonaowen, line 6: Can we add a story for the new `<Navbar>`? And/or for the `<Header>`?

I'm not sure how the `<User>` component will work, though; I'll have to look at the Storybook docs...

jasonaowen (changes requested):
This is looking good to me! A few concerns below, but certainly nothing major.

src/index.css
jasonaowen: Oh awesome, I'll take a look - thanks for sharing, @reefdog!

src/App.css
reefdog: @jasonaowen Ah, I did mean `flex-start`! I thought I'd fixed that, but maybe I habit-typed it again. Good catch!

src/components/User.tsx
reefdog: I did think about that, but I don't _think_ I want to use the button styling up here. Maybe… maybe so though. Let me play with it! (It would certainly let me remove some stuff, as I had to tweak the navbar-item to make it button-compatible.)

src/components/Navbar.tsx
reefdog: I wasn't thinking that we would bother making stories for "one-time use" components… but as an artifact of communicating use, maybe it's worthwhile even for that. Good idea.

src/components/Navbar.tsx
reefdog: Oh, it defaults to that? Yes, I'm on board with that for sure.

jasonaowen (approved):
Looking good!

Screen Shot 2023-03-10 at 16 17 46

Thanks, @reefdog!

Issue #9, opened 2023-02-28
Set up CD


reefdog:
After choosing a host (#8), we'll want to configure CD so that code is automatically deployed. I like Netlify's ability to spin up deployment preview URLs for each new pull request.

We've not discussed the QA or production deployment workflow and I'll bet @jasonaowen has better ideas than me. If it's possible to do something like the following, I think it's ideal:

- PRs auto-deploy to PR-specific domains (definitely doable)
- `main` auto-deploys to a stable development/QA domain
- Releases auto-deploy to production

jasonaowen:
I've got Netlify set up, and we now have DNS: https://pilot.philanthropydatacommons.org/ !

PR #38, opened 2023-03-10
Build and publish Storybook in CI & CD

jasonaowen:
When we first introduced Storybook, we had considered whether or not to publish the component previews it provides as part of our continuous delivery system, and initially decided to defer that work. Now that we're starting to use it in earnest, let's make it easier for others to see the value it provides!

Update the `build` script, which is invoked by both our continuous integration and continuous delivery systems, to build both the application and the Storybook content. This should be published under `/storybook`.

Issue #2 Set up CI
Issue #9 Set up CD
Issue #27 Set up Storybook for documenting front-end components
PR #28 Add and configure Storybook

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,8 @@
   },
   "scripts": {
     "start": "react-scripts start",
-    "build": "react-scripts build",
+    "build": "npm run build:app && npm run build:storybook",
+    "build:app": "react-scripts build",
     "build:storybook": "storybook build -o build/storybook",
     "storybook": "storybook dev -p 6006",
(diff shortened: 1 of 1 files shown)


jasonaowen:
You can see the built version on this PR's deploy preview: https://deploy-preview-38--philanthropy-data-commons-viewer.netlify.app/storybook/

Later we might add a link from the app header, but for now you'll need to know the magic path.

reefdog (approved):
Huzzah! I knew we should do this from the very beginning. [citation needed]

PR #40, opened 2023-03-13
Add and use Source Sans Pro font

### LINKED ISSUE #37: Can we use Source Sans Pro?
Discussion time!

I designed our mockup UI with the [Source Sans Pro](https://fonts.google.com/specimen/Source+Sans+Pro/about) typeface. After considering not using a webfont at all, I decided it's worthwhile for style and layout. But, I want to make sure we're okay with it!

- It's licensed under the [Open Font License](https://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL), which I believe is compatible with our use.
- Rather than load the font from Google, we'd download the files and serve them directly ourselves.

Any objections or additional thoughts?

reefdog:
_Pending sign-off on #37._

This PR adds the Source Sans Pro font to our project (as a Node package using Fonstource), enables it site-wide, and adapts our button styling since it's a slightly smaller typeface than our previous one.

It also includes a small tweak to how we specify font-weights around the project, which is related because for performance reasons we have to be specific about which weights of the webfont we want to include in the project. Specifying our font-weights as CSS variables makes it easier to be consistent and not accidentally use an excluded weight.

I'm marking this as closing the discussion issue, because presumably we won't merge this in unless the discussion has concluded that it's okay to do so.

Affects #12
Closes #37


jasonaowen (approved):
Awesome, I'm glad Fontsource worked so well! This looks great to me; +1 to merging. Thanks for figuring out how it works in practice, @reefdog!

@reefdog, if you find yourself blocked on this at all - even just thinking "hmm, maybe I should rebase this other branch onto `12-add-source-sans-pro`", but especially if you start thinking about how to structure subsequent PRs - I'd say go ahead and merge it. Either way, we can discuss it at tomorrow's team meeting.

reefdog:
@jasonaowen Since I don't think there's an urgent rush for it, I'll wait until after the meeting in case someone raises a blocking objection. Assuming none, you can merge it after the meeting or I'll do it Wed morning.

jasonaowen:
@reefdog license blessings received in #37! Please merge when ready. :rocket:

PR #42, opened 2023-03-22
Clean up CSS and markup scaffolding

reefdog:
This PR contains an assortment of small and mostly-unrelated fixes that preface two more PRs coming shortly. There is no connecting thread here aside from being small improvements that don't belong in the other PRs.

**Testing:**
- Nothing should really look different before and after this PR. It's all underlying improvements.

PR #43, opened 2023-03-22
Add Panel and Table components

reefdog:
This PR adds an assortment of `Panel` and `Table` components that will be used in a forthcoming PR.

This takes a pretty opinionated approach of essentially componentizing _everything_, even if the component does nothing more than generate a single element. This is because I expect we'll want to use more React internals soon and I'd prefer to have just defined everything as components already. I'd be happy to reverse this approach if it proved necessary.

There are unfortunately no stories associated with this PR, but a forthcoming PR adds a story that utilizes these components.

Addresses #12


jasonaowen:
Rebased onto `main` after merging #42! Now to review.

Oh hey and that triggered Netlify; nice!

jasonaowen (approved):
Thank you, @reefdog!

I found it a bit challenging to review this in isolation; it only really made since when viewed in the context of #44. But, since I need to add a bunch of functionality to #44 (as planned!), I don't think it makes sense to combine this PR with that one, or to wait to merge this PR.

> This takes a pretty opinionated approach of essentially componentizing everything, even if the component does nothing more than generate a single element. This is because I expect we'll want to use more React internals soon and I'd prefer to have just defined everything as components already. I'd be happy to reverse this approach if it proved necessary.

We'd previously discussed this, and I was and am +1 to this approach!

> There are unfortunately no stories associated with this PR, but a forthcoming PR adds a story that utilizes these components.

This is the drawback to the very-componentized approach; I took a pass at writing some stories, but these components in isolation turned out not to be very useful, and so the stories fell pretty flat.

As you say, we might go back and refactor these, but for now this is a useful foundation!

Issue #19, opened 2023-02-28
Create proposal detail page


reefdog:
Users should be able to visit a detail page for an individual proposal, listing all the proposal's canonical fields and more recent values:

<img width="1512" alt="PDC Data Viewer 02 - Proposal" src="https://user-images.githubusercontent.com/4731/220751450-d328b240-42f7-41fe-a1dd-887f3318edd9.png">

The page should at least extract the proposal title, applicant name/EID, and data source into the pre-table UI. Also, a toggle for word-wrapping long cells should be present (see multi-bar icon in the right of the Version 4 header).

The proposal list in the sidebar will be the same state and content as the [table view](https://github.com/PhilanthropyDataCommons/data-viewer/issues/14), but condensed to the sidebar. (I may split this off to a separate task.)

Several features shown in the mockup will be delivered as separate tasks, e.g. [version history](https://github.com/PhilanthropyDataCommons/data-viewer/issues/22) and [key](https://github.com/PhilanthropyDataCommons/data-viewer/issues/20)/[source](https://github.com/PhilanthropyDataCommons/data-viewer/issues/21) toggles.

reefdog:
@jasonaowen I would advocate for closing this issue and opening up any remaining tasks and features as separate issues.

jasonaowen:
I agree!

PR #44, opened 2023-03-22
Create Proposal Detail stub

reefdog:
_This PR ladders off #43, which ladders off #42. It is expected that #42 and #43 will be merged into their bases before this one is._

This PR adds a very early but functional version of the Proposal Detail page that can be wired up to the API. It adds two things:

- A `ProposalTable` component that combines the `Table` components from #43 into a coherent interface (with some bugs and limitations that will be addressed). This also includes a Storybook story, but it's quite minimal until more functionality is completed.
- A Proposal Detail page (`GetProposal`) that renders the `ProposalTable` into a `Panel`. I imagine @jasonaowen will want to change quite a bit about this page.

proposal-detail-page

Issue #19

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { Route, Routes } from 'react-router-dom';
 import { Header } from './components/Header';
 import { NotFound } from './pages/NotFound';
+import { GetProposal } from './pages/GetProposal';
 import { Placeholder } from './pages/Placeholder';
 import './App.css';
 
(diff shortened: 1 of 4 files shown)


jasonaowen:
Rebased onto `main` after merging #43.

jasonaowen:
This has been incredibly helpful - thank you so much, @reefdog! I have been using this as a basis for my work, but I think I'm going to organize it a bit differently. In particular, I'm going to (temporarily) leave out some of the design elements that are present here, such as the various buttons. For now, I'm going to close this without merging, so that I can come back later and use those elements.

Issue #15, opened 2023-02-28
Implement pagination on proposal lists


reefdog:
We need to paginate the lists of proposals. This pagination should be available/usable both in the table view and sidebar card views. It will reflect the state of any applied filter on the list.

<img src="https://user-images.githubusercontent.com/4731/221948360-1d9a40dc-7cec-4fb4-8fd5-35b41e4be183.png" width="400" alt="Paginator" />

This pagination should, of course, be abstract enough for other entity lists.

reefdog:
The draft version of the Paginator component (#98) wasn't informed by the attributes that the API actually expects: `_page` (which page do we want?) and `_count` (what are the maximum number of items that should be on that page?).

Also, @jasonaowen has suggested that rather than the previous/next buttons being a `<Button>` component that maps clicks to a navigation/load function, that we actually use a `<Link>` component and lean on native routing.

I need to take a second look at the Paginator component to make sure it's built with a compatible internal logic as the API surface area (not that it _has_ to be that way, but to reduce our own cognitive friction) and see if the `<Link>` approach will work.

slifty:
Per #988 we aren't going to be supporting this functionality.

Issue #14, opened 2023-02-28
Create proposals table page


reefdog:
We need to display a table of proposals to users that approximates this mockup:

<img width="1512" alt="PDC Data Viewer 01 - Proposals Table" src="https://user-images.githubusercontent.com/4731/220751445-2701db4c-16d0-4e3b-83ce-df2b575f6f57.png">

We should feel free to break this task into sub-tasks if necessary. We have already broken out follow-up tasks for [pagination](https://github.com/PhilanthropyDataCommons/data-viewer/issues/15), [sorting](https://github.com/PhilanthropyDataCommons/data-viewer/issues/16), [filtering](https://github.com/PhilanthropyDataCommons/data-viewer/issues/17), or [column customization](https://github.com/PhilanthropyDataCommons/data-viewer/issues/18).

This task depends on #13 for choosing which columns (canonical fields) and in which order we should display.

PR #46, opened 2023-03-30
Reference: Proposals table list panel and paginator

reefdog:
This PR is really too large and may be considered a WIP for plucking and restructuring. It also slips in the Paginator component and its story, but stops short of showing how to use it.

Notes:
- `ProposalListTablePanel` is the primary component which will be dropped onto the new proposals table page itself. It has its own story so you can see how it renders.
   - In the current iteration, there are no buttons in the panel header, because all that functionality (filtering, searching, columns) will come in future work. So for now, it just has a title.
- `ProposalListTable` is the main component that drives `ProposalListTablePanel`. Right now there's some redundancy between them, but I expect that will reduce as we feed it with real data. Or, maybe not!
- `Paginator` is here! I'm taking a bit of a stab at how we want to structure this, but I think it's correct: pagination is generated programmatically based on the number of total items and the number of items per page. "Current page" is assumed to be set in state outside of the paginator itself and passed in via a prop, which means we also need to pass in the state-updating method. Or maybe there's another way of doing that these days, I can't recall if the data store is universal (ala Redux) or not, but I don't think so?
- This commit adds the ability to click on a row and go somewhere. Right now, it just `console.log`s, because I'm not sure what the path to the detail page will be.
- Relatedly, I'm completely guessing that we'll have a stable proposal ID (as in, internal to the PDC) on each `proposal` object, and am guessing its fieldName will be `"id"`. We can change.

**Testing paginator:** In `ProposalListTablePanel.tsx`, add the paginator to the `<PanelBody>` so that it looks like so:

```html
<PanelBody>
  <PaginatorWrapper>
    <Paginator
      itemsPerPage={100}
      totalItems={950}
      currentPage={1}
      setCurrentPage={(page) => { console.log(page); }}
    />
  </PaginatorWrapper>
  <ProposalListTable
    fields={fields}
    proposals={proposals}
  />
</PanelBody>
```

Here's how the `ProposalListTablePanel` story would look if you made that change:
Screenshot of a proposal list table panel

Affects #14 
Affects #15

### DIFF
--- a/src/components/Paginator.tsx
+++ b/src/components/Paginator.tsx
@@ -1,0 +1,153 @@
+import React from 'react';
+import { ArrowLeftIcon, ArrowRightIcon } from '@heroicons/react/24/solid';
+import './Paginator.css';
+
+interface PaginatorPage {
+  pageNumber: number,
+  firstItem: number,
+  lastItem: number,
+}
+
+// Utility method that takes the total items and items per page and returns an array
+// of pages with the page number and boundaries.
+const makePages = (totalItems: number, itemsPerPage: number) => {
+  const pages: PaginatorPage[] = [];
+
+  const pageNumbers = Math.ceil(totalItems / itemsPerPage);
+  for (let i = 1; i <= pageNumbers; i += 1) {
+    const previousPageEnd = (i - 1) * itemsPerPage;
+    pages.push({
+      pageNumber: i,
+      firstItem: previousPageEnd + 1,
+      lastItem: Math.min(i * itemsPerPage, totalItems),
+    });
+  }
+
+  return pages;
+};
+
+// Utility method that takes a `PaginatorPage` and returns a range string,
+// e.g. "1–100" or "201–300". If the range is just a single value, returns just that.
+function formatPageBoundaries(page: PaginatorPage) {
+  if (page.firstItem === page.lastItem) {
+    return page.firstItem.toLocaleString();
+  }
+
+  return `${page.firstItem.toLocaleString()}–${page.lastItem.toLocaleString()}`;
+}
+
+interface PaginatorButtonProps {
+  direction: 'previous' | 'next',
+  onClick: () => void;
+  disabled?: boolean,
+}
+
+const PaginatorButton = ({
+  direction,
+  onClick,
+  disabled = false,
+}: PaginatorButtonProps) => (
+  <button
+    className="paginator-button"
+    disabled={disabled}
+    onClick={onClick}
+    type="button"
+  >
+    {direction === 'previous'
+      ? <ArrowLeftIcon className="icon" />
+      : <ArrowRightIcon className="icon" />}
+  </button>
+);
+
+interface PaginatorProps {
+  itemsPerPage: number,
+  totalItems: number,
+  /**
+   * Externally-managed current page.
+   * Note that there is no protection against this being set
+   * to a value larger than `totalItems / itemsPerPage`.
+   */
+  currentPage: number,
+  /**
+   * Function that will update the `currentPage`.
+   * Assumed to be a state hook, but not required.
+   * All that's required is that the external component
+   * change the `currentPage` property above.
+   */
+  setCurrentPage: (page: number) => void,
+}
+
+/**
+ * A component for navigating between pages of a set of items.
+ * Calculates the number of pages and the range start/end values
+ * based on the `totalItems` and `itemsPerPage` props.
+ * Note that "current page" state should be managed externally,
+ * so changing the `currentPage` prop internally has no effect.
+ */
+export const Paginator = ({
+  itemsPerPage,
+  totalItems,
+  currentPage,
+  setCurrentPage,
+}: PaginatorProps) => {
+  const pages = makePages(totalItems, itemsPerPage);
+
+  const totalPages = pages.length;
+
+  const goToPreviousPage = () => {
+    setCurrentPage(Math.min(totalPages, Math.max(1, currentPage - 1)));
+  };
+
+  const goToNextPage = () => {
+    setCurrentPage(Math.max(1, Math.min(totalPages, currentPage + 1)));
+  };
+
+  const handleSelectorChange = (event: React.FormEvent) => {
+    const selector = event.currentTarget as HTMLSelectElement;
+    setCurrentPage(Number(selector.value));
+  };
+
+  return (
+    <div className="paginator">
+      <PaginatorButton
+        direction="previous"
+        onClick={goToPreviousPage}
+        disabled={currentPage <= 1}
+      />
+      <div className="paginator-content">
+        <div className="paginator-status">
+          {'Showing '}
+          {totalPages > 1 ? (
+            <select
+              onChange={handleSelectorChange}
+              value={currentPage}
+              className="paginator__page-selector"
+            >
+              {pages.map((page) => (
+                <option
+                  key={page.pageNumber}
+                  value={page.pageNumber}
+                >
+                  {formatPageBoundaries(page)}
+                </option>
+              ))}
+            </select>
+          ) : (
+            <span className="paginator-status__value">
+              {formatPageBoundaries(pages[0])}
+            </span>
+          )}
+          {' of '}
+          <span className="paginator-status__value">
+            {totalItems.toLocaleString()}
+          </span>
+        </div>
+      </div>
+      <PaginatorButton
+        direction="next"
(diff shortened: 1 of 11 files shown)


src/components/ProposalListTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+    shortCode: string,
+    label: string,
+  }[],
+  // Note the subtle difference from `proposals` on `ProposalListTableProps`:
+  // this is a single proposal, not an array of them.
+  proposal: {
+    fieldName: string,
+    value: string,
+  }[],
+  proposalId: string,
```
jasonaowen, line 25: > There is no expectation that the proposal’s internal order matches the column (aka field) order, and so there are some potentially expensive iterations in this implementation. We may be able to optimize these if we can prepare the data by putting it in desired order before passing it in.

@reefdog and I discussed this on a call! To recap, I think we should definitely do some data processing to get to the desired format. We should also delegate the responsibility of deciding what whether and in what order to show columns to this display component, while parsing the API response and transforming the data to the expected format should be the responsibility of the page component.

In this case, I think the data might look something like:

```
fieldNames: {
  [shortCode: string]: string;
}
values: {
  [shortCode: string]: string[];
}
```
...that is, we'd pass in two objects: one with the names of canonical fields, and one with arrays of values for those fields, both keyed by the canonical field short codes.

This kind of consolidation is possible here where it isn't on the proposal detail because we don't care about field order in a table with multiple proposals from multiple forms; there does not seem to be an obvious order available.

src/components/ProposalListTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+    <TableRow onClick={handleRowClick}>
+      {/*
+        Iterate over fields to ensure order. Alternatively, if we can guarantee the proposal
+        field order matches the order of `fields`, we can do away with this and just iterate
+        over proposal fields sequentially.
+      */}
+      {fields.map((field) => {
+        // Note the two assumptions baked in here (also mentioned in the props design above):
+        // 1. Proposal `fieldName` corresponds to field `shortCode`
+        // 2. Each proposal has only one value for each matching field
```
jasonaowen, line 49: I think we'll want to assume there can be multiple (or zero) values for a given field, and just always expect a list.

src/components/ProposalListTable.tsx
```diff
@@ -0,0 +1,116 @@
...
+        );
+      })}
+    </TableRow>
+  );
+};
+
+interface ProposalListTableProps {
+  /**
+   * An array of the fields that should be shown on this table.
+   * These correspond to "columns", but we're using a more semantic name.
```
jasonaowen, line 64: On the name specifically: you're right to accept "fields" from the page component, but I think we'll want to convert that to columns in this component.

jasonaowen (commented):
This is looking good so far! I haven't had a chance to dig into the paginator yet, and I'll do a deeper dive and start wiring this up tomorrow.

src/components/ProposalListTable.tsx
reefdog: @jasonaowen lol, yes, absolutely. I clumsily was trying to say "here are the two _incorrect_ assumptions baked in to this implementation". Your suggested format above, with values as an array, is great and easily resolves this limitation.

src/components/ProposalListTable.tsx
reefdog: @jasonaowen I think I love that suggestion, but let me make sure I understand it with a pseudo-implementation:

```ts
const fieldNames = {
  [shortCode: string]: string;
}

const proposals = {
  [shortCode: string]: string[];
}[]
```

Right? This is, I think, the correct way to handle the dictionary of field names, being able to access proposal data via a key, and handling multiple values per proposal field.

> …we don't care about field order in a table with multiple proposals from multiple forms; there does not seem to be an obvious order available.

I read this as saying "column order doesn't matter", but it really does! We very much want to arrange the columns in a specific order (which, by the way, will initially involve us hardcoding some shortcodes into the front-end) and eventually by users rearranging them.

But I think that's compatible with your design above, just by adding a `columnOrder` array of shortCodes, the order of which determines the column order. Since we can now access individual fieldNames and proposal attributes using the shortCode key, we're relieved of some performance problems. (Could probably relieve even more by having proposals themselves be arrays of values with no shortCode keys and rearranged as the user changes the column order, but that is premature now and possibly diminishing returns later.)

reefdog: @jasonaowen After re-reading your comment entirely, I realize at the top you directly mention the column ordering, so I definitely misread the bit I quoted. Was that a reference to how field ordering doesn't matter on the detail page?

reefdog:
@jasonaowen I rewrote this somewhat substantially based on our conversation on the phone and the comments above. I think this is a better design, now. It's certainly less code (and cleaner). Hopefully this is more in line with what you were thinking?

src/components/ProposalListTable.tsx
jasonaowen: > let me make sure I understand it with a pseudo-implementation: ...

I think you're right; I think I commented on the interface definition for a particular row, but then mixed in the field names which is entirely outside the proposal data! The `proposals` data will need to contain a bit more than that, but I'll comment on the new code rather than get into that here.

> > …we don't care about field order in a table with multiple proposals from multiple forms; there does not seem to be an obvious order available.
> 
> I read this as saying "column order doesn't matter", but it really does! 

Ah sorry, I was referring to the `position` attribute of application form fields. My point is that because we are showing proposals for multiple application forms, I don't think we can order the columns by that `position` attribute, because each application form can and likely will have a different list of fields with different `position`s. Picking any particular application form's field order will necessarily mean showing the fields of proposals to other application forms out-of-order, and that's even before considering that a single application form can have multiple canonical fields with different positions.

I think we're on the same page, though, as you go on to say:

> We very much want to arrange the columns in a specific order (which, by the way, will initially involve us hardcoding some shortCodes into the front-end) and eventually by users rearranging them. But I may be misreading what you wrote because I think you're aware of the column order need.

Yep! Hardcoding is a great first pass, and allowing users to rearrange (and select?) them is exactly what I was imagining and I think we discussed yesterday.

> In any case, though, I think that's still compatible with your design above, just by adding a `columnOrder` array of shortCodes, the order of which determines the column order:
> 
> ```ts
> columnOrder: string[] // Which strings are just `shortCode`s
> ```

Yes, although I would suggest that the column order is the responsibility of the display component, not the API component. (I really need to come up with a better name for these; maybe this is the MVC pattern, with the model being the props, the view being the display component, and the controller being the API component?) That is, the component that parses the API response and transforms the data into what the display component wants should not know or care about the column order - particularly once we add support for re-ordering columns. That does mean the display component will eventually have some state and logic, just none relating to the passed-in data.

> Since we can now access individual fieldNames and proposal attributes using the shortCode key, we're relieved of some performance problems. (Could probably relieve even more by having proposals themselves be arrays of values with no shortCode keys and rearranged as the user changes the column order, but that is premature now and possibly diminishing returns later.)

Yeah, that sounds like an optimization we can consider once we have confirmed we have a performance problem. I'd prefer to start as clearly as possible!

jasonaowen (commented):
Calling it for the day - sorry I didn't make more progress on this today!

jasonaowen:
> @jasonaowen I rewrote this somewhat substantially based on our conversation on the phone and the comments above. I think this is a better design, now. It's certainly less code (and cleaner). Hopefully this is more in line with what you were thinking?

It is - great work!

src/components/Paginator.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+const PaginatorButton = ({
+  direction,
+  onClick,
+  disabled = false,
+}: PaginatorButtonProps) => (
+  <button
+    className="paginator-button"
+    disabled={disabled}
+    onClick={onClick}
```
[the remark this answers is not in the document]
reefdog: Oh, very good call! Will change.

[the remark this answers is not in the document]
reefdog: So, this is the correct change, but I want to wait to do it until we're actually implementing the paginator functionality. I realized I was guessing at a few things (such as the math around when to disable the link) without knowing which values will actually be available to me, and where.

So I'll hold off for now. But, good change!

reefdog:
Most of the work here has been implemented, yay! The paginator work is being handled on its own separately (#98). This PR can be closed and will soon have its branch deleted. (@jasonaowen see anything in the commit history that we haven't used?)

PR #47, opened 2023-03-31
Show proposal details

jasonaowen:
Add a new proposal detail page at `/proposal/:proposalId` that will fetch and display the details of the most recent version of that proposal. Visiting that page will automatically start the log in process if you are not yet logged in.

This is the first example of a pattern I'm proposing of separating display components (such as `<ProposalDetailPanel>`) from logic components that make API requests, transform API responses into the format display components expect, and so on. This separation will, I hope, make it easier to prototype new components, play with components in the storybook, and disentangle display logic from business logic. (This PR does not introduce any display logic, but an example would be the button that toggles between showing canonical field names and short codes; I'd expect the display components to accept all the data they'd need, and then internally decide what to display.)

Note that to use this locally, you will need to define the new `REACT_APP_API_URL` environment variable.

This draws heavily from PR #44, but leaves some things behind (such as buttons that are not yet implemented) and adds functionality.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -17,3 +17,7 @@ REACT_APP_OIDC_AUTHORITY=https://auth.example.com/realms/example
 
 # The client ID is assigned by the IdP
 REACT_APP_OIDC_CLIENT_ID=example-data-viewer
+
+# The API URL is the location of a running PhilanthropyDataCommons/service instance
+# https://github.com/PhilanthropyDataCommons/service
+REACT_APP_API_URL=https://api.example.com
--- a/src/components/ProposalDetailPanel.tsx
+++ b/src/components/ProposalDetailPanel.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { BuildingLibraryIcon, DocumentIcon } from '@heroicons/react/24/solid';
+import { DocumentTextIcon } from '@heroicons/react/24/solid';
 import {
   Panel,
   PanelBody,
@@ -38,7 +38,7 @@ const ProposalDetailPanel = ({
         </PanelTitle>
         <PanelTitleTags>
           {applicantId && (
-            <PanelTag icon={<BuildingLibraryIcon />}>
+            <PanelTag badge="EIN">
               {applicantId}
             </PanelTag>
           )}
--- a/src/components/ProposalDetailPanel.tsx
+++ b/src/components/ProposalDetailPanel.tsx
@@ -38,14 +38,12 @@ const ProposalDetailPanel = ({
         </PanelTitle>
         <PanelTitleTags>
           {applicantId && (
-            <PanelTag>
-              <BuildingLibraryIcon className="icon" />
+            <PanelTag icon={<BuildingLibraryIcon />}>
               {applicantId}
             </PanelTag>
           )}
           {title && (
-            <PanelTag>
-              <DocumentIcon className="icon" />
+            <PanelTag icon={<DocumentTextIcon />}>
               {title}
             </PanelTag>
           )}
--- a/src/components/ProposalDetailPanel.tsx
+++ b/src/components/ProposalDetailPanel.tsx
@@ -1,0 +1,64 @@
+import React from 'react';
+import { BuildingLibraryIcon, DocumentIcon } from '@heroicons/react/24/solid';
+import {
+  Panel,
+  PanelBody,
+  PanelHeader,
+  PanelTag,
+  PanelTitle,
+  PanelTitleTags,
+  PanelTitleWrapper,
+} from './Panel';
+import { ProposalTable } from './ProposalTable';
+
+interface ProposalDetailPanelProps {
+  title: string | undefined;
+  applicant: string;
+  applicantId: string | undefined;
+  version: number;
+  values: {
+    fieldName: string,
+    position: number,
+    value: string,
+  }[],
+}
+
+const ProposalDetailPanel = ({
+  title,
+  applicant,
+  applicantId,
+  version,
+  values,
+}: ProposalDetailPanelProps) => (
+  <Panel>
+    <PanelHeader>
+      <PanelTitleWrapper>
+        <PanelTitle>
+          {applicant}
+        </PanelTitle>
+        <PanelTitleTags>
+          {applicantId && (
+            <PanelTag>
+              <BuildingLibraryIcon className="icon" />
+              {applicantId}
+            </PanelTag>
+          )}
+          {title && (
+            <PanelTag>
+              <DocumentIcon className="icon" />
+              {title}
+            </PanelTag>
+          )}
+        </PanelTitleTags>
+      </PanelTitleWrapper>
+    </PanelHeader>
+    <PanelBody>
+      <ProposalTable
+        version={version}
+        values={values}
+      />
+    </PanelBody>
+  </Panel>
+);
+
+export { ProposalDetailPanel };
--- a/src/components/ProposalTable.tsx
+++ b/src/components/ProposalTable.tsx
@@ -1,0 +1,49 @@
+import React from 'react';
+import {
+  Table,
+  TableHead,
+  ColumnHead,
+  TableBody,
+  TableRow,
+  RowHead,
+  RowCell,
+} from './Table';
+
+interface ProposalTableProps {
+  version: number;
+  values: {
+    fieldName: string,
+    position: number,
+    value: string,
+  }[],
+}
+
+export const ProposalTable = ({
+  values,
+  version,
+}: ProposalTableProps) => (
+  <Table>
+    <TableHead fixed>
+      <TableRow>
+        <ColumnHead>
+          Canonical Field
+        </ColumnHead>
+        <ColumnHead>
+          {`Version ${version}`}
+        </ColumnHead>
+      </TableRow>
+    </TableHead>
+    <TableBody>
+      {values.map(({ fieldName, position, value }) => (
+        <TableRow key={position}>
+          <RowHead>
+            {fieldName}
+          </RowHead>
+          <RowCell>
+            {value}
+          </RowCell>
+        </TableRow>
+      ))}
+    </TableBody>
+  </Table>
+);
--- a/public/index.html
+++ b/public/index.html
@@ -24,7 +24,7 @@
       work correctly both with client-side routing and a non-root public URL.
       Learn how to configure a non-root public URL by running `npm run build`.
     -->
-    <title>React App</title>
+    <title>Philanthropy Data Commons Data Viewer</title>
   </head>
   <body>
     <noscript>You need to enable JavaScript to run this app.</noscript>
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { Route, Routes } from 'react-router-dom';
 import { Header } from './components/Header';
 import { NotFound } from './pages/NotFound';
+import { ProposalDetail } from './pages/ProposalDetail';
 import { Placeholder } from './pages/Placeholder';
 import './App.css';
 
@@ -11,6 +12,7 @@ const App = () => (
     <main className="App-main">
       <Routes>
         <Route path="/" element={<Placeholder />} />
+        <Route path="/proposal/:proposalId" element={<ProposalDetail />} />
         <Route path="*" element={<NotFound />} />
       </Routes>
     </main>
--- a/src/components/Panel/Panel.css
+++ b/src/components/Panel/Panel.css
@@ -61,6 +61,10 @@
   height: var(--icon-size--text);
 }
 
+.panel-tag .badge {
+  --badge--background-color: var(--color--gray--medium);
+}
+
 .panel-actions {
   display: flex;
   gap: var(--fixed-spacing--1x);
--- a/src/components/Panel/PanelTag.tsx
+++ b/src/components/Panel/PanelTag.tsx
@@ -2,14 +2,17 @@ import React from 'react';
 
 interface PanelTagProps {
   children: React.ReactNode;
+  icon?: React.ReactElement | null;
   className?: string;
 }
 
 export const PanelTag = ({
   children,
+  icon = undefined,
   className = '',
 }: PanelTagProps) => (
   <div className={`panel-tag ${className}`.trim()}>
+    {icon ? React.cloneElement(icon, { className: 'icon' }) : null}
     {children}
   </div>
 );
--- a/src/components/Panel/PanelTag.tsx
+++ b/src/components/Panel/PanelTag.tsx
@@ -3,16 +3,19 @@ import React from 'react';
 interface PanelTagProps {
   children: React.ReactNode;
   icon?: React.ReactElement | null;
+  badge?: string | null;
   className?: string;
 }
 
 export const PanelTag = ({
   children,
   icon = undefined,
+  badge = undefined,
   className = '',
 }: PanelTagProps) => (
   <div className={`panel-tag ${className}`.trim()}>
     {icon ? React.cloneElement(icon, { className: 'icon' }) : null}
+    {badge ? <span className="badge">{badge}</span> : null}
     {children}
   </div>
 );
--- a/src/components/Table/Table.css
+++ b/src/components/Table/Table.css
@@ -17,6 +17,10 @@
   white-space: nowrap;
 }
 
+.table tbody th {
+  white-space: nowrap;
+}
+
 .table th,
 .table td {
   text-align: left;
--- a/src/index.css
+++ b/src/index.css
@@ -30,6 +30,9 @@
   --transition-duration: 200ms;
 
   --icon-size--text: 1.25em;
+
+  --badge--background-color: var(--color--gray--darker);
+  --badge--color: white;
 }
 
 body {
@@ -77,4 +80,12 @@ h6 {
   padding: 0;
   font-weight: var(--font-weight--normal);
   line-height: 1.2;
+}
+
+.badge {
+  background-color: var(--badge--background-color);
+  color: var(--badge--color);
+  font-size: 0.8em;
+  padding: 0 0.25em;
+  border-radius: 2px;
 }
--- a/src/pages/ProposalDetail.tsx
+++ b/src/pages/ProposalDetail.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { OidcSecure } from '@axa-fr/react-oidc';
 import { useParams } from 'react-router-dom';
 import {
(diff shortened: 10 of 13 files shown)


src/pages/ProposalDetailPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const title = getValueOfCanonicalField(canonicalFields, proposal, 'proposal_name')
+    ?? getValueOfCanonicalField(canonicalFields, proposal, 'proposal_title')
+    ?? `Proposal ${proposalId}`;
+  const author = getValueOfCanonicalField(canonicalFields, proposal, 'organization_name')
+    ?? getValueOfCanonicalField(canonicalFields, proposal, 'organization_dba_name')
+    ?? getValueOfCanonicalField(canonicalFields, proposal, 'organization_legal_name')
+    ?? getValueOfCanonicalField(canonicalFields, proposal, 'proposal_primary_contact_name')
+    ?? getValueOfCanonicalField(canonicalFields, proposal, 'proposal_submitter_name')
+    ?? 'Unknown Author';
+  const authorId = getValueOfCanonicalField(canonicalFields, proposal, 'organization_tax_id');
```
jasonaowen, line 85: @jim-mcgowan I'd appreciate your thoughts on this priority list for picking canonical fields to fill the header! In case the code is unclear, it picks the first present value, so for title it'd look for `proposal_name`, then if that's absent `proposal_title`, then if that's absent it'll construct a name like `Proposal 7`.

jasonaowen:
I added the environment variable to Netlify, and added the deploy preview to the Keycloak client configuration.

A few example pages:
- [Proposal 5](https://deploy-preview-47--philanthropy-data-commons-viewer.netlify.app/proposal/5) has all the information we're looking for
- [Proposal 7](https://deploy-preview-47--philanthropy-data-commons-viewer.netlify.app/proposal/7) has no title
- [Proposal 0](https://deploy-preview-47--philanthropy-data-commons-viewer.netlify.app/proposal/0) does not exist, and will stay on the loading screen forever, as mentioned in the PR description
- The [Proposal Detail Storybook](https://deploy-preview-47--philanthropy-data-commons-viewer.netlify.app/storybook/?path=/story/stories-proposaldetail--default) does not require logging in, and allows you to play with the component

src/components/ProposalTable.tsx
```diff
@@ -0,0 +1,49 @@
...
+          Canonical Field
+        </ColumnHead>
+        <ColumnHead>
+          {`Version ${version}`}
+        </ColumnHead>
+      </TableRow>
+    </TableHead>
+    <TableBody>
+      {values.map(({ fieldName, position, value }) => (
+        <TableRow key={position}>
```
reefdog, line 38: Ah, is `position` actually unique in each dataset (e.g. safe to use for `key` on its own rather than as a string like `${fieldName}-${position}`)? I assumed it was relative to other fields with the same fieldName. I assume you're right, but just flagging it in case.

src/pages/ProposalDetailPage.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const values = mapCanonicalFields(canonicalFields, proposal);
+
+  return (
+    <OidcSecure>
+      <ProposalDetailPanel
+        title={title}
+        author={author}
+        authorId={authorId}
+        version={version}
+        values={values}
```
reefdog, line 96: I'm a little nervous at how quickly we're going to want to replace this architecture when we really support versions, but you noted that limitation in the description and I think it's the right choice. Interested to know if you already have the adaptations in mind!

src/use-pdc-api.ts
```diff
@@ -0,0 +1,36 @@
...
+      .then((res) => res.json())
+      .then(setResponse)
+      .catch((error: unknown) => logger.error({ error }, `Error fetching ${path}`));
+
+    /* eslint-disable-next-line react-hooks/exhaustive-deps --
+     * fetch ought to be a dependency, but including it causes
+     * this effect to repeat forever
+     * https://github.com/AxaFrance/react-oidc/issues/1023
+     */
+  }, [path, setResponse]);
```
reefdog, line 31: I _think_ this implementation misuses effect dependencies.

This is treating them like function parameters, but the `useEffect` hook has access to the outer scope and doesn't need that (which is why `fetch` works). These are dependencies that should trigger the effect to run again when they change (which is why including `fetch` in the dependencies causes it to repeatedly re-run).

Assuming we only want this to run on initial render, we still want the dependency array, but I think it should be empty, right?

```suggestion
  }, []);
```

I may be totally misunderstanding what's going on here, but I don't quite follow as-is.

reefdog (changes requested):
This is wonderful work! I'm not currently able to run it locally because I just realized I don't know how to authenticate into the PDC from localhost (or if that's even possible), but looking at the deployment preview it's a great start.

I do have a scattering of individual comments and requests. If you resolve them to your satisfaction, you can feel free to merge, or I can take a second look on Monday.

jim-mcgowan:
@jasonaowen I need to see this to be able to comment.

.env.example
```diff
@@ -17,3 +17,8 @@ REACT_APP_OIDC_AUTHORITY=https://auth.example.com/realms/example
 
 # The client ID is assigned by the IdP
 REACT_APP_OIDC_CLIENT_ID=example-data-viewer
+
+# The API URL is the location of a running PhilanthropyDataCommons/service instance
+# https://github.com/PhilanthropyDataCommons/service
+#REACT_APP_API_URL=http://localhost:3000
+REACT_APP_API_URL=https://api.philanthropydatacommons.org
```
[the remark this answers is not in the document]
jasonaowen: I did, but I'm very open to the feedback that I should stick to the existing pattern - I will fix!

src/pages/ProposalDetailPage.tsx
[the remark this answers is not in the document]
jasonaowen: I did consider writing a helper function, like "get value of first present field" or so, but it's just a few lines of code and this seems clear enough for now.

...but I bet we could do it with a map, a filter, and maybe a reduce!

src/pages/ProposalDetailPage.tsx
jasonaowen: I do! I'm sure what I have in mind will change a bit as I implement it, but I think I have a good enough plan. And yeah these'll change pretty quick, but I think that's okay!

src/components/ProposalDetailPanel.tsx
```diff
@@ -0,0 +1,57 @@
...
+  PanelTitle,
+  PanelTitleTags,
+  PanelTitleWrapper,
+} from './Panel';
+import { ProposalTable } from './ProposalTable';
+
+interface ProposalDetailPanelProps {
+  title: string;
+  author: string;
+  authorId: string | undefined;
```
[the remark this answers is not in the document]
jasonaowen: That sounds like a better name, thank you! I will fix.

src/use-pdc-api.ts
jasonaowen: Including `setResponse` is probably being over-cautious, yes. But I think we do need to include `path`, because otherwise we can't navigate without a full page reload.

I tried it by adding prev/next buttons, and removing `path` from the dependency array, and navigating to another proposal did not load any new data. I believe this is because we have the same React component tree, and the only difference React sees is the new react-router path parameter ([via the `useParams()` hook](https://reactrouter.com/en/main/hooks/use-params)). We use that to construct the API relative path, but if we do not include that relative path (the `path` function parameter here) in the dependency list, React looks and says "I don't need to re-run this effect", and so we do not update the data.

I *think* we would ideally want `fetch` to be in the dependency list, too, but I'm not completely certain. I would expect it to trigger the effect to run once its token expires and is refreshed, but that's not how it works currently (see the upstream bu g I filed), and even so that's kinda getting into IdP configuration territory - if we set very short expiration times on our tokens, do we want viewers to continuously be making API calls? Probably not, so maybe the comment is incomplete...

src/pages/ProposalDetailPage.tsx
jasonaowen: I talked with @jim-mcgowan on the phone, and my takeaway was that this is good for now, so resolving.

src/components/ProposalDetailPanel.tsx
```diff
@@ -0,0 +1,57 @@
...
+        <PanelTitle>
+          {title}
+        </PanelTitle>
+        <PanelTitleTags>
+          <PanelTag>
+            <UsersIcon className="icon" />
+            {author}
+            {authorId && ` (${authorId})`}
+          </PanelTag>
+        </PanelTitleTags>
```
jasonaowen, line 45: After talking with @jim-mcgowan, his main feedback was that we should be highlighting the applicant above the proposal title. The EIN is of secondary importance, and should not be given the same weight as the promoted organization name; he suggested displaying it below, but maybe alongside in gray might work too - I'll play with it a bit, or possibly wait for @reefdog's feedback.

src/components/ProposalDetailPanel.tsx
```diff
@@ -0,0 +1,57 @@
...
+}: ProposalDetailPanelProps) => (
+  <Panel>
+    <PanelHeader>
+      <PanelTitleWrapper>
+        <PanelTitle>
+          {title}
+        </PanelTitle>
+        <PanelTitleTags>
+          <PanelTag>
+            <UsersIcon className="icon" />
```
jasonaowen, line 41: @jim-mcgowan was not a fan of this users icon: "I'm not sure what that's trying to convey. Three people would suggest an organization to me."

There is a `user-group` in heroicons; I'll try that.

jasonaowen (commented):
@jim-mcgowan and I spoke on the phone, and he shared some feedback. I believe I've documented all the front end concerns below; there were also some data concerns that I'll document elsewhere.

src/components/ProposalTable.tsx
jasonaowen: Excellent question! I was slightly confident when I committed, but after doing some more reading and thinking, I am now more confident. (One of the better articles I've found was [React key attribute: best practices for performant lists](https://www.developerway.com/posts/react-key-attribute).)

I think the two key (hah) things to think about here is what state items in the list carry, when this list will change. 

Maintaining state across reordering or complete switches of underlying data is important to maintain correct behavior. The example from the article is instructive: if you have the first item in the list checked, and then change the sort, the checkmark should follow the item rather than the index. However, in our case, list items (`<TableRow>`s, here) do not carry any state, so I do not think there is a risk of incorrect behavior no matter what approach we take.

If correctness is not a concern, then that leaves performance. I believe this list will change under two circumstances:  when navigating from one proposal to another or toggling the version history. The version history is simpler: turning it on can only add rows at the end (fields that existed on previous versions of the proposal but not the current version of the proposal), and turning it off again will only delete rows at the end, so we could even use the `index` and be okay.

Switching proposals is more complicated, but reusing positions will allow React to maintain most of the existing DOM structure, changing only the values. Some positions might be missing, and others added (eg a proposal for an application form with field positions `[1, 2, 4]` switching to one with positions `[1, 2, 3, 5]`), and there'd be some performance loss from using `position` as a key rather than `index`, but on the other hand some navigations will be between proposals for the same application form, and in that case performance should be ideal - matching rows will match not only their `position` but also their row header.

src/components/ProposalDetailPanel.tsx
jasonaowen: I don't think this looks great:

Screen Shot 2023-03-31 at 18 44 27

It's not just the alignment, although that certainly doesn't help. I think I'll omit the icon entirely, unless someone speaks up.

src/components/ProposalDetailPanel.tsx
jasonaowen: Here's what I have right now:

Screen Shot 2023-03-31 at 18 54 49

jasonaowen:
Alright, I added several fixup commits to this. I'll squash them away on Monday, but for now hopefully they make seeing the changes I made in response to review comments easier.

jasonaowen:
> I'm not currently able to run it locally because I just realized I don't know how to authenticate into the PDC from localhost (or if that's even possible)

It is possible! You should be able to plug the relevant values into your `.env.local` file; that's what I've been doing! Let's pair next week on getting you set up, @reefdog.

jasonaowen:
> If you resolve them to your satisfaction, you can feel free to merge, or I can take a second look on Monday.

I ended up making some design changes in response to @jim-mcgowan's feedback, so yeah I'd love a second look from you please, @reefdog!

src/pages/ProposalDetailPage.tsx
```diff
@@ -0,0 +1,117 @@
...
+    version: number;
+    fieldValues: {
+      applicationFormField: {
+        canonicalFieldId: number;
+        position: number;
+      };
+      value: string;
+    }[];
+  }[];
+}
```
jasonaowen, line 24: These should live in the API helpers file.

src/pages/ProposalDetailPage.tsx
```diff
@@ -0,0 +1,117 @@
...
+    ?? getValueOfCanonicalField(canonicalFields, proposal, 'proposal_primary_contact_name')
+    ?? getValueOfCanonicalField(canonicalFields, proposal, 'proposal_submitter_name')
+    ?? 'Unknown Applicant'
+);
+
+const ProposalDetailPage = () => {
+  const params = useParams();
+  const proposalId = params.proposalId ?? 'missing';
+  const canonicalFields = usePdcApi<CanonicalField[]>('/canonicalFields');
+  const proposal = usePdcApi<Proposal>(`/proposals/${proposalId}?includeFieldsAndValues=true`);
```
jasonaowen, line 72: Extract helpers, and put them in the API file.

src/use-pdc-api.ts
jasonaowen, line 1: Remove `use` from the file name.

src/pages/ProposalDetailPage.tsx
jasonaowen, line 1: Remove `Page` from the file name - it's already in the path.

jasonaowen (commented):
Before merging, I want to clean up a few things based on what I learned working on #49!

src/components/ProposalTable.tsx
reefdog: Really thorough and instructive explanation, thanks @jasonaowen!

src/pages/ProposalDetailPage.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+  useEffect(() => {
+    if (canonicalFields === null || proposal === null) {
+      document.title = 'Proposal Detail - Philanthropy Data Commons';
+    } else {
+      const applicant = getApplicant(canonicalFields, proposal);
+      document.title = `${applicant} Proposal Detail - Philanthropy Data Commons`;
+    }
+  }, [canonicalFields, proposal]);
```
reefdog, line 81: When you navigate away from a proposal detail page, the title remains as set here.

Ideally, all our pages would define their titles, so this would be immediately overwritten. But we might also consider returning a cleanup function from this hook which resets the title to some default (e.g., "Philanthropy Data Commons") in case the destination page _doesn't_ have a title defined.

The downside to this is (1) the potential for the title to change twice on every navigation away slowly enough for people to notice and (2) the added boilerplate for every page that uses a title. I suspect, however, we'll want to abstract out some of this title-setting anyway.

For here and now, maybe just do the cleanup method? I think that would just mean adding `return () => document.title = 'Philanthropy Data Commons';` to the end of the hook.

src/use-pdc-api.ts
reefdog: Ah, excellent point about `path`, and I agree it should be passed in.

`setResponse` I'm pretty sure doesn't. I read the [React guidance here](https://legacy.reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) which helped me quite a bit, but it was really the [next section on performance](https://legacy.reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often) that demonstrated specifically that the state-setting function doesn't need to be passed in.

As for `fetch`, I assumed that `useOidcFetch()` was simply returning a custom fetch method that decorated the request with the token, and any determination about token freshness occurred when `fetch()` itself was called. But I'm at the edges of my understanding here. If you think the definition of `fetch` itself changes from render to render, then yes, let's keep it in the list (after that upstream bug is solved)! I still don't fully understand, but trust your knowledge of it more than mine.

src/components/ProposalDetailPanel.tsx
reefdog: I agree with both of you, in fact! I too wanted the three-person icon, but ended up not liking it because it looks vaguely like a paw print more than a crowd — at least, I couldn't unthink it after thinking it.

I don't think the big title needs an icon, so I also agree with where you landed. But, as noted in my other comment, I want to take a crack at this header altogether after understanding Jim's feedback better.

I did notice that you used the building library icon for the EIN. I feel like without the organization label also in the same tag, we can't use such an "organization as a building" metaphor. I'm experimenting with just a small badge reading "EIN", which no doubt we'll change once we replace EIN as the most visible organization ID:

Organization header

src/components/ProposalDetailPanel.tsx
```diff
@@ -0,0 +1,62 @@
...
+              {applicantId}
+            </PanelTag>
+          )}
+          {title && (
+            <PanelTag icon={<DocumentTextIcon />}>
+              {title}
+            </PanelTag>
+          )}
+        </PanelTitleTags>
+      </PanelTitleWrapper>
```
reefdog, line 51: I'm pretty uncomfortable with how this incarnation makes the page feel like an Organization page, not a Proposal page. Which I understand is part of the intent, but I think it risks muddying the user's understanding of what entity they're actually looking at here.

For the demo purposes, let's just stick with it, but I think I'm going to take another pass at understanding the feedback and suggesting some more alternatives that don't relegate the proposal title itself to the panel tag area.

reefdog (changes requested):
@jasonaowen This is great and could merge as-is (pending your own noted TODOs).

I did make a few changes myself (one in particular, the icon to badge change for the EIN, you can feel free to discard if you dislike it), and want to take _more_ of a crack at resolving Jim's concerns with the title/tags because I think relegating the proposal title to merely a tag with the org as the headline may confuse users in the long term. Fine for the demo, though, and heck maybe the feedback will be instructive.

A few other notes/requests but all disposable at your discretion. Sorry for the review delay!

src/components/ProposalDetailPanel.tsx
reefdog: (This is more of a musing than anything else, feel free to mark as resolved.)

src/use-pdc-api.ts
jasonaowen: > `setResponse` I'm pretty sure doesn't. I read the [React guidance here](https://legacy.reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) which helped me quite a bit, but it was really the [next section on performance](https://legacy.reactjs.org/docs/hooks-faq.html#what-can-i-do-if-my-effect-dependencies-change-too-often) that demonstrated specifically that the state-setting function doesn't need to be passed in, unless I'm misunderstanding our use of it.

Thank you, I hadn't read closely enough! In particular, the [old docs specifically call out `setState` as not needing to be a dependency](https://legacy.reactjs.org/docs/hooks-reference.html):

> React guarantees that `setState` function identity is stable and won’t change on re-renders. This is why it’s safe to omit from the `useEffect` or `useCallback` dependency list.

I will fix!

> As for `fetch`, I assumed that `useOidcFetch()` was simply returning a custom fetch method that decorated the request with the token, and any determination about token freshness occurred when `fetch()` itself was called. But I'm at the edges of my understanding here. If you think the definition of `fetch` itself changes from render to render, then yes, let's keep it in the list (after that upstream bug is solved)! I still don't fully understand, but trust your knowledge of it more than mine.

Yeah, after thinking further, you're right. `useOidcFetch()` does not depend on any of the state or props we care about; it has its own internal state, but we don't need or want to re-fetch data based on any of its state - updating a token should not cause any API requests! I'll update the comment.

src/components/ProposalDetailPanel.tsx
jasonaowen: Oh, that looks very good!

src/pages/ProposalDetailPage.tsx
jasonaowen: This is a fantastic note - thank you for catching it!

I will do.

src/components/ProposalDetailPanel.tsx
jasonaowen: Ah, thank you for that thoughtfulness - I implemented the feedback I got without considering that! Would you like me to make an issue for following up on that?

reefdog: Jim and I had a meeting about the header today, and I'm trying out a few design ideas in Figma. I'll create an issue for it; what you have here can roll in without further edits.

reefdog (approved):
After a discussion w/@jasonaowen I'm changing my review to an "approve" to unblock his ability to unilaterally merge it in once he's made the edits he's already got queued up and/or in mind. Really happy with where this landed and agree with the edits he has planned! Thanks for all the positive iterating on it.

src/components/ProposalDetailPanel.tsx
reefdog: Oh, hah, I already did make an issue for it! https://github.com/PhilanthropyDataCommons/data-viewer/issues/54

jasonaowen:
Thanks, @reefdog!

I believe I've addressed all the feedback in this PR, and I've squashed down all the fixup commits. The tests pass, the linter is happy, the site builds, and the deploy preview looks good.

Merging!

PR #49, opened 2023-04-04
Proposal list page

### LINKED ISSUE #14: Create proposals table page

jasonaowen:
This PR uses some of changes in #46 to add a list page at `/proposals?page=1&count=20` that will retrieve the list of proposals from the API, transform the data, and display it. Clicking on a row will navigate to the proposal detail page.
    
Currently, the API does not provide the necessary details in the proposal list response, so we have to do an N+1 query where we get the details for each proposal in the list. This can take a long time and be a lot of traffic, but until the API offers more information, it's the only thing to do.
    
This has some pagination support, via the `page` and `count` query parameters, like `/proposals?count=25&page=3`. These parameters are passed along to the API. However, because the API also does not return any information about how many proposals or how many pages there are, we cannot use the `<Paginator>` component created in #46, so instead just set the default `count` parameter to be very high. Again, until the API offers more information, the other alternatives lead to a bad user experience.

For now, we are hardcoding a list of field shortnames to show as columns, but the inner `<ProposalListTable>` does not know or care, so we should be well positioned to allow user customization eventually.

The table has a toggle for whether the contents should be displayed wrapped or truncated, and defaults to showing truncated data.

The table also has a non-functional search box to demonstrate eventual intended functionality.

Resolves #14

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -1,4 +1,6 @@
 import type { Preview } from "@storybook/react";
+import React from 'react';
+import { MemoryRouter } from 'react-router-dom';
 import '@fontsource/source-sans-pro/400.css';
 import '@fontsource/source-sans-pro/600.css';
 import '@fontsource/source-sans-pro/700.css';
@@ -17,6 +19,13 @@ const preview: Preview = {
       },
     },
   },
+  decorators: [
+    (Story) => (
+      <MemoryRouter>
+        <Story />
+      </MemoryRouter>
+    )
+  ]
 };
 
 export default preview;
--- a/src/components/ProposalListTablePanel.tsx
+++ b/src/components/ProposalListTablePanel.tsx
@@ -3,10 +3,10 @@ import {
   Panel,
   PanelHeader,
   PanelBody,
-  PanelTitle,
-  PanelTitleWrapper,
+  PanelActions,
 } from './Panel';
 import { ProposalListTable } from './ProposalListTable';
+import { Button } from './Button';
 
 interface ProposalListTablePanelProps {
   fieldNames: Record<string, string>;
@@ -44,9 +44,16 @@ export const ProposalListTablePanel = ({
 }: ProposalListTablePanelProps) => (
   <Panel>
     <PanelHeader>
-      <PanelTitleWrapper>
-        <PanelTitle>Proposals</PanelTitle>
-      </PanelTitleWrapper>
+      <PanelActions>
+        <input
+          type="text"
+          placeholder="Text to search for…"
+          className="input"
+        />
+        <Button disabled>
+          Search
+        </Button>
+      </PanelActions>
     </PanelHeader>
     <PanelBody>
       <ProposalListTable
--- a/src/components/ProposalListTablePanel.tsx
+++ b/src/components/ProposalListTablePanel.tsx
@@ -1,0 +1,59 @@
+import React from 'react';
+import {
+  Panel,
+  PanelHeader,
+  PanelBody,
+  PanelTitle,
+  PanelTitleWrapper,
+} from './Panel';
+import { ProposalListTable } from './ProposalListTable';
+
+interface ProposalListTablePanelProps {
+  fieldNames: Record<string, string>;
+  proposals: {
+    id: string;
+    values: Record<string, string[]>;
+  }[];
+}
+
+// For now, we are hard-coding this list.
+// In the future, this will be user-configurable.
+const DEFAULT_COLUMNS = [
+  'organization_name',
+  'organization_tax_id',
+  'organization_city',
+  'organization_state_province',
+  'organization_country',
+  'organization_website',
+  // 'organization_mission_statement', // FIXME: Disabled until we have wrap
+  'organization_start_date',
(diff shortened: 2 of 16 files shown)


jasonaowen:
I configured Keycloak to allow login from the preview domain, so you should be able to see the list of proposals at https://deploy-preview-49--philanthropy-data-commons-viewer.netlify.app/proposals/ . No pagination yet, so if you want the full list add `?count=1000` to the end.

This draft is also not ready because
- linter errors
- clicking a row should go to the detail page instead of logging a message to the console
- needs tests for the data transformation functions
- pagination

reefdog:
@jasonaowen Faaaantastic! Thank you for this. I'm going to add a few more commits to this in the WIP vein, but when you're online later we should decide how to organize this and #47.

jasonaowen:
Rebased onto `main` after merging #47.

jasonaowen:
To run through the list right now:

- [ ] this PR description is far too brief
- [ ] commit history desperately needs revision
- [x] builds on un-merged PRs [Reference: Proposals table list panel and paginator #46](https://github.com/PhilanthropyDataCommons/data-viewer/pull/46) and [Show proposal details #47](https://github.com/PhilanthropyDataCommons/data-viewer/pull/47)
  - #47 has been merged, and I think the tentative plan was always not to merge #46; I'm happy to make that The Plan going forward
- [ ] linter errors
- [x] clicking a row should go to the detail page instead of logging a message to the console
  - @reefdog took care of this!
- [ ] needs tests for the data transformation functions
- [x] ~~pagination~~
  - we're going to add pagination in a later PR, because https://github.com/PhilanthropyDataCommons/service/issues/179 is not complete and we can't really do much until it is

reefdog:
@jasonaowen Would you have time to add the `getValueOfCanonicalField` fallback behavior to the proposal list table as well, for the same three effective fields (org name, org EIN, and proposal name)?

src/components/ProposalListTablePanel.tsx
```diff
@@ -0,0 +1,80 @@
...
+
+  return (
+    <Panel>
+      <PanelHeader>
+        <PanelActions>
+          <input
+            type="text"
+            placeholder="Search…"
+            className="input"
+          />
```
jasonaowen, line 58: I know that you were instructed to do this, but it feels Not Great - there's no indicator as to why it won't work, and especially if we're going to be handing out logins soon, I feel like this is likely to be distracting. Can we give it a popup or `disable` it or give it a `disabled` search button or something?

reefdog: Agree all around. I think Karl's actual suggestion was that, upon submitting, we display a "coming soon" sort of message. I just punted on that complexity.

What about a dreaded `alert()` on submit?

I'm hoping to avoid a submit button until I style one I like (I have Ideas™); do you think just enter-to-submit is sufficient UI, or do we need a stand-in submit button?

.storybook/preview.tsx
```diff
@@ -17,6 +19,13 @@ const preview: Preview = {
       },
     },
   },
+  decorators: [
+    (Story) => (
+      <BrowserRouter>
+        <Story />
+      </BrowserRouter>
```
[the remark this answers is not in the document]
reefdog: @jasonaowen Ah, that's why all the examples used it! 😆 I thought I was being so clever.

jasonaowen:
> @jasonaowen Would you have time to add the `getValueOfCanonicalField` fallback chain to the proposal list table as well, for the same three effective fields (org name, org EIN, and proposal name)?

Oof, I'm sorry @reefdog, I don't think I'm going to be able to get that in tonight. Now that that new data is loaded, I see why you want it and it makes sense, but while I think I have an approach it's going to be kinda complicated and I'd rather not do that under time pressure. I opened up #67 to track that.

src/components/ProposalListTablePanel.tsx
jasonaowen: Can you please add that button, then, and squash it into commit 2c425aa? I think that's the last thing this needs!

jasonaowen:
@reefdog okay, I think this is nearly as good as it's going to get! I'd appreciate your help on the search button, but I'll leave that to your discretion - please merge when you're ready!

src/components/ProposalListTablePanel.tsx
reefdog: On it!

reefdog (approved):
It was a hard-fought battle, but this is great stuff. 🎉

PR #59, opened 2023-04-06
Show previous/next proposal buttons in dev mode

jasonaowen:
Navigating by proposal ID is not an interface we want to show to users: there is no semantic meaning to the proposal ID, and it depends on the unsafe and incorrect assumption that there are no gaps in the sequence. Long-term, we plan to build a narrow list on the left side of the detail page that will allow navigating from one proposal to another.

However, developers do not really care about the lack of semantic meaning around proposal IDs; our concerns are more around how the data is displayed than what the data means. For that, quickly navigating through multiple proposals - in any arbitrary but consistent order - is very convenient.

Add previous/next links to the proposal detail page when run in dev mode (`npm run start`). Do not show these links when run in production mode (`npm run build`).

Screen Shot 2023-04-06 at 02 07 55

@reefdog I don't know why one icon is much bigger than the other; if this is revealing an underlying bug/assumption about how `<PanelActions>` are composed, or if it bothers you very much, please feel free to fix it!

### DIFF
--- a/src/components/ProposalDetailPanel.tsx
+++ b/src/components/ProposalDetailPanel.tsx
@@ -1,7 +1,9 @@
 import React from 'react';
-import { DocumentTextIcon } from '@heroicons/react/24/solid';
+import { Link } from 'react-router-dom';
+import { BackwardIcon, DocumentTextIcon, ForwardIcon } from '@heroicons/react/24/solid';
 import {
   Panel,
+  PanelActions,
   PanelBody,
   PanelHeader,
   PanelTag,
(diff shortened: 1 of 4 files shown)


reefdog (approved):
@jasonaowen This is incredibly convenient, thank you! I added two commits:

- Style them like buttons. I'm not sure _why_ you were getting Le Giant Icon but in any case they look better now.
- Show both buttons using a single conditional. Arguably I should've just noted this instead of doing it in case there was a reason you were doing it this way, my apologies. You can override if you'd like.

Anyway, very much approved.

jasonaowen:
Gorgeous!

Screen Shot 2023-04-06 at 11 32 47

Thanks, @reefdog!

I double-checked that the Netlify deploy preview does not render these buttons, and it does not. Merging!

PR #66, opened 2023-04-07
Improve navbar

### LINKED ISSUE #61: Style login/user item in navbar like other links
For Reasons, the Log in button in our navbar doesn't look like the others:

Navbar

It should:

Navbar

(Ignore the Home/Proposals difference.)

### LINKED ISSUE #65: Add Feedback button to navbar
We want to add a "Feedback" icon to the navbar that links to `mailto:info@philanthropydatacommons.org?Subject=Feedback%20on%20the%20Philanthropy%20Data%20Commons%20pilot%20data%20viewer`

reefdog:
This small PR offers two navbar improvements:

- Style the navbar login/user element like the others
   - Arguably, this _shouldn't_ look like the others since it's not clickable. I agree, but it will (soon, hopefully) work as a link to basic user functionality like going to the settings and logging out (probably through a menu). I'm okay with this temporary weirdness. It doesn't have a `cursor:pointer` at least.
- Add a "feedback" item to the navbar for pilot participants to email the team

In the current state:

image

This PR will have conflicts with the #49 currently in flight, but it'll be simple to resolve.

Closes #61 
Closes #65

### DIFF
--- a/src/components/User.tsx
+++ b/src/components/User.tsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import { useOidc, useOidcUser, OidcUserStatus } from '@axa-fr/react-oidc';
 import { UserIcon } from '@heroicons/react/24/solid';
-import { Button } from './Button';
 import { getLogger } from '../logger';
 
 const logger = getLogger('<User>');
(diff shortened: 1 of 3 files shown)


src/components/User.tsx
```diff
@@ -13,34 +12,35 @@ const User = () => {
...
+        <button
+          className="App-navbar__item"
           onClick={() => { login('/').catch(logger.error); }}
+          type="button"
         >
           <UserIcon />
           Log in
-        </Button>
+        </button>
       );
```
jasonaowen, line 30: I'm afraid I don't quite follow why this one should be a lower-class `<button>` and not an upper-class `<Button>`?

jasonaowen (approved):
Looks good to me! I was able to run it locally and see the signed out, loading, and signed in views. Thanks, @reefdog!

src/components/User.tsx
reefdog: `button` is just a vanilla HTML element. `Button` is a custom component that renders a `button`, but adds the necessary classes for styling according to our design system, and adds properties for common variant states.

I avoided using a `Button` here because I didn't want any of its styling (and wasn't using any of its variant states). But arguably, for consistency, to support this case we should just add a boolean property like `<Button unstyled>` that forewent other styling. 🤔

I quite like that idea but if it's okay I'll hold that as an enhancement for later!

jasonaowen: That makes sense, thank you for the explanation! And yeah, definitely hold that as an enhancement for later.

PR #68, opened 2023-04-07
Change proposal detail URL to `/proposals/{id}`

### LINKED ISSUE #63: Use `/proposals/{id}` instead of `/proposal/{id}` for proposal detail URL
Could we change our individual proposal routes to `/proposals/{id}`?

I know that's not as semantically defensible, but it seems to be more idiomatic (see: GitHub's `/issues/{id}`, StackOverflow's `/questions/{id}`) and ergonomic (go between the index and the individual just by the presence or absence of an ID).

Here's a [good SO discussion](https://stackoverflow.com/questions/6845772/should-i-use-singular-or-plural-name-convention-for-rest-resources) about it. The accepted answer promotes our current syntax, but by far the one with the most votes promotes the Railsy, all-plural form. The electoral college strikes again!

I tested it and it works (including still supporting our query params on the index page). The diff is simple:

```diff
diff --git a/src/App.tsx b/src/App.tsx
index 1e41341..53944e2 100644
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,7 +13,7 @@ const App = () => (
     <main className="App-main">
       <Routes>
         <Route path="/" element={<Placeholder />} />
-        <Route path="/proposal/:proposalId" element={<ProposalDetail />} />
+        <Route path="/proposals/:proposalId" element={<ProposalDetail />} />
         <Route path="/proposals" element={<ProposalListPage />} />
         <Route path="*" element={<NotFound />} />
       </Routes>
diff --git a/src/components/ProposalDetailPanel.tsx b/src/components/ProposalDetailPanel.tsx
index 91a9111..cf2c957 100644
--- a/src/components/ProposalDetailPanel.tsx
+++ b/src/components/ProposalDetailPanel.tsx
@@ -58,14 +58,14 @@ const ProposalDetailPanel = ({
         {process.env.NODE_ENV !== 'production' && (
           <>
             <Link
-              to={`/proposal/${proposalId - 1}`}
+              to={`/proposals/${proposalId - 1}`}
               className="button button--color-gray"
             >
               <BackwardIcon className="icon" />
               Previous
             </Link>
             <Link
-              to={`/proposal/${proposalId + 1}`}
+              to={`/proposals/${proposalId + 1}`}
               className="button button--color-gray"
             >
               Next
diff --git a/src/components/ProposalListTable.tsx b/src/components/ProposalListTable.tsx
index f0b661e..2f991f6 100644
--- a/src/components/ProposalListTable.tsx
+++ b/src/components/ProposalListTable.tsx
@@ -24,7 +24,7 @@ const ProposalListTableRow = ({
   const navigate = useNavigate();

   const handleRowClick = () => {
-    navigate(`/proposal/${proposal.id}`);
+    navigate(`/proposals/${proposal.id}`);
   };

   return (
```

Any objection? I'd prefer to do it today before people start sharing links to proposals tomorrow.

reefdog:
This PR implements the change from `/proposal/{id}` to `/proposals/{id}` for the proposal detail page. More justification in the issue and commit message.

To test, just make sure you can still navigate to the proposal detail page, and that the proposal list page still works (including with magic `count` changing).

Closes #63

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,7 +13,7 @@ const App = () => (
     <main className="App-main">
       <Routes>
         <Route path="/" element={<Placeholder />} />
-        <Route path="/proposal/:proposalId" element={<ProposalDetail />} />
+        <Route path="/proposals/:proposalId" element={<ProposalDetail />} />
         <Route path="/proposals" element={<ProposalList />} />
         <Route path="*" element={<NotFound />} />
       </Routes>
--- a/src/components/ProposalDetailPanel.tsx
+++ b/src/components/ProposalDetailPanel.tsx
@@ -58,14 +58,14 @@ const ProposalDetailPanel = ({
         {process.env.NODE_ENV !== 'production' && (
           <>
             <Link
-              to={`/proposal/${proposalId - 1}`}
+              to={`/proposals/${proposalId - 1}`}
               className="button button--color-gray"
             >
               <BackwardIcon className="icon" />
               Previous
             </Link>
             <Link
-              to={`/proposal/${proposalId + 1}`}
+              to={`/proposals/${proposalId + 1}`}
               className="button button--color-gray"
             >
               Next
--- a/src/components/ProposalListTable.tsx
+++ b/src/components/ProposalListTable.tsx
@@ -24,7 +24,7 @@ const ProposalListTableRow = ({
   const navigate = useNavigate();
 
   const handleRowClick = () => {
-    navigate(`/proposal/${proposal.id}`);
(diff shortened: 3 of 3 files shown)


reefdog:
I'm going to break protocol and accept this on my own behalf, as it's a small change, soundly tested locally, and important to have finished before the demo.

jasonaowen (commented):
I don't think this was a breach of protocol! An edge case, perhaps - we discussed this in #63, where you shared exactly this diff as a comment and I +1'd it.

I appreciate your thoughtfulness, but please don't feel at all bad about this - and thanks for getting it done so quick!

PR #69, opened 2023-04-07
Add landing page

### LINKED ISSUE #48: Create landing page
I think we will want a landing page for logged-out users to see. Currently, we're still showing the React logo and a few PDC links. Soon we will have the option of showing the proposals list (#14) in the root, which would require logging in immediately. I think it's worthwhile to have a brief landing page that would introduce the project and maintain the links we currently have.

reefdog:
This PR adds a landing page with lightwight welcome and guidance for use.

**Logged out:**

logged-out

**Logged in:**

logged-in

Some prerequisite changes that I also made:
- Buttons are now `inline-flex`, so they don't have to occupy 100% of available width.
- It's possible to have panels that don't occupy the entire viewport. (This was always the intent, I just didn't complete the work for it because I didn't have a situation that called for it.)
- Added `EmailLink` and `OffsiteLink` components for responsibly handling those respective link types with minimal boilerplate. See story and documentation for use.
- Added styling rules for `<dl>` lists, which are great for key/value pairs like dictionaries or FAQs. This also introduces a tiny new component, `Dli`, that wraps the `<dt>`/`<dd>` children of `<dl>`s.
- Added a `.subtle` text decorator for small, gray text.

Closes #48


src/index.css
```diff
@@ -86,10 +87,52 @@ h6 {
   line-height: 1.2;
 }
 
+/*
+  Our implementation of DLs assumes you will wrap each DT/DD pair in a
+  wrapper DIV with the class name `.dli`, e.g.:
+  <dl>
+    <div class="dli">
```
slifty, line 94: why doesn't the html spec just have li as a valid child of dl?

Please open a proposed change to the HTML spec thanks.

slifty: Oh actually, and please boo me off the stage: would it be worth make a dli component so you can just do `<dl><dli>...` instead of the div class blah blah blah?

You could have that component expect / require content for the dd and the dt.  You'd be able to throw out this comment I think, and you'd also be able to avoid repetition of the class stuff in your dl implementations.

src/pages/Landing.tsx
```diff
@@ -0,0 +1,102 @@
...
+            <UserIcon className="icon" />
+            Log in
+          </Button>
+        )}
+        <dl>
+          {!isAuthenticated && (
+            <div className="dli">
+              <dt>Need an account?</dt>
+              <dd>
+                Email Jim McGowan:
```
slifty, line 56: I don't know if we're in "move fast and break things" mode for this but this strikes me as a value that should be configurable / pulled from some kind of configuration file or environment variable.

src/pages/Landing.tsx
```diff
@@ -0,0 +1,102 @@
...
+                {' '}
+                <a href="mailto:jimmcgowan@opentechstrategies.com?subject=PDC%20Pilot%20account">
```
slifty, line 58: Same suggestion here regarding configuration / not hardcoding values like this.

slifty: (Alternatively, as a middle ground solution it might feel less sore if we made it more generic than Jim's name and email address specifically, e.g. something more like "admin@...")

src/pages/Landing.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  return (
+    <Panel
+      fill={false}
+      className="landing-panel"
+    >
+      <PanelBody>
+        <h3 style={{ fontWeight: '600' }}>Welcome to the Philanthropy Data Commons pilot.</h3>
+        <p>
+          This data viewer displays the data in the Philanythropy Data Commons pilot.
```
slifty, line 23: I picked this line to ask this question: do we care about internationalization at this stage of the project?  I assume the answer is no; sorry if I'm rehashing conversations that have already been had!

slifty (commented):
I'm going to actually set this up locally but wanted to leave these comments now for your consideration in the mean time.

src/pages/Landing.tsx
reefdog: That's a great question, and yes, we do care!

Unfortunately while we're accruing debt here, that ship has sailed (in that the site is going to be in user's hands in a matter of minutes without i18n) so that shouldn't block this particular PR. But it's important and I'm glad you noted it.

src/index.css
reefdog: Oh, I like that idea! I wonder if I can lowercase it though so it doesn't look weird between `<dl>` and `<dd>`…

> You could have that component expect / require content for the dd and the dt.

I'd very much like for it to just use the `children` it was handed, e.g.:

```jsx
<dl>
  <dli>
    <dt />
    <dd />
  </dli>
</dl>
```

Do you know if React will let me dictate which elements are allowed as `children`?

slifty (approved):
Ran this locally -- I don't have an account yet so can't confirm the actual functionality of post-login, but will defer to you on whether you'd like to block on that!

I left comments above but I don't think any of them are blocking comments unless you agree with them and would like to incorporate them.

src/index.css
reefdog: Implemented. This was a great suggestion, thanks @slifty!

> Do you know if React will let me dictate which elements are allowed as children? 

Looks like no, so this just accepts any children. Ah well, the some holds true for any HTML element: if you pass it something invalid, the browser will scream at you.

src/pages/Landing.tsx
reefdog: Since this is for a pilot group of like 20 people atm, all of whom know Jim, the clarity of this direction is actually preferred. As a point of process, we may want a more general inbox, but that's a decision out of scope of this implementation so I'm gonna move forward on this point!

I mean, I get that this could be `Email ${ACCOUNT_ADMIN_NAME}` and fair enough! But I'll reserve that for the first time we need to change it and/or have downtime before getting the pilot participant logins out this afternoon.

reefdog: @slifty My response here was initially written in response to [this comment](https://github.com/PhilanthropyDataCommons/data-viewer/pull/69#discussion_r1160888777) and then copied here with slight changes, so some of it actually addresses your point there.

reefdog:
@slifty That was great feedback, thanks! You're spot on that we've got values hardcoded that shouldn't be, but I'm happy punting on that for now until we better understand how often we'll be doing that, how we'd want to store those values, how we'd want to pass them into our i18n, etc.

On which point, good call on i18n. We intended to have that from launch, and currently do not, and should! I'll open the issue.

I added a `Dli` component (can't go with `dli` as much as I'd like to) as syntactic sugar for working with `<dl>`s, and slipped in some new components for dealing with email and offsite links more responsibly. Re-review appreciated!

slifty (approved):
Looks great!  Thanks!

src/index.css
```diff
@@ -1,6 +1,7 @@
 :root {
   --color--gray--darker: #343434;
   --color--gray--dark: #444;
+  --color--gray--medium-dark: #999;
```
jasonaowen, line 4: I found this a little hard to read, and looked up what the [accessibility standards for color contrast](https://developer.mozilla.org/en-US/docs/Web/Accessibility/Understanding_WCAG/Perceivable/Color_contrast) are. In short, the minimum is 4.5:1, and ideal is 7:1; [#999 against white](https://webaim.org/resources/contrastchecker/?fcolor=999999&bcolor=FFFFFF) has a ratio of 2.84:1.

`#757575` has a ratio of 4.6:1, which clears the minimum; is that subtle enough?

```suggestion
  --color--gray--medium-dark: #757575;
```

src/index.css
```diff
@@ -87,6 +87,43 @@ h6 {
   line-height: 1.2;
 }
 
+/*
+  Our implementation of DLs assumes you will wrap each DT/DD pair in a
+  custom `<Dli>` component, which generates the following markup:
```
jasonaowen, line 92: Can you please say more about this? Why is adding a `div` into the tree better than adding a class on `<dl>`? I don't understand the problem this solves!

reefdog:
Huzzah! 🚀

src/components/OffsiteLink.tsx
```diff
@@ -0,0 +1,40 @@
...
+   * Visible contents of the link.
+   * If omitted, will default to the destination URL (`to`).
+   */
+  children?: React.ReactNode;
+  className?: string;
+}
+
+/**
+ * Component for linking off-site.
+ * Defaults to opening links in a new window/tab (which can be disabled) and
```
jasonaowen, line 23: I'm going to push back a bit on this, too!

Why do we want links to open in a new tab? Users have the ability to choose whether to open in a new window/tab in every modern browser. If they click an external link and decide they wanted it to be in a new tab after all, they can hit the back button and middle click. If we force it to open in a new tab, users can't make that decision.

jasonaowen (commented):
Sorry for the late review!

src/pages/Landing.tsx
bickelj: `s/Philanythropy/Philanthropy/g`

reefdog: Oh nooooo

src/index.css
reefdog: Great note, will change!

src/index.css
reefdog: Somewhat infamously, definition lists don't have individual children. Instead, their children are a soup of terms and definitions, e.g.:

```html
<dl>
  <dt>Dog</dt>
  <dd>Furry quadruped with a sweet nature.</dd>
  <dt>Cat</dt>
  <dd>Furry quadruped with a sneaky nature.</dd>
</dl>
```

It's very common to want to style the `dt`/`dd` pairs as a unit, e.g. to put spaces between them so they appear somewhat like:

```
Dog
Furry quadruped with a sweet nature.

Cat
Furry quadruped with a sneaky nature.
```

Of course it's possible to style them by, say, adding margin below `dt` or above `dd`, and using `:first-child` and `:last-child` to get rid of boundary margins.

But the whole thing is very unsemantic, and doesn't let you do neat things like "use flexbox column layout and put a gap between each pair". So it's now valid to wrap them in a `div` — acting equivalent of an `li` — while still getting the semantics of `dl`/`dt`/`dd`.

That's certainly a lot of justification, but do you consider it _good_ justification?

src/components/OffsiteLink.tsx
reefdog: This is where I appreciate your rethinking my priors.

I philosophically agree with you but have lost this battle so often that I not reflexively default to what I expect the client will want ("offsite links open in new tabs"). I am _ecstatic_ to invert this behavior and default to web-native behavior. Excellent pushback.

src/index.css
jasonaowen: That's great justification, and more importantly, great explanation! Thank you, @reefdog, TIL.

PR #79, opened 2023-04-10
Go to proposals page from log in button

### LINKED ISSUE #75: Go to Dashboard after initial login
When an unauthenticated user tries to access an authenticated page, they're routed to Keycloak and then back to the original requested page upon successful login. 👍

The one exception to this, however, should be the root. When users visit the home page and log in, we should send them to the Dashboard (`/proposals` currently).

Can this be special-cased?

jasonaowen:
Change the target of the two login buttons - in the `<User>` component of the navbar, and in the landing page - to go to `/proposals` instead of back to the landing page.

This does not affect the existing behavior of navigating to a secured page while signed out, which will continue to go through the login process and then redirect to the destination secured page.

Resolves #75 Go to Dashboard after initial login

### DIFF
--- a/src/components/User.tsx
+++ b/src/components/User.tsx
@@ -21,7 +21,7 @@ const User = () => {
       return (
         <button
           className="App-navbar__item"
-          onClick={() => { login('/').catch(logger.error); }}
+          onClick={() => { login('/proposals').catch(logger.error); }}
           type="button"
         >
           <UserIcon />
--- a/src/pages/Landing.tsx
+++ b/src/pages/Landing.tsx
@@ -41,7 +41,7 @@ const Landing = () => {
           <Button
             color="blue"
             inverted
-            onClick={() => { login('/').catch(logger.error); }}
+            onClick={() => { login('/proposals').catch(logger.error); }}
           >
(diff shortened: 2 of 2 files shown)


reefdog (approved):
It works! Thanks for the quick solution.

I'll admit this is a _bit_ weird to me, as I would have expected us to have some sort of… callback recipient page that handled the routing and let us centrally specify "if a path is included, go there; if none/root, go to this default". So it feels weirdly unidiomatic to have to specify the default post-login destination on each login button.

But! this is essentially a no-code solution and I can see it being useful to have per-button destinations. So, good enough (and definitely the right level of effort for the value of the request).

Thanks!

jasonaowen:
> I would have expected us to have some sort of… callback recipient page that handled the routing and let us centrally specify "if a path is included, go there; if none/root, go to this default".

The library takes care of all that routing for us! Which is really great, because I've written that kind of thing myself and it is non-trivial. 

> So it feels weirdly unidiomatic to have to specify the default post-login destination on each login button.

We already were! That's what the `'/'` parameter to `login()` was.

PR #80, opened 2023-04-20
Handle API change to use bundles

jasonaowen:
In PR [service#296](https://github.com/PhilanthropyDataCommons/service/pull/296), the return format of the `/proposals` API endpoint is changing from a top-level array of proposals to an object containing proposals under the `entries` key. It also returns the `total` number of proposals, which will allow us to do pagination, but that's separate work; for now, just update the parsing code to use the new format.

### DIFF
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -75,6 +75,7 @@ const useProposals = (page: string, count: string) => {
     fetch(new URL(path, API_URL))
       .then(throwNotOk)
       .then((res) => res.json())
+      .then(({ entries }: { entries: Proposal[]; }) => entries)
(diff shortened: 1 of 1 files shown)


jasonaowen:
Note that there is definitely room for improvement here - for example, with some effort, this could be made backwards compatible with the API changes - but this is intended to be short-lived code to be replaced as part of adding pagination.

slifty (approved):

jasonaowen:
We need to wait to merge (and therefore automatically deploy) this until PhilanthropyDataCommons/service#296 is deployed. @bickelj, can I ask you to merge when ready?

PR #82, opened 2023-04-27
Improve sticky table headers

reefdog:
Our table heads have top-borders. When the header is sticky, that top border used to remain visible when the header was sticky. That’s not ideal; a border is no longer needed to distinguish it from other content.

Before; watch the top of the panel as the header becomes sticky, how it looks like the shadow gets darker:

https://user-images.githubusercontent.com/4731/235010060-0b1a1060-49de-4a8b-a757-6c28d01a537a.mov

This PR makes the border defined via a CSS variable rather than statically, and sets the `top` property of the sticky header to the negative value of that variable so that the header scrolls that much out of view.

After; watch the top of the panel for how the shadow doesn't appear to get darker:

https://user-images.githubusercontent.com/4731/235010130-c40d2842-189e-4dff-8f85-c4d8dba27fd5.mov

### DIFF
--- a/src/components/Table/Table.css
+++ b/src/components/Table/Table.css
@@ -1,3 +1,7 @@
+:root {
+  --table--border-width: 1px;
+}
+
 .table {
   width: 100%;
 }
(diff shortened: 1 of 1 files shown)


jasonaowen (approved):
Nice! Thanks, @reefdog!

Issue #83, opened 2023-04-27
Enable search


jasonaowen:
We want to allow searching proposals. Previously, we introduced a non-functional search box in #49, intended to promise future functionality. This issue is for implementing that functionality, and is the front-end counterpart to https://github.com/PhilanthropyDataCommons/service/issues/277 .

jasonaowen:
In order to meet some demo deadlines, and while the API implementation is being figured out, as a first draft we will implement a front-end, filter-based "search". This takes advantage of the fact that we're not yet doing pagination (see #15), and instead just loading [the first 1,000 proposals](https://github.com/PhilanthropyDataCommons/data-viewer/blob/c908f0208c5dfb2aa190be9d66ce0471236fb52b/src/pages/ProposalList.tsx#L45). To restate that in another way, we cannot have both front-end filter search and pagination at the same time, and we are choosing to prioritize search. Once the API supports searching, we will switch to that, and be able to implement pagination.

jasonaowen:
We got client-side search in #84, but I'm leaving this open until we get server-side search finished.

reefdog:
@jasonaowen I think this can be closed now, right?

jasonaowen:
Oops! Yes, this should have been resolved by #111.

PR #84, opened 2023-04-28
Search filter

jasonaowen:
In order to meet some demo deadlines, and while the API implementation is being figured out, implement "search" on the front-end as simple string matching. As we intend the back-end to do, this shows any proposal containing a field value that matches the search query; unlike what we intend the back-end to do, it does no stemming, is not aware of word boundaries, includes stopwords, and so on - it's strictly string matching, with some lowercasing thrown in for case insensitivity.

This is intended to be reverted once the API supports searching, both because of the additional functionality it will offer and because this approach precludes pagination.

See the individual commits for more details.

Issue #83 Enable search


jasonaowen:
I've updated Keycloak to allow authentication from this deploy preview. Example: https://deploy-preview-84--philanthropy-data-commons-viewer.netlify.app/proposals?q=test

reefdog (approved):
Code-wise, this is great!

I have the following non-blocking UX requests:
- Set the `disabled` attribute on the submit button when there is no text in the field (and make sure other methods of form submission are blocked, e.g. enter-submit)
- Add a "Clear search" capability so users can reset the page short of clicking "Dashboard" or submitting an empty string (especially since the latter would be blocked by my first UX request)

I was going to request a search results count, but I realize the paginator element will be the correct place to report that, so disregard.

I can implement the above if you'd like; we can also punt them to enhancement issues if you prefer. Approving this so you can decide which.

reefdog:
@jasonaowen Any objections to me adding the "Clear search" functionality that I'd like to see? It wouldn't be throwaway because we'd want it for "real search" too.

PR #90, opened 2023-04-28
Fix proposal name field on table page

### LINKED ISSUE #89: Proposal name not showing on proposal list page
Our proposal list page has a [hard-coded list of columns](https://github.com/PhilanthropyDataCommons/data-viewer/blob/16f383e257321bffe03154da4673da78eb9e144c/src/components/ProposalListTablePanel.tsx#L20-L40). The proposal name is [currently defined as `proposal_title`](https://github.com/PhilanthropyDataCommons/data-viewer/blob/16f383e257321bffe03154da4673da78eb9e144c/src/components/ProposalListTablePanel.tsx#L32), but should be `proposal_name`.

reefdog:
The field that contains the proposal name is `proposal_name`, but the code is currently using `proposal_title` (which doesn't exist).

This PR updates the default column list on the proposal table page so the proposal name appears.

Testing:
1. Before pulling this code, look at the `/proposals` page and notice that there is an empty column between the "Organization Operating Budget" and "Proposal Summary" columns.
2. Pull the code and reload; now there should be a "Proposal Name" column (with data for some proposals).

Closes #89

### DIFF
--- a/src/components/ProposalListTablePanel.tsx
+++ b/src/components/ProposalListTablePanel.tsx
@@ -29,7 +29,7 @@ const DEFAULT_COLUMNS = [
   'organization_mission_statement',
   'organization_start_date',
   'organization_operating_budget',
-  'proposal_title',
+  'proposal_name',
   'proposal_summary',
   'proposal_amount_requested',
   'proposal_budget',


jasonaowen (approved):
Thank you for jumping on this, @reefdog!

I was able to run this branch locally and see the correct column and no empty columns.

PR #93, opened 2023-05-01
Use deep proposals data from API

jasonaowen:
Recently, [the API gained the ability to include all the necessary data in the `/proposals` endpoint](https://github.com/PhilanthropyDataCommons/service/pull/313), instead of only including IDs. As a result, we can stop making an additional N queries to `/proposal/{id}` to fill out the list, significantly improving load times.

We're still not using the [also-recently-introduced `total` field](https://github.com/PhilanthropyDataCommons/service/pull/296), which was introduced to support pagination, because instead we're doing [client-side search](https://github.com/PhilanthropyDataCommons/data-viewer/pull/84). Once [server-side search is ready](https://github.com/PhilanthropyDataCommons/service/pull/320), we'll be able to use that and implement pagination.

### DIFF
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -67,8 +67,7 @@ const useProposal = (proposalId: string) => (
 
 const useProposals = (page: string, count: string) => {
   const { fetch } = useOidcFetch();
-  const [proposals, setProposals] = useState<number[]>([]);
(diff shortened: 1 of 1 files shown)


jasonaowen:
D'aw, thank you, @reefdog!

Issue #62, opened 2023-04-07
Add sidebar to Proposal Detail page


reefdog:
As demonstrated in #19, we need a sidebar of proposals from the table:

<img width="1512" alt="PDC Data Viewer 02 - Proposal" src="https://user-images.githubusercontent.com/4731/220751450-d328b240-42f7-41fe-a1dd-887f3318edd9.png">

(Note that this is an old screenshot and is merely demonstrative of the design and features at the time.)

This sidebar is intended to have all the same features as the main table, and include the same context (page, sort order, filter settings) as the main table had at the time the proposal was opened. (Defaulting to the starting information if the page was loaded directly.)

Proposed by @jim-mcgowan [here](https://github.com/PhilanthropyDataCommons/data-viewer/issues/13#issuecomment-1453760386) are the following fields for the sidebar cards:

> - Organization Name
> - Organization City, Organization State/Province, Organization Country
> - Proposal Name

PR #94, opened 2023-05-02
Add grid-based proposal list for sidebar

reefdog:
This PR adds the components necessary for displaying a list of proposals in a grid-based format suitable for narrow contexts like a sidebar.

[Screenshot redacted because it contains identifiers of PDC pilot participant organizations.]

To add it to a page, you simply:

```diff
- <PanelGrid>
+ <PanelGrid sidebarred>
+   <PanelGridItem>
+     <ProposalListGridPanel
+       proposals={…}
+   />
+   </PanelGridItem>
    <PanelGridItem>
      {Other panel}
    </PanelGridItem>
  </PanelGrid>
```

Issue #62

### DIFF
--- a/src/components/ProposalListGridPanel.tsx
+++ b/src/components/ProposalListGridPanel.tsx
@@ -1,0 +1,21 @@
+import React from 'react';
+import {
+  Panel,
+  PanelBody,
+} from './Panel';
+import { ProposalListGrid } from './ProposalListGrid';
+
+interface ProposalListGridPanelProps {
+  proposals: {
+    id: string;
+    values: Record<string, string[]>;
+  }[];
+}
+
+export const ProposalListGridPanel = ({ proposals }: ProposalListGridPanelProps) => (
+  <Panel>
+    <PanelBody>
+      <ProposalListGrid proposals={proposals} />
+    </PanelBody>
+  </Panel>
+);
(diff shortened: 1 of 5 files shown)


reefdog:
@jasonaowen This version of the panel doesn't include any header actions; that's because I assume that we shouldn't attempt to tackle searching in this version, and that should wait for the true "pagination and search via API" version.

src/components/ProposalListGridPanel.tsx
```diff
@@ -0,0 +1,21 @@
...
+import { ProposalListGrid } from './ProposalListGrid';
+
+interface ProposalListGridPanelProps {
+  proposals: {
+    id: string;
+    values: Record<string, string[]>;
+  }[];
+}
+
+export const ProposalListGridPanel = ({ proposals }: ProposalListGridPanelProps) => (
```
jasonaowen, line 15: > To add it to a page, ...

Would you like me to wire it up as part of this PR, or shall we merge this and I'll open a follow-up?

jasonaowen (approved):
Oh sorry, I just realized I hadn't submitted my pending review! I looked through this the other day, including looking at the storybook, and it looks great - excellent work, @reefdog!

src/components/ProposalListGridPanel.tsx
reefdog: Since the page we want it on isn't populating the `proposals` yet, I figured we'd wait until then to actually add it.

PR #95, opened 2023-05-02
Fix proposal test data

reefdog:
In #90, we updated the attribute that contains the proposal name. We forgot to update the mock data in the Storybook stories, however.

This PR repairs that. Consequently, the `ProposalListTable` and `ProposalListTablePanel` stories should work properly again.

Belated fix as part of #89


src/stories/ProposalListTable.stories.tsx
```diff
@@ -35,7 +35,7 @@ export const Default: Story = {
       organization_mission_statement: 'Organization Mission Statement',
       organization_start_date: 'Organization Start Date',
       organization_operating_budget: 'Organization Operating Budget',
-      proposal_title: 'Proposal Title',
+      proposal_name: 'Proposal Title',
```
jasonaowen, line 38: I think we want to fix the field name too, right? I checked api.pdco and it does have `Proposal Name` as the label.

```suggestion
      proposal_name: 'Proposal Name',
```

jasonaowen (changes requested):
Good catch!

src/stories/ProposalListTable.stories.tsx
reefdog: 🤦 Good catch, yes!

reefdog: @jasonaowen Fixed (via a force-push rather than suggestion-acceptance since this was in two files). Requires re-review when you have a moment!

jasonaowen (approved):
Awesome, thank you @reefdog!

PR #96, opened 2023-05-02
Conjoin search input and button

reefdog:
This PR adds a `FormElementGroup` component that we can use to conjoin form elements (buttons, inputs) into horizontal or vertical groups. It also applies that component to the search atop the proposal list table, so the search input and its submit button are visually bound.

**Before:**
1-before

**After:**
2-after

**Testing:**
- `npm start`
- Open the app in browser and confirm the search form looks like "After" above
- `npm run storybook`
- Open the Storybook in browser and look at the `FormElementGroup` docs and confirm they are 👍

### DIFF
--- a/src/components/FormElementGroup.css
+++ b/src/components/FormElementGroup.css
@@ -1,0 +1,52 @@
+.form-element-group {
+  display: flex;
+  gap: 0;
+}
+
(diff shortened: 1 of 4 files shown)


jasonaowen (commented):
Very fancy!

PR #98, opened 2023-05-04
Add Paginator component

reefdog:
This PR adds the `Paginator` component and its story, as well as the `PaginatorWrapper` that we use when actually placing the paginator on a real body:

Paginator

There are also two utility methods that could easily be moved to a utility file, but we don’t yet have a pattern for that.

**Testing:**
1. `npm run storybook` and look at the `Paginator` story
2. `npm start`
3. To see it in context (but without true functionality), modify `ProposalListTablePanel.tsx` so the `<PanelBody>` looks like this:
    ```jsx
    <PanelBody>
      <PaginatorWrapper>
        <Paginator
          itemsPerPage={100}
          totalItems={950}
          currentPage={1}
          setCurrentPage={(page) => { console.log(page); }}
        />
      </PaginatorWrapper>
      <ProposalListTable
        fieldNames={fieldNames}
        proposals={proposals}
        columns={DEFAULT_COLUMNS}
        wrap={wrap}
      />
    </PanelBody>
    ```

Contributes to #15

_Extracted from #46, which can now be closed._

### DIFF
--- a/src/components/Paginator/Paginator.css
+++ b/src/components/Paginator/Paginator.css
@@ -1,0 +1,83 @@
+:root {
+  --paginator--font-size: 0.875rem;
+  --paginator--min-height: calc(var(--paginator--font-size) * 4);
+  --paginator--border-radius: 9999px;
+
+  --paginator--button-size: var(--paginator--min-height);
+  --paginator--button-icon-size: calc(var(--paginator--button-size) / 2.5);
+}
+
+.paginator-wrapper {
+  /* Cover the entire parent element and don't scroll with it. */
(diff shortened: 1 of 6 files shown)


reefdog:
@jasonaowen Ah, I just remembered your suggestion to turn the page arrow buttons into links. Converting this to a draft until I make that change!

reefdog:
It's pointless to keep this PR open as a draft when we don't have pagination scheduled as work, and when the implementation of this component needs to change in response to how the API actually works. Closing it, but will rename and keep the branch around for future reference.

PR #99, opened 2023-05-04
Share the DataViewerProposal interface

reefdog:
This PR creates a pattern for sharing TypeScript interfaces between files (`src/interfaces/Interface.ts`) and creates our first shared interface: `DataViewerProposal`. This is the interface that the DataViewer expects for a given proposal.

On its own, this PR changes nothing besides establishing the pattern with a single example. Future PRs will actually make more use of this shared interface.

**Testing:** Nothing should have broken or changed!

### DIFF
--- a/src/components/ProposalListGrid.tsx
+++ b/src/components/ProposalListGrid.tsx
@@ -1,11 +1,9 @@
 import React from 'react';
 import { ProposalListGridItem } from './ProposalListGridItem';
+import { DataViewerProposal } from '../interfaces/DataViewerProposal';
 
 interface ProposalListGridProps {
-  proposals: {
-    id: string;
-    values: Record<string, string[]>;
-  }[];
+  proposals: DataViewerProposal[];
 }
 
 export const ProposalListGrid = ({ proposals }: ProposalListGridProps) => (
--- a/src/components/ProposalListGridItem.tsx
+++ b/src/components/ProposalListGridItem.tsx
@@ -1,12 +1,10 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
+import { DataViewerProposal } from '../interfaces/DataViewerProposal';
 import './ProposalListGridItem.css';
 
 interface ProposalListGridItemProps {
-  proposal: {
-    id: string;
-    values: Record<string, string[]>;
-  };
+  proposal: DataViewerProposal;
 }
 
 export const ProposalListGridItem = ({ proposal }: ProposalListGridItemProps) => {
--- a/src/components/ProposalListGridPanel.tsx
+++ b/src/components/ProposalListGridPanel.tsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { DataViewerProposal } from '../interfaces/DataViewerProposal';
 import {
   Panel,
   PanelBody,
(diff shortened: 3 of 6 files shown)


reefdog:
This `DataViewerProposal` interface reflects the structure we create [here](https://github.com/PhilanthropyDataCommons/data-viewer/blob/1662b9e1657c930df1e65a9607a980a9a84a03cd/src/pages/ProposalList.tsx#L25-L40). It strikes me we should probably use it there, then?

jasonaowen (approved):
It definitely makes sense to promote this to a named concept! In retrospect, we should probably have done so a while ago. Thanks, @reefdog!

I appreciate the desire to distinguish between an internal thing vs an API thing. I'm not sure I like `DataViewer` as the prefix - particularly with rumblings about data editing on the horizon. Maybe instead we should prefix API things with `Api`, like `ApiProposal` and `ApiCanonicalField` (or soon `ApiBaseField`)?

However, renaming this kind of thing is not that big a deal, and it's better to have the concept named than not!

I haven't tested this yet, but CI has run the linter and the build, so I'm not too worried.

reefdog:
@jasonaowen Thank you much, both for the review and that great feedback. I like your suggestion of essentially renaming the other `Proposal` to `ApiProposal` and then this one can become `Proposal`; I also like your implied suggestion of doing that later so this doesn't balloon.

Re: your thumbs up on my [comment](https://github.com/PhilanthropyDataCommons/data-viewer/pull/99#issuecomment-1535510091), do you think I should make that change as part of this PR?

jasonaowen:
> @jasonaowen Thank you much, both for the review and that great feedback. I like your suggestion of essentially renaming the other `Proposal` to `ApiProposal` and then this one can become `Proposal`; I also like your implied suggestion of doing that later so this doesn't balloon.

It would probably be a smaller change to do that rename now rather than later, if you've got the bandwidth; doing it later mainly helps you keep your momentum - which yes to make explicit is a worthwhile tradeoff!

> > This `DataViewerProposal` interface reflects the structure we create [here](https://github.com/PhilanthropyDataCommons/data-viewer/blob/1662b9e1657c930df1e65a9607a980a9a84a03cd/src/pages/ProposalList.tsx#L25-L40). It strikes me we should probably use it there, then?
> 
> Re: your thumbs up on my [comment](https://github.com/PhilanthropyDataCommons/data-viewer/pull/99#issuecomment-1535510091), do you think I should make that change as part of this PR?

Looking more closely, I believe it's already done! By setting the prop type to this new interface, TypeScript should be validating that the values passed into it match. Try renaming one of the keys in that linked code; hopefully you'll get a compilation error.

reefdog:
> It would probably be a smaller change to do that rename now rather than later, if you've got the bandwidth; doing it later mainly helps you keep your momentum - which yes to make explicit is a worthwhile tradeoff!

Ah, okay! In that case (1) yes I have the bandwidth and (2) I will go ahead and make that change. Since the other two PRs ladder off this one, I'll need to rebase them, but that's no biggie.

I appreciate the suggestion! Will you want to re-review after I do that?

> Try renaming one of the keys in that linked code; hopefully you'll get a compilation error.

Ayyy, you're right! 🎉

jasonaowen:
Awesome, thanks @reefdog! No, go ahead and merge once you're done.

reefdog:
@jasonaowen I just realized the renaming will conflict with your work on #104, too, which tips the punt scales. I'll make an issue and we'll tackle just as soon as the boat stops rocking for a moment.

PR #101, opened 2023-05-04
Get proposal values safely

### LINKED ISSUE #100: Sidebar list of proposals doesn't display org name correctly

sidebar

reefdog:
_This PR ladders off #99._

This PR adds two utility methods, as well as tests of those methods.

- `getProposalValue()` lets you safely fetch the value of a particular attribute, returning `undefined` if the value isn't valid (missing, empty array, or array of empty strings), or the value if it is.
- `getProposalValueFromCandidates()` accepts an array of keys and returns the first good one, or undefined if none are.

While this PR doesn't demonstrate their use, a future PR will use these to fix #100.

Note: Neither of these methods handles the concern of how we join arrays of values for display in the interface. That would be a different method, or possibly even a React component. Either way, we're punting on it for now because we don't have any data like that.

**Testing:**
- `npm run test` yay!

### DIFF
--- a/src/utils.test.ts
+++ b/src/utils.test.ts
@@ -1,5 +1,6 @@
 import {
   getProposalValues,
+  getProposalValuesFromCandidates,
 } from './utils';
 
 test('can get proposal value', () => {
@@ -27,4 +28,38 @@ test('returns undefined for missing values', () => {
   expect(getProposalValues(proposal, 'empty_array')).toBeUndefined();
   expect(getProposalValues(proposal, 'string_only')).toBeUndefined();
   expect(getProposalValues(proposal, 'whitespace')).toBeUndefined();
+});
+
+test('can get best proposal value from candidates', () => {
+  const proposal = {
+    id: 'proposal',
+    values: {
+      empty_array: [],
+      string_only: [''],
+      org_legal_name: ['ABC Company, LLC'],
+      org_dba_name: ['XYZ Company'],
+    },
+  };
+
+  // Intentionally includes some attributes that should be skipped before the target attributes.
+  const preferDbaName = [
+    'missing_attribute',
+    'empty_array',
+    'string_only',
+    'org_dba_name',
+    'org_legal_name',
+  ];
+  const preferLegalName = [
+    'missing_attribute',
+    'empty_array',
+    'string_only',
+    'org_legal_name',
+    'org_dba_name',
+  ];
+
+  expect(getProposalValuesFromCandidates(proposal, preferDbaName))
+    .toEqual(proposal.values.org_dba_name);
+
+  expect(getProposalValuesFromCandidates(proposal, preferLegalName))
+    .toEqual(proposal.values.org_legal_name);
 });
--- a/src/utils.test.ts
+++ b/src/utils.test.ts
@@ -1,0 +1,30 @@
+import {
+  getProposalValues,
+} from './utils';
+
+test('can get proposal value', () => {
+  const proposal = {
+    id: 'proposal',
+    values: {
+      org_name: ['ABC Company'],
+    },
+  };
+
+  expect(getProposalValues(proposal, 'org_name')).toEqual(proposal.values.org_name);
+});
+
+test('returns undefined for missing values', () => {
+  const proposal = {
+    id: 'proposal',
+    values: {
+      empty_array: [],
+      string_only: [''],
+      whitespace: [' '],
+    },
+  };
+
+  expect(getProposalValues(proposal, 'missing_attribute')).toBeUndefined();
+  expect(getProposalValues(proposal, 'empty_array')).toBeUndefined();
+  expect(getProposalValues(proposal, 'string_only')).toBeUndefined();
+  expect(getProposalValues(proposal, 'whitespace')).toBeUndefined();
+});
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -1,0 +1,25 @@
+import { DataViewerProposal } from './interfaces/DataViewerProposal';
+
+/**
+   * Safely gets the proposal value array for a given proposal attribute key.
+   *
+   * @param  {DataViewerProposal} proposal
+   * @param  {string} key The key for the proposal attribute we want.
+   * @return {(Array|undefined)} The proposal value array. Falls back to undefined.
+   */
+const getProposalValues = (proposal: DataViewerProposal, key: string) => {
+  // This is a temporary fix:
+  // The API currently allows fields to be set to blank strings,
+  // and dutifully includes those in the value array.
+  // We do not consider those valid when choosing a proposal value,
+  // so here we are making sure we have at least one non-blank/whitespace value.
+  if (!proposal.values[key]?.some((value) => value.trim() !== '')) {
+    return undefined;
+  }
+
+  return proposal.values[key];
+};
+
+export {
+  getProposalValues,
+};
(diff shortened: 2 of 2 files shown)


src/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+   * @param  {string} key The key for the proposal attribute we want.
+   * @return {(Array|undefined)}
+   */
+const getProposalValue = (proposal: DataViewerProposal, key: string) => {
+  // This is a temporary fix:
+  // The API currently allows fields to be set to blank strings,
+  // and dutifully includes those in the value array.
+  // We do not consider those valid when choosing a proposal value,
+  // so here we are making sure we have at least one non-blank value.
+  if (!proposal.values[key]?.some((value) => value !== '')) {
```
jasonaowen, line 16: We might want to [trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) the values, too, in case some are whitespace only:

```suggestion
  if (!proposal.values[key]?.some((value) => value.trim() !== '')) {
```

src/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+   * Safely gets the proposal value for a given proposal attribute key.
+   *
+   * @param  {DataViewerProposal} proposal
+   * @param  {string} key The key for the proposal attribute we want.
+   * @return {(Array|undefined)}
+   */
+const getProposalValue = (proposal: DataViewerProposal, key: string) => {
+  // This is a temporary fix:
+  // The API currently allows fields to be set to blank strings,
+  // and dutifully includes those in the value array.
```
jasonaowen, line 13: This is helpful, but I feel like this is the wrong place to solve this; instead, I think we should be filtering such empty values during mapping the API response into our own data structures.

How comfortable do you feel tackling that? I know this function is a critical component of this work, so I don't want to block the rest of it on doing temporary API cleanup in the "right" place!

src/utils.ts
```diff
@@ -0,0 +1,41 @@
...
+/**
+   * From a list of proposal attribute keys, safely gets the first valid proposal value.
+   *
+   * Key order is relevant, as the first matching key's value will be returned.
+   *
+   * @param  {DataViewerProposal} proposal
+   * @param  {string[]} keys An array of keys, ordered by preference with most-preferred first.
+   * @return {(Array|undefined)} The proposal value array. Falls back to undefined.
+   */
+const getProposalValueFromCandidates = (proposal: DataViewerProposal, keys: string[]) => {
```
jasonaowen, line 32: Since this returns an array (for the case where there multiple answers to a single form field), should this be `getProposalValuesFromCandidates` (emphasis on the plural) or so?

src/utils.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      string_only: [''],
+    },
+  };
+
+  expect(getProposalValue(proposal, 'missing_attribute')).toBeUndefined();
+  expect(getProposalValue(proposal, 'empty_array')).toBeUndefined();
+  expect(getProposalValue(proposal, 'string_only')).toBeUndefined();
+});
+
+test('can get best proposal value from candidates', () => {
```
jasonaowen, line 31: :star_struck: tests!

src/utils.ts
```diff
@@ -0,0 +1,41 @@
+import { DataViewerProposal } from './interfaces/DataViewerProposal';
+
+/**
+   * Safely gets the proposal value for a given proposal attribute key.
+   *
+   * @param  {DataViewerProposal} proposal
+   * @param  {string} key The key for the proposal attribute we want.
+   * @return {(Array|undefined)}
+   */
+const getProposalValue = (proposal: DataViewerProposal, key: string) => {
```
reefdog, line 10: This method used to have more checks in it, such as verifying `key in proposal.values`, until I realized (with the help of my test coverage) that they weren't necessary.

So in the end it's like… what does this do? Well, for now it removes those blank values. If the API promises not to send blank values, then I'm not entirely sure we do need it?

jasonaowen (commented):
This is a bigger change than the other two, and I want to take a deeper look at it, which I unfortunately won't be able to do until late tonight. A few thoughts to get started, though!

src/utils.ts
reefdog: Yes, it should! And I guess if `getProposalValue` hangs around (tbd) it should be renamed, too.

src/utils.ts
reefdog: Oof, I think you're right… although (I hate to say it) maybe after the demo?

Do you think this would belong somewhere like [between L77 and L78](https://github.com/PhilanthropyDataCommons/data-viewer/blob/ac46a38cc9c9fac8e89bbb4826f4983f9fe4a23e/src/pdc-api.ts#L77-L78)?

I wonder what role this utility would even play, after that. Like I mentioned, the shape of this function changed quite a lot as I had tests that let me strip out internals. I'll probably be able to remove it entirely.

reefdog: Okay, I've played around with an implementation of this, where I modified `mapProposals` thusly:

```diff
-        .filter((pair): pair is [string, string] => !!pair[0])
+        .filter((pair): pair is [string, string] => (!!pair[0] && pair[1]?.trim() !== ''))
```

Is this the starting point you were envisioning, @jasonaowen? If so, I can do that as the first commit on this PR, and then modify the following changes accordingly. For instance, we lose the need to even have `getProposalValues` at all if we're guaranteeing there are no blank values.

One consequence is that we'll need to create tests for `mapProposals` to cover some of the scenarios that were being tested by `getProposalValues` and `getProposalValuesFromCandidates` tests, since [structures like this](https://github.com/PhilanthropyDataCommons/data-viewer/pull/101/files#diff-04abd0e283864d1e858d70a5f86f67ccf45fcd3b752e56b136c3eee449b06b28R18-R25) should no longer even exist. And I'd really love to have an integration test that made sure an entire "API response → `mapProposals()` → `getProposalValuesFromCandidates()`" flow worked properly, but that's probably out of scope.

jasonaowen: Awesome! Yes, that's roughly what I'd expect. I'd suggest adding a subsequent filter, though, since it's for a different purpose - the current filter is mostly a type guard that shouldn't ever filter anything out in practice, whereas the new filter is actually excluding data. Maybe something like:

```diff
         .filter((pair): pair is [string, string] => !!pair[0])
+        .filter(([key, value]) => (value.trim() !== ''))
```

> For instance, we lose the need to even have `getProposalValues` at all if we're guaranteeing there are no blank values.

Yes! That's the goal.

> One consequence is that we'll need to create tests for `mapProposals`...

The middle commit of #106 extracts `mapProposals` to its own file, which should make writing tests for it simpler; I considered doing so as part of that extraction, but decided that was a distraction that we could follow up on later.

reefdog: Somehow I forgot that this was back on my plate! 🤦 On it…

reefdog: @jasonaowen Okay! Here's what I decided to do:

1. Split off the `mapProposals()` modification (plus some tests for it) to #112 [DONE]
2. Add `getProposalValuesFromCandidates()` directly on #102 (and rebase it, etc.) [TODO]

At which point I'll close this PR.

reefdog: After way too much churn, I've realized making this change resolves the entire issue without the need for the non-idiomatic syntactic sugar that `getProposalValuesFromCandidates()` provided. Thus, #112 is able to replace _both_ this PR and #102. 🎉

reefdog:
Due to the discussion @jasonaowen started in [his review here](https://github.com/PhilanthropyDataCommons/data-viewer/pull/101#discussion_r1185597861) and the consequent work in #112, this PR is no longer necessary. Happily closing!

PR #102, opened 2023-05-04
Safely select organization name in proposal list

### LINKED ISSUE #100: Sidebar list of proposals doesn't display org name correctly

reefdog:
_This PR ladders off #101, and that one needs merging before this one._

This PR uses `getProposalValueFromCandidates()` to correctly select the organization name for display in the proposal list sidebar, fixing the issue where we were displaying empty organization names for many orgs.

**Testing:**
1. On `main`, add the sidebar to `pages/ProposalList.tsx` (see description of #94 for code)
2. `npm start` and look at `/proposals`, scrolling the sidebar list
3. You should see lots of blank cells (may be below the fold)
4. Switch to this branch
5. Now those cells should be populated with an organization name

Closes #100

### DIFF
--- a/src/components/ProposalListGridItem.tsx
+++ b/src/components/ProposalListGridItem.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { Link } from 'react-router-dom';
 import { DataViewerProposal } from '../interfaces/DataViewerProposal';
+import { getProposalValuesFromCandidates } from '../utils';
 import './ProposalListGridItem.css';
 
 interface ProposalListGridItemProps {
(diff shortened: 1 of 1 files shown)


jasonaowen (approved):
Love it!

It'd probably be good to refactor [the original organization name lookup](https://github.com/PhilanthropyDataCommons/data-viewer/blob/1662b9e1657c930df1e65a9607a980a9a84a03cd/src/pages/ProposalDetail.tsx#L47-L57) to support this, but that's operating on the API response rather than the internal version, so would be a heavier lift.

reefdog:
> It'd probably be good to refactor [the original organization name lookup](https://github.com/PhilanthropyDataCommons/data-viewer/blob/1662b9e1657c930df1e65a9607a980a9a84a03cd/src/pages/ProposalDetail.tsx#L47-L57) to support this, but that's operating on the API response rather than the internal version, so would be a heavier lift.

I took a glance at that, but what it seemed to be doing was significantly different enough that I noped out. :)

reefdog:
@jasonaowen Do you think we should tackle that heavier lift here, or merge and fix separately? I'd prefer the latter so this isn't held up, and because I think that heavier lift will require some prerequisite changes to how we approach the problem on the detail page.

jasonaowen:
Let's handle it separately! I didn't intend to suggest that it should be part of this PR, although I can see how it came across that way - sorry about that.

reefdog:
Due to the discussion @jasonaowen started in [his review of #101](https://github.com/PhilanthropyDataCommons/data-viewer/pull/101#discussion_r1185597861) and the consequent work in #112, this PR is no longer necessary. Happily closing!

PR #104, opened 2023-05-05
Refactor API client code

jasonaowen:
As I was working on supporting server-side search, and implementing the proposal list sidebar on the proposal detail page, I found that the API client code wasn't quite well suited to the task. Partly, this is because of previous changes, and partly because we're introducing new requirements.

In order to reduce the scope of these changes, make them now, before we implement the sidebar.

This is purely a refactor, and should not change the behavior of the application. See the individual commit messages for more details.

Issue https://github.com/PhilanthropyDataCommons/data-viewer/issues/15 Implement pagination on proposal lists
Issue #62 Add sidebar to Proposal Detail page
Issue https://github.com/PhilanthropyDataCommons/data-viewer/issues/83 Enable search

### DIFF
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -97,7 +97,6 @@ const useProposals = (page: string, count: string) => (
 
 export {
   useCanonicalFields,
-  usePdcApi,
   useProposal,
   useProposals,
 };
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -13,28 +13,38 @@ const throwNotOk = (res: Response): Response => {
   return res;
 };
 
-const logError = (error: unknown, path: string) => {
-  logger.error({ error }, `Error fetching ${path}`);
+const logError = (error: unknown, path: string, params: URLSearchParams) => {
+  logger.error({ error, params }, `Error fetching ${path}`);
 };
 
 // Custom React hook to make authenticated requests to the configured API
-const usePdcApi = <T>(path: string): T | null => {
+const usePdcApi = <T>(
+  path: string,
+  params: URLSearchParams = new URLSearchParams(),
+): T | null => {
   const { fetch } = useOidcFetch();
   const [response, setResponse] = useState(null);
 
   useEffect(() => {
-    fetch(new URL(path, API_URL))
+    const url = new URL(path, API_URL);
+    url.search = params.toString();
+    fetch(url)
       .then(throwNotOk)
       .then((res) => res.json())
       .then(setResponse)
-      .catch((e) => logError(e, path));
+      .catch((e) => logError(e, path, params));
 
     /* eslint-disable-next-line react-hooks/exhaustive-deps --
+     *
      * fetch should not be a dependency, because although it or its internal
      * state may change from render to render, such changes are not relevant:
      * a change in the way we make a request should not trigger an API request
+     *
+     * params is a dependency, but as an object - and often a newly-created
+     * object - depending on it directly causes spurious renders; instead, use
+     * its string value, which should be stable
      */
-  }, [path]);
+  }, [path, params.toString()]);
 
   return response;
 };
@@ -62,7 +72,12 @@ interface Proposal {
 }
 
 const useProposal = (proposalId: string) => (
-  usePdcApi<Proposal>(`/proposals/${proposalId}?includeFieldsAndValues=true`)
+  usePdcApi<Proposal>(
+    `/proposals/${proposalId}`,
+    new URLSearchParams({
+      includeFieldsAndValues: 'true',
+    }),
+  )
 );
 
 interface Proposals {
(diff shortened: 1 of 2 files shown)


src/pdc-api.ts
```diff
@@ -13,28 +13,38 @@ const throwNotOk = (res: Response): Response => {
...
-  logger.error({ error }, `Error fetching ${path}`);
+const logError = (error: unknown, path: string, params: URLSearchParams) => {
+  logger.error({ error, params }, `Error fetching ${path}`);
 };
 
 // Custom React hook to make authenticated requests to the configured API
-const usePdcApi = <T>(path: string): T | null => {
+const usePdcApi = <T>(
+  path: string,
+  params: URLSearchParams = new URLSearchParams(),
```
reefdog, line 23: Is there an advantage to `params` explicitly being a `URLSearchParams` here, instead of just an object? (As a TypeScript novice, I don't know if you're allowed to just say "it's an object, but we don't know what shape"; though of course we don't know the shape of the data passed to `URLSearchParams`, either.)

We'd have to add a line like `const searchParams = new URLSearchParams(params)` to the body of the hook, but the advantage would be a cleaner invocation of `usePdcApi()`.

Is that possible / a good idea?

src/pdc-api.ts
```diff
@@ -62,25 +72,28 @@ interface Proposal {
 }
 
 const useProposal = (proposalId: string) => (
-  usePdcApi<Proposal>(`/proposals/${proposalId}?includeFieldsAndValues=true`)
+  usePdcApi<Proposal>(
+    `/proposals/${proposalId}`,
+    new URLSearchParams({
+      includeFieldsAndValues: 'true',
+    }),
```
reefdog, line 79: Just to be explicit, my suggestion above would mean this could just become:

```suggestion
    {
      includeFieldsAndValues: 'true',
    },
```

src/pdc-api.ts
```diff
@@ -13,28 +13,38 @@ const throwNotOk = (res: Response): Response => {
...
 
     /* eslint-disable-next-line react-hooks/exhaustive-deps --
+     *
      * fetch should not be a dependency, because although it or its internal
      * state may change from render to render, such changes are not relevant:
      * a change in the way we make a request should not trigger an API request
+     *
+     * params is a dependency, but as an object - and often a newly-created
+     * object - depending on it directly causes spurious renders; instead, use
+     * its string value, which should be stable
```
reefdog, line 45: Wow, good eye. Not sure I'd have realized this is what was going on!

reefdog (approved):
I like it, all good changes!

I have a question (+suggestion depending on the answer) about how often we're having to invoke `new URLSearchParams`, but I throw it completely back in your court.

src/pdc-api.ts
jasonaowen: Definitely possible! In fact, I went back and forth on that point, and tried it both ways. The advantage of passing a `URLSearchParams` instead of a `Record<string, string>` is that the former allows for multiple instances of the same parameter, such as `?foo=bar&foo=baz`, which the latter cannot represent. In practice, that'd look like:

```ts
const params = new URLSearchParams();
params.append('foo', 'bar');
params.append('foo', 'baz');
const example = usePdcApi('/example', params);
```

That seems like it could be relevant for the someday future of faceted search, depending on how we end up encoding such queries into the API... but yes, I'll buy the argument that we don't need that complexity now, and could cross that bridge when we come to it.

src/pdc-api.ts
jasonaowen: I noticed it when my network tab starting going nuts! See also [Removing unnecessary object dependencies ](https://react.dev/reference/react/useEffect#removing-unnecessary-object-dependencies).

src/pdc-api.ts
jasonaowen: Oh, and actually now that I try a second time to write it this way, I remember the more compelling reason: using the object inside of the effect gets to be kind of a pain.

We can't use the object as one of the `useEffect` dependencies for the same reason we can't use the `URLSearchParams`: because [object literals are different on every render](https://react.dev/reference/react/useEffect#removing-unnecessary-object-dependencies). `.toString()` just returns `'[object Object]'`, so if we wanted to take the same string based approach we'd need to `JSON.stringify(params)`, which is somewhat feasible but feels far afield.

We could construct the `URLSearchParams` outside of the `useEffect` function, but then we need another name, and it feels slightly wrong to build the things needed for the effect outside of the effect. I say only "slightly" because we *are* doing that work outside of the effect in this approach, but it's far enough away that somehow it seems less bad, which is of course very silly.

> the advantage would be a cleaner invocation of `usePdcApi()`

Ultimately, where I landed is that `usePdcApi` is an internal helper, and the only callers of it should be in this file. (Side note: we shouldn't be exporting it, then!) In this case, a slightly more awkward invocation feels like a reasonable trade-off for a slightly more elegant body.

jasonaowen: > (Side note: we shouldn't be exporting [`usePdcApi`], then!)

Export removed in 0ff5d44!

jasonaowen:
Thanks for the thoughtful review, @reefdog! Hopefully my answers were persuasive. I think I'd like to lean towards keeping the current approach, and it sounds like you'd be okay with that decision, so I'll merge this shortly.

PR #106, opened 2023-05-05
 Add proposal list sidebar to proposal detail page 

### LINKED ISSUE #62: Add sidebar to Proposal Detail page

jasonaowen:
Use the new `<ProposalListGridPanel>` component introduced in #94 in a sidebar on the proposal detail page.

See the individual commits for more details.

Resolves #62 Add sidebar to Proposal Detail page

### DIFF
--- a/src/pages/ProposalDetail.tsx
+++ b/src/pages/ProposalDetail.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect } from 'react';
-import { OidcSecure } from '@axa-fr/react-oidc';
+import { withOidcSecure } from '@axa-fr/react-oidc';
 import { useParams } from 'react-router-dom';
 import {
   CanonicalField,
@@ -56,7 +56,7 @@ const getApplicant = (
     ?? 'Unknown Applicant'
 );
 
-const ProposalDetail = () => {
+const ProposalDetailLoader = () => {
   const params = useParams();
   const proposalId = params.proposalId ?? 'missing';
   const canonicalFields = useCanonicalFields();
@@ -74,20 +74,18 @@ const ProposalDetail = () => {
 
   if (canonicalFields === null || proposal === null) {
     return (
-      <OidcSecure>
-        <PanelGrid>
-          <PanelGridItem>
-            <ProposalDetailPanel
-              proposalId={0}
-              title="Loading..."
-              applicant="Loading..."
-              applicantId="00-0000000"
-              version={0}
-              values={[]}
-            />
-          </PanelGridItem>
-        </PanelGrid>
-      </OidcSecure>
+      <PanelGrid>
+        <PanelGridItem>
+          <ProposalDetailPanel
+            proposalId={0}
+            title="Loading..."
+            applicant="Loading..."
+            applicantId="00-0000000"
+            version={0}
+            values={[]}
+          />
+        </PanelGridItem>
+      </PanelGrid>
     );
   }
 
@@ -99,21 +97,20 @@ const ProposalDetail = () => {
   const values = mapCanonicalFields(canonicalFields, proposal);
 
   return (
-    <OidcSecure>
-      <PanelGrid>
-        <PanelGridItem>
-          <ProposalDetailPanel
-            proposalId={proposal.id}
-            title={title}
-            applicant={applicant}
-            applicantId={applicantId}
-            version={version}
-            values={values}
-          />
-        </PanelGridItem>
-      </PanelGrid>
-    </OidcSecure>
+    <PanelGrid>
+      <PanelGridItem>
+        <ProposalDetailPanel
+          proposalId={proposal.id}
+          title={title}
+          applicant={applicant}
+          applicantId={applicantId}
+          version={version}
+          values={values}
+        />
+      </PanelGridItem>
+    </PanelGrid>
   );
 };
 
+const ProposalDetail = withOidcSecure(ProposalDetailLoader);
 export { ProposalDetail };
--- a/src/pages/ProposalDetail.tsx
+++ b/src/pages/ProposalDetail.tsx
@@ -6,9 +6,36 @@ import {
   Proposal,
   useCanonicalFields,
   useProposal,
+  useProposals,
 } from '../pdc-api';
+import { mapProposals } from '../map-proposals';
 import { PanelGrid, PanelGridItem } from '../components/PanelGrid';
 import { ProposalDetailPanel } from '../components/ProposalDetailPanel';
+import { ProposalListGridPanel } from '../components/ProposalListGridPanel';
+
+interface ProposalListGridLoaderProps {
+  canonicalFields: CanonicalField[] | null;
+}
+
+const ProposalListGridLoader = (
+  { canonicalFields }: ProposalListGridLoaderProps,
+) => {
+  const proposals = useProposals('1', '1000');
+
+  if (canonicalFields === null || proposals === null) {
+    return (
+      <div>Loading data...</div>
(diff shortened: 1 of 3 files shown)


src/pages/ProposalDetail.tsx
```diff
@@ -99,21 +127,33 @@ const ProposalDetail = () => {
...
+        canonicalFields={canonicalFields}
+      />
+      <ProposalDetailPanelLoader
+        canonicalFields={canonicalFields}
+      />
+    </PanelGrid>
   );
 };
 
+const ProposalDetail = withOidcSecure(ProposalDetailLoader);
```
reefdog, line 158: I think the HOC approach is great, good change!

The `Loader` suffix, however, doesn't feel quite right. "Loader" implies (to me) a narrow scope of loading data, whereas this is truly creating components. They just happen to be components that need additional [something] before exporting.

So the goals as I understand them:
- Create a component, which needs securing
- Secure it
- Export it with a sensible name

One option that feels a little gymnastic but actually makes sense to me is calling the unsecured component `ProposalDetail`, suffix the _secured_ version, but then export it "as" the unsecured name. That's… maybe dangerout, but seems to work? E.g.:

```ts
const ProposalDetail = () => {
 // Build unsecured component
}
const ProposalDetailSecured = withOidcSecure(ProposalDetail);
export { ProposalDetailSecured as ProposalDetail };
```

If that's too gross, then I'd suggest either:
1. Doing what you're doing but with `Unsecured` instead of `Loader`, or
2. Export the component as `ProposalDetailSecured` and use it that way.

Or… tell me that no, `Loader` is the right name after all! 😆

reefdog (approved):
I left a comment about the `Loader` terminology, but I consider that very much non-blocking. This is great. 🙏

src/pages/ProposalDetail.tsx
jasonaowen: Yes, that's a great summary of the goals!

> One option that feels a little gymnastic but actually makes sense to me is calling the unsecured component `ProposalDetail`, suffix the _secured_ version, but then export it "as" the unsecured name. That's… maybe dangerous, but seems to work?

I would prefer not to do that - I'm not sure I'd go so far as to call it dangerous, but it's definitely confusing! One of the reasons we have the `no-default-export` rule enabled is to avoid that kind of acrobatics, because calling something one name in one place and another name in another place makes it more difficult to understand the code.

> Doing what you're doing but with `Unsecured` instead of `Loader`

Yeah, I considered that, but I feel like that's kind of too... disparaging? And it's not exactly that it's insecure, it's that it's meant to be used inside a secured component.

> Export the component as `ProposalDetailSecured`

I also considered that, and rejected it for the same reasons you did!

> Or… tell me that no, `Loader` is the right name after all! :laughing:

No, I'm also not 100% happy with `Loader`. I'll think about this some more.

jasonaowen: Another option I considered and discarded was to use `withOidcSecure` in the route definition. This is what the [react-oidc docs suggest](https://github.com/AxaFrance/react-oidc/tree/master/packages/react#how-to-secure-a-component--hoc-method):

```ts
    <Route path="/dashboard" component={withOidcSecure(Dashboard)} />
```

However, the [react-router docs](https://reactrouter.com/en/main/route/route) warn

> When using `RouterProvider`, if you do not wish to specify a React element (i.e., `element={<MyComponent />}`) you may specify a `Component` instead (i.e., `Component={MyComponent}`) and React Router will call `createElement` for you internally. You should only do this for `RouterProvider` applications though since using `Component` inside of `<Routes>` will de-optimize React's ability to reuse the created element across renders.

We are not currently using a `RouterProvider`, although [react-router recommends doing so](https://reactrouter.com/en/main/routers/picking-a-router). I looked at that while I was setting up routing initially, and wasn't really sold on their data APIs concept, in part because it seemed difficult to combine with react-oidc's fancy-`fetch`-via-hook interface.

Maybe that performance warning is worth ignoring, though?

reefdog: > > Doing what you're doing but with Unsecured instead of Loader
> Yeah, I considered that, but I feel like that's kind of too... disparaging? And it's not exactly that it's insecure, it's that it's meant to be used inside a secured component.

omg, I actually wrote and then edited out a note about why `Unsecure` was intentional so as to avoid the disparaging implication of "insecure". 😆 Do you think it's still too disparaging?

> Another option I considered and discarded was to use withOidcSecure in the route definition.…

Oof, I leave the choice about `RouterProvider` up to you. One note from the look of it is that we would always be securing whole pages this way, right? Do we ever expect (and does react-oidc even support) only securing individual components on a page, and would this in any way prevent us from doing that?

----

Do think on it just a bit longer but if you can't think of anything you like better than `Loader` or `Unsecured` then let's just leave the bike shed beige!

jasonaowen: I did a [survey on GitHub](https://github.com/search?q=withOidcSecure&type=code), and saw three patterns for using `withOidcSecure`:

- the documented pattern with react-router 5 or earlier: `<Route component={withOidcSecure(Foo)} />`
- prefix `Protected`: `const ProtectedFoo = withOidcSecure(Foo)`
- prefix `Secured`: `const SecuredFoo = withOidcSecure(Foo)`

So, seems like the option 2 you outlined is what other users have done. Does that sway you at all?

For now, I'll sleep on it rather than spinning my wheels further.

reefdog: I'm gonna sidebar on insecure/unsecured for a moment: I realize now I forgot to mention that while the inverse of "secure" in our context is obviously "insecure", I went the "unsecured" route because it seemed closer to a secured/unsecured loan. Perhaps the protected/unprotected terminology your research uncovered makes for better / less disparaging terminology.

> So, seems like the option 2 you outlined is what other users have done. Does that sway you at all?

I think we're in agreement that exporting them as an alias isn't idiomatic (for our platform), so are you talking about actually using `SecuredFoo`/`ProtectedFoo` around the app (and renaming the files)? 🤔 I mean, it's not a bad idea, I guess. It does make it terribly obvious that they're secured.

I guess I leave it up to you, truly. Including kicking the whole conversation into a well.

jasonaowen: Okay, after sleeping on it, I came up with another option!

```ts
const Foo = withOidcSecure(() => {
  ...
});
```

...unfortunately, that leaves the React component tree with an `Anonymous` component:

image

I think I'm just going to name them `Foo / SecuredFoo`, export and use `SecuredFoo`, and be done.

jasonaowen: ...which leaves us with the next problem: what to name the component that contains both a `<ProposalDetailPanel>` and a `<ProposalListGrid>` to share the fields state?

I'm afraid I'm coming back around to your original suggestions, that we should just live with this suboptimal name as-is, because this isn't worth the time.

Issue #108, opened 2023-05-05
Rename shared TypeScript interfaces for clarity


reefdog:
In #99, I introduced `src/interfaces/DataViewerProposal.ts` as a named and shared TypeScript interface. I used the `DataViewer` prefix to distinguish it from the existing `Proposal` interface that represents the proposal as modeled by the API.

@jasonaowen [pointed out](https://github.com/PhilanthropyDataCommons/data-viewer/pull/99#pullrequestreview-1414029214) that this is inverted, and we should rename `Proposal` to `ApiProposal` (and do the same for other API-specific interfaces), and let `DataViewerProposal` (and other first-party, local-to-this-project interfaces) go prefixless.

For expedience and to reduce conflicts with other branches, I marched onward with the merge but we should make this change ASAP.

reefdog:
#110 renames the API-modeled interfaces, but we still need to determine a good name for `DataViewerProposal`. This is the simplified shape of a single version of a proposal ([the most recent version](https://github.com/PhilanthropyDataCommons/data-viewer/blob/6e1f242dec0bc3fba096786fc5f98ffc00d5853b/src/map-proposals.ts#L15)):

```json
{
  "id": "248",
  "values": {
    "organization_name": ["ABC 123"],
    "board_of_directors": ["John Smith", "Jane Doe"]
  }
}
```

So here we have:
- ID of this proposal
- The shortCode for each field (as the `values` key)
- An array of values for that field (since multiple values can be assigned to a single field)

We aren't capturing:
- Explicit `position` order
- Multiple versions

We currently only use this to compose proposal list views (tables, grids) where we don't need deep version information.

Semantically, what I think this name needs to communicate
- This data has been prepared/shaped for a certain purpose
   - Do we embed this purpose into the name? Or have a generic name like `Processed`? I lean former.
- This is only a single version of data

jasonaowen:
Excellent summary, @reefdog!

I like the idea of encoding into the name that this is for a particular purpose. Maybe something like `ProposalListItem`? If so, perhaps it would be worthwhile to also rename `<ProposalListGridItem>` to avoid confusion - maybe `<ProposalListGridCell>` or `<ProposalListGridRow>`?

reefdog:
[...]

PR #110, opened 2023-05-08
Rename API interface type definitions

jasonaowen:
In PR #99, we introduced a name for the display-centric models we are using. The name introduced, `DataViewerProposal`, is prefixed to distinguish itself from the type definitions for how the API represents data.

Rename the API-centric interfaces to include an `Api` prefix.

These new names make it clearer that we should not be using these interfaces so widely, but cleaning that up will be part of future work.

Issue #108 Rename shared TypeScript interfaces for clarity

### DIFF
--- a/src/map-proposals.ts
+++ b/src/map-proposals.ts
@@ -1,4 +1,4 @@
-import type { CanonicalField, Proposal } from './pdc-api';
+import type { ApiCanonicalField, ApiProposal } from './pdc-api';
 
 const extendMultimapReducer = (
   multimap: Record<string, string[]>,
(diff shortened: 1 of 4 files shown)


jasonaowen:
Side note: after doing this refactor, I think it is still useful to have some kind of prefix for the processed data, especially so that you can search for one type while excluding the other. `DataViewerProposal` is perhaps not it, not least because that interface / data structure is narrowly scoped to list contexts - it doesn't contain all the data needed for `<ProposalDetailPanel>` - but we can come up with another name for that separately.

PR #111, opened 2023-05-08
Use server search API

jasonaowen:
Rather than always loading the full list of proposals from the API and doing basic string matching locally, instead use the recently introduced search API. The server search API includes such fanciness as word stemming (so that a query for "summaries" would match a proposal containing "summary") and operators (like "a OR b" which would match proposals containing either).

Note that we have not yet implemented pagination, so this is somewhat of a tradeoff: the initial page load will get the full list of proposals, and searching will cause another server round trip, and clearing the search will cause a third round trip. However, this is a necessary step on the way to pagination.

This should not be deployed (read: merged) until https://github.com/PhilanthropyDataCommons/service/pull/320 and https://github.com/PhilanthropyDataCommons/service/pull/337 have been deployed.

Issue #15 Implement pagination on proposal lists
Issue #83 Enable search
Issue #115 Show message on "no results" searches

### DIFF
--- a/src/components/ProposalListTablePanel.tsx
+++ b/src/components/ProposalListTablePanel.tsx
@@ -16,6 +16,7 @@ interface ProposalListTablePanelProps {
   proposals: DataViewerProposal[];
   onSearch: (query: string) => void;
   searchQuery?: string;
+  loading?: boolean;
 }
 
 // For now, we are hard-coding this list.
@@ -45,10 +46,13 @@ export const ProposalListTablePanel = ({
   proposals,
   onSearch,
   searchQuery = '',
+  loading = false,
 }: ProposalListTablePanelProps) => {
   const [wrap, setWrap] = useState(false);
 
   const handleWrapClick = () => setWrap((previous) => !previous);
+
+  const hasSearchQuery = searchQuery !== '';
 
   return (
     <Panel>
@@ -67,12 +71,18 @@ export const ProposalListTablePanel = ({
         </PanelActions>
       </PanelHeader>
       <PanelBody>
-        <ProposalListTable
-          fieldNames={fieldNames}
-          proposals={proposals}
-          columns={DEFAULT_COLUMNS}
-          wrap={wrap}
-        />
+        {loading ? (
+          <div className="panel-message">
+            {hasSearchQuery ? 'Searching…' : 'Loading…'}
+          </div>
+        ) : (
+          <ProposalListTable
+            fieldNames={fieldNames}
+            proposals={proposals}
+            columns={DEFAULT_COLUMNS}
+            wrap={wrap}
+          />
+        )}
       </PanelBody>
     </Panel>
   );
--- a/src/components/ProposalListTablePanel.tsx
+++ b/src/components/ProposalListTablePanel.tsx
@@ -69,15 +69,17 @@ export const ProposalListTablePanel = ({
         <PanelActions>
           <Search onSearch={onSearch} initialQuery={searchQuery} />
         </PanelActions>
-        <PanelActions>
-          <Button
-            onClick={handleWrapClick}
-            color={wrap ? 'blue' : 'gray'}
-          >
-            <Bars3BottomLeftIcon className="icon" />
-            Toggle wrapping
-          </Button>
-        </PanelActions>
+        {hasProposals && (
+          <PanelActions>
+            <Button
+              onClick={handleWrapClick}
+              color={wrap ? 'blue' : 'gray'}
+            >
+              <Bars3BottomLeftIcon className="icon" />
+              Toggle wrapping
+            </Button>
+          </PanelActions>
+        )}
       </PanelHeader>
       <PanelBody>
         {hasProposals ? (
--- a/src/components/ProposalListTablePanel.tsx
+++ b/src/components/ProposalListTablePanel.tsx
@@ -52,7 +52,16 @@ export const ProposalListTablePanel = ({
 
   const handleWrapClick = () => setWrap((previous) => !previous);
 
+  const hasProposals = proposals.length > 0;
   const hasSearchQuery = searchQuery !== '';
+
+  const generateFallbackMessage = () => {
+    if (hasSearchQuery) {
+      return loading ? 'Searching…' : 'No search results for that query.';
+    }
+
+    return loading ? 'Loading…' : 'No data available.';
+  };
 
   return (
     <Panel>
@@ -71,17 +80,17 @@ export const ProposalListTablePanel = ({
         </PanelActions>
       </PanelHeader>
       <PanelBody>
-        {loading ? (
-          <div className="panel-message">
-            {hasSearchQuery ? 'Searching…' : 'Loading…'}
-          </div>
-        ) : (
+        {hasProposals ? (
           <ProposalListTable
             fieldNames={fieldNames}
             proposals={proposals}
             columns={DEFAULT_COLUMNS}
             wrap={wrap}
           />
+        ) : (
+          <div className="panel-message">
+            {generateFallbackMessage()}
+          </div>
         )}
       </PanelBody>
     </Panel>
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -26,6 +26,7 @@ const usePdcApi = <T>(
   const [response, setResponse] = useState(null);
 
   useEffect(() => {
+    setResponse(null);
     const url = new URL(path, API_URL);
     url.search = params.toString();
     fetch(url)
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -87,13 +87,15 @@ interface ApiProposals {
 
 const PROPOSALS_DEFAULT_PAGE = '1';
 const PROPOSALS_DEFAULT_COUNT = '1000';
+const PROPOSALS_DEFAULT_QUERY = '';
 
-const useProposals = (page: string, count: string) => (
+const useProposals = (page: string, count: string, query: string) => (
   usePdcApi<ApiProposals>(
     '/proposals',
     new URLSearchParams({
       _page: page,
       _count: count,
+      _content: query,
     }),
   )
 );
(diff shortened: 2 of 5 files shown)


src/pdc-api.ts
```diff
@@ -85,12 +85,13 @@ interface Proposals {
   total: number;
 }
 
-const useProposals = (page: string, count: string) => (
+const useProposals = (page: string, count: string, query: string) => (
```
reefdog, line 88: Could we default to `query = ''` so this can be an optional parameter?

In fact, I think it makes sense for `page` to be optional (defaulting to `1`), too. For practical purposes (i.e. discouraging countless queries) it makes sense to have a default `count`.

What do you think?

jasonaowen: One or two optional parameters can be helpful, but making all three optional starts to be confusing!

Right now, there are only two call sites: `<ProposalListLoader>` and `<ProposalListGridLoader>`. The latter currently has placeholder values, but once we figure out how we want to do sidebar search & pagination, that'll change.

In `<ProposalListLoader>`, and eventually in `<ProposalListGridLoader>`, we'll always be pulling out the search params from the data-viewer URL and passing them along to `useProposals`, so I don't think we'd ever take advantage of the optional parameter. Passing the possibly-`null` values returned by `searchParams.get('key')` into this hook complicates the hook somewhat.

In general, I think I'd prefer to initialize default values as close to when we name their variables as possible. Passing around "missing" values is harder (for me) to think about, and makes working with such values more difficult.

That said, something you might have been hinting at is that we should have a single constant for each default value, rather than repeating magic numbers (which are strings, in this case). Again, the call in `<ProposalListGridLoader>` is some intentional tech debt, but that's no reason not to start naming and using constants now! I will fix.

@reefdog is that at all convincing?

jasonaowen:
I rebased this on `main`, extracted constants for the default values, and updated Keycloak to allow logging in to this deploy preview.

src/pdc-api.ts
reefdog: I did _also_ think about specifying the defaults as constants, so I'm glad you did that!

I was more thinking the ergonomics of calling `useProposals()` and how, even if we will always be invoking it with all three params, there's nothing inherently required about any of them (that I'm aware of). Still, I'm happy to let this one go and call it resolved! I do understand your logic and am not overly concerned.

reefdog (changes requested):
👏 I love how relatively light-touch this was!

Since search is no longer "instantaneous", we need to disable the search form while a search is in progress — I'm leaving that language intentionally naive to communicate the UX goal without suggesting an implementation — before this lands in production. UI-wise, it's sufficient for now to just set the `disabled` attribute on the search submit `Button`, i.e. [expand this conditional](https://github.com/PhilanthropyDataCommons/data-viewer/blob/8dbac8704c65f4e892e6576121d256611371070d/src/components/Search.tsx#L34).

Just like ["permanent, temporary, and situational disabilities" w/r/t accessibility](https://digital.gov/resources/introduction-accessibility/), we should consider permanent, temporary, and situational bandwidth/performance issues and how they affect the UX. Throttling the network to 2G speeds demonstrated how a lack of feedback could make the user think they needed to keep mashing the submit button.

(I would _also_ like us to add a global on-page loading indicator (leaning toward GitHub/etc.'s "thin blue line") during API fetches, but that's definitely out of scope of this PR.)

jasonaowen:
> Since search is no longer "instantaneous", we need to disable the search form while a search is in progress

Oh, that is an excellent callout - thank you for flagging that, @reefdog! I will fix.

jasonaowen:
@reefdog okay, I see an easy approach and a harder approach!

The easy approach is arguably a bug fix: update `usePdcApi` to include a `setResponse(null)` at the beginning of its `useEffect`. It probably shouldn't continue to return the old API response for a while after a change that triggers a new API request, and this would fix that bug. Here's how it looks in practice, and please excuse the terrible sample data:

[Screencast showing the loading screen coming up between searches](https://github.com/PhilanthropyDataCommons/data-viewer/assets/1494855/4157bf92-cf6b-4c03-8938-c0f4f58ceed2)

That's with a 2-second sleep added to the API, so the transition is more obvious. We can - and probably should - iterate on that loading screen, but we could defer that work while avoiding double submits in the meantime. Now, this is arguably more correct, but maybe a worse search experience; what do you think?

The harder approach is one I haven't fully thought through, but broadly: keep track of pending requests somehow, which would allow us to pass a flag to the `<Search>` component to prevent form submission.

reefdog:
> The easy approach is arguably a bug fix: update `usePdcApi` to include a `setResponse(null)` at the beginning of its `useEffect`. It probably shouldn't continue to return the old API response for a while after a change that triggers a new API request, and this would fix that bug.

I _think_ this is correct and the right way to go. UX-wise, this is certainly more jarring (especially with our current loading indicator, as you said), but I just checked and it's the same thing Google does on its search results, so it's at least a pattern! Are there other places in our UI where this would be apparent/effected, and does it work fine there too?

My only concern would be state restoration in the event of an error (if we want that). What's the right UX if someone has searched for something — or, in future, arranged a complex filter — and then tweaks parameters/terms, submits it, and it errors out? I would think we should maintain or restore their previous state, but with an error message about the new thing. I _think_. This approach would seem to make that impossible without a restoration mechanism.

That's food for thought, not an objection to this approach. Go forth!

jasonaowen:
> Are there other places in our UI where this would be apparent/effected, and does it work fine there too?

Great question! It would also cause a loading indicator on the proposal detail page when navigating from one proposal directly to another, via either the sidebar list and the dev-only prev/next buttons.

> My only concern would be state restoration in the event of an error (if we want that). What's the right UX if someone has searched for something — or, in future, arranged a complex filter — and then tweaks parameters/terms, submits it, and it errors out? I would think we should maintain or restore their previous state, but with an error message about the new thing. I _think_. This approach would seem to make that impossible without a restoration mechanism.
> 
> That's food for thought, not an objection to this approach. Go forth!

Right now, the restoration mechanism is the back button! And, if we continue to use query parameters for storing such state - as I currently intend to - that would continue to be a remediation.

You are right to bring up error handling, which currently we're just not doing. We should explicitly say to click the back button as part of such error messages!

reefdog:
Sounds good to me!

jasonaowen:
Fixed that loading bug, and rebased onto main!

reefdog (commented):
Playing with the preview, I have one minor UX concern and one significant one.

**Minor:** When a search result set is empty, we don't show any message. This is not a regression from the on-page filter-search, so we can punt to an enhancement ticket.

**Significant:** Losing the entire panel body during search feels jarring and slow. This is exactly what you described/demonstrated would happen with closing the `setResponse` bug, but now experiencing it in person, I think we need to fully resolve this before releasing server search. Happily, I have a solution and am happy to commit it, if you agree!

1. Render the panel even while data is loading.
2. Move the "Loading…" message into the body of the panel. (With a visible loading message that displays both on initial render and during searches, I don't mind the search form not disabling.)

With these two changes, the experience is significantly improved. I don't have any concerns with adding those changes to this PR since it reflects the full change necessary to switch to server search. But, if you disagree, I can open a follow-up PR.

reefdog:
@jasonaowen To be clear, I'm super impressed by how little actually needed changing to switch from an on-page filter to API search. That is evidence of a fantastic design, and my UX critiques are intended as collaborative steps toward a completed feature, not a critique.

jasonaowen:
> To be clear, I'm super impressed by how little actually needed changing to switch from an on-page filter to API search. That is evidence of a fantastic design, and my UX critiques are intended as collaborative steps toward a completed feature, not a critique.

Thank you for saying so! I hadn't taken it in any other light, but direct praise is always welcome. :blue_heart: 

> Playing with the preview, I have one minor UX concern and one significant one.
> 
> **Minor:** When a search result set is empty, we don't show any message. This is not a regression from the on-page filter-search, so we can punt to an enhancement ticket.

Yes, that is something I've noticed and thought about. I hadn't opened an issue for it yet, but here we go: https://github.com/PhilanthropyDataCommons/data-viewer/issues/115 .

> **Significant:** Losing the entire panel body during search feels jarring and slow. This is exactly what you described/demonstrated would happen with closing the `setResponse` bug, but now experiencing it in person, I think we need to fully resolve this before releasing server search. Happily, I have a solution and am happy to commit it, if you agree!
> 
>     1. Render the panel even while data is loading.
> 
>     2. Move the "Loading…" message into the body of the panel. (With a visible loading message that displays both on initial render and during searches, I don't mind the search form not disabling.)
> 
> 
> With these two changes, the experience is significantly improved. I don't have any concerns with adding those changes to this PR since it reflects the full change necessary to switch to server search. But, if you disagree, I can open a follow-up PR.

Yes, that sounds like a great approach, and yes, I'd be happy for you to do the work! Thank you for the suggestion and the volunteering, @reefdog!

reefdog:
@jasonaowen Super! I took a break to work in my garage, but I'll try to have my work committed be EOD.

reefdog:
@jasonaowen Okay, a little later than I intended, but I've pushed out the changes to the UI so that the panel is rendered even while the data is loading. This does depend on a `loading` prop which I'm not wild about — it feels like we should have a more generalized solution — but I'm okay with learning as we go.

I also arguably slipped in two things that could have been part of a follow-up PR: #115 and conditional rendering of panel actions. However, my excuse is that both issues are so much more noticeable with server search. (And admittedly, I already had them done.)

src/components/ProposalListTablePanel.tsx
```diff
@@ -69,15 +69,17 @@ export const ProposalListTablePanel = ({
...
-        <PanelActions>
-          <Button
-            onClick={handleWrapClick}
-            color={wrap ? 'blue' : 'gray'}
-          >
-            <Bars3BottomLeftIcon className="icon" />
-            Toggle wrapping
-          </Button>
-        </PanelActions>
+        {hasProposals && (
```
jasonaowen, line 72: Oh great catch; I would have missed that!

jasonaowen:
Rebased onto `main`.

reefdog (approved):
This is in great shape. 👏 Thanks @jasonaowen!

jasonaowen:
Thanks for all your hard work on it, @reefdog!

PR #112, opened 2023-05-09
Filter blank proposal values

### LINKED ISSUE #100: Sidebar list of proposals doesn't display org name correctly

reefdog:
_This PR is an extraction of work and discussion from #101. It essentially replaces both #101 and #102._

This PR adds mock data and tests for the `mapProposals()` function, and then modifies that function to filter out blank strings when mapping API data to our internal data structure for proposal lists.

**Testing:**
1. `npm run test`
2. If you'd like, too, you can remove the `.filter((pair) => (pair[1].trim() !== ''))` line to see the test fail.
3. As this actually resolves the sidebar issue in #100, you can also `npm start` and load up a proposal detail page; no more proposals in the sidebar should have blank organization names.

**Incidentally…**

The intention of this PR was to extract some [great advice](https://github.com/PhilanthropyDataCommons/data-viewer/pull/101#discussion_r1185597861) from @jasonaowen in his review of #101, then rewrite #102 to use it. But after doing this work, I realized #102 (and the whole "get best proposal values from a list of keys" approach) was becoming a whole lot of code to add very little syntactic sugar, and that filtering blank values at the source (as this PR demonstrates) actually solves the problem without further changes needed.

Closes #100

### DIFF
--- a/src/map-proposals.test.ts
+++ b/src/map-proposals.test.ts
@@ -1,0 +1,93 @@
+import {
+  ApiCanonicalField,
+  ApiProposal,
+} from './pdc-api';
+import { mapProposals } from './map-proposals';
+
+const canonicalFields: ApiCanonicalField[] = [
+  {
+    id: 1,
+    label: 'Organization Name',
+    shortCode: 'organization_name',
+  },
+  {
+    id: 2,
+    label: 'Organization City',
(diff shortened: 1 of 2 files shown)


jasonaowen (approved):
This is great - very nicely done, @reefdog!

PR #113, opened 2023-05-11
Rename canonical fields to base fields

jasonaowen:
Update the names in our code to follow the rename from "canonical fields" to "base fields", including the API endpoint. The user-facing strings used in the UI were previously changed in PR #105.

This must be deployed in coordination with https://github.com/PhilanthropyDataCommons/service/pull/334 , as deploying one without the other will break the front-end.

Issue https://github.com/PhilanthropyDataCommons/service/issues/275

PR #114, opened 2023-05-11
Fix applicant name display

### LINKED ISSUE #109: Display applicant name from candidate fields
There are multiple times we want to display an organization name using the fallback candidate list:

1. In the "Organization Name" column on the proposal list table
   - Will need to render the `RowCell` in a special way when `shortCode` === `organization_name`:
   https://github.com/PhilanthropyDataCommons/data-viewer/blob/5d7a99f8838fbc389bb5149f46f1fb173d818585/src/components/ProposalListTable.tsx#L18-L38
2. As the main text element in the proposal list sidebar (spun off as its own issue #100)
3. Atop the Proposal Detail page
   - Will need to modify [`getApplicant()`](https://github.com/PhilanthropyDataCommons/data-viewer/blob/5d7a99f8838fbc389bb5149f46f1fb173d818585/src/pages/ProposalDetail.tsx#L47-L57) to populate from one of our sanitized `DataViewerProposals` rather than the API's `Proposal`. It's possible there are other places on this page we also want to do this.

One place we notably _don't_ want to use this logic is when composing the Proposal Detail table itself, where we want each proposal attribute to appear, including all variations of org name.

reefdog:
_This PR changes some LOC that contain the "canonical" terminology changed by #113. Whichever PR is merged into `main` first — and I assume it's #113 but that's dependent on the API deployment — will cause conflicts with the other. That's fine, we'll resolve._

#112 resolved display of applicant name in the sidebar, but as documented in #109, there were still two more places that blank-string applicant names could mess up display:

- In the proposal list table's "Organization Name" column
- In the panel header atop proposal detail pages (and the `<title>` of that page)

This PR resolves both remaining places. It introduces a `getPreferredApplicantNameValues()` utility (yes, similar to the `getProposalValuesFromCandidates()` utility that I abandoned earlier) + tests, uses it, and then uses a separate method of resolving the issue on panel headers where we don't use the `DataViewerProposal` data structure.

**Testing:**
- Switch to `main`, fire up the app, and notice that…
   - The proposal list table (`/proposals`) contains a bunch of blank Organization Names
   - Some proposals (e.g., `/proposals/406` on the production API) are missing a title in the panel header.
- Switch to this branch and check those pages again. Now each should have data where once there was none!
- And of course, `npm run test`

Closes #109

### DIFF
--- a/src/components/ProposalListTable.tsx
+++ b/src/components/ProposalListTable.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { useNavigate } from 'react-router-dom';
 import { DataViewerProposal } from '../interfaces/DataViewerProposal';
+import { getPreferredApplicantNameValues } from '../utils/proposals';
 import {
   Table,
   TableHead,
@@ -25,12 +26,26 @@ const ProposalListTableRow = ({
     navigate(`/proposals/${proposal.id}`);
   };
 
+  const getProposalCellContents = (shortCode: string) => {
+    let proposalValue;
+
+    switch (shortCode) {
+      case 'organization_name':
+        proposalValue = getPreferredApplicantNameValues(proposal);
+        break;
+      default:
+        proposalValue = proposal.values[shortCode];
+    }
+
+    return proposalValue;
+  };
+
   return (
     <TableRow onClick={handleRowClick}>
       {/* Iterate over columns to ensure order. */}
       {columns.map((shortCode) => (
         <RowCell key={shortCode}>
-          {proposal.values[shortCode]}
+          {getProposalCellContents(shortCode)}
         </RowCell>
       ))}
     </TableRow>
--- a/src/utils/proposals.ts
+++ b/src/utils/proposals.ts
@@ -1,0 +1,38 @@
+import { DataViewerProposal } from '../interfaces/DataViewerProposal';
+
+const PROPOSAL_APPLICANT_NAME_CASCADE = [
+  'organization_name',
+  'organization_dba_name',
+  'organization_legal_name',
+  'proposal_primary_contact_name',
+  'proposal_submitter_name',
+];
+
+const PROPOSAL_APPLICANT_NAME_FALLBACK = 'Unknown Applicant';
+
+/**
+   * Proposals may store the applicant name in one of several fields.
+   * This utility returns iterates through the list of attributes where
+   * we expect to find a name, in a preferential order, and returns the
+   * first matching value.
+   *
+   * If none is found, falls back to the fallback value.
+   *
+   * @param  {DataViewerProposal} proposal
+   * @return {(Array)} The applicant name value array.
+   */
+const getPreferredApplicantNameValues = (proposal: DataViewerProposal) => {
+  const bestNameKey = PROPOSAL_APPLICANT_NAME_CASCADE.find((key) => proposal.values[key]);
+
+  if (bestNameKey) {
+    return proposal.values[bestNameKey];
+  }
+
+  return [PROPOSAL_APPLICANT_NAME_FALLBACK];
+};
+
+export {
+  PROPOSAL_APPLICANT_NAME_CASCADE,
+  PROPOSAL_APPLICANT_NAME_FALLBACK,
+  getPreferredApplicantNameValues,
+};
--- a/src/components/ProposalListGridItem.css
+++ b/src/components/ProposalListGridItem.css
@@ -17,11 +17,11 @@
   border-bottom: none;
 }
 
-.proposal--organization-name {
+.proposal--applicant-name {
   font-weight: var(--font-weight--medium);
 }
 
-.proposal--organization-address,
+.proposal--applicant-address,
 .proposal--proposal-name {
   color: var(--color--gray--medium-dark);
   font-size: 0.9em;
--- a/src/components/ProposalListGridItem.tsx
+++ b/src/components/ProposalListGridItem.tsx
@@ -9,9 +9,9 @@ interface ProposalListGridItemProps {
 }
 
 export const ProposalListGridItem = ({ proposal }: ProposalListGridItemProps) => {
-  const organizationName = getPreferredApplicantNameValues(proposal);
+  const applicantName = getPreferredApplicantNameValues(proposal);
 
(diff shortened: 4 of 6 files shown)


src/components/ProposalListTable.tsx
```diff
@@ -25,12 +26,26 @@ const ProposalListTableRow = ({
     navigate(`/proposals/${proposal.id}`);
   };
 
+  const getProposalCellContents = (shortCode: string) => {
+    let proposalValue;
+
+    switch (shortCode) {
```
jasonaowen, line 32: > Note that `getProposalCellContents()` is implemented with a switch statement because I predict we’ll do something similar for other columns very soon. We may not stick with this syntax but it’s a good bet.

What other columns are you thinking about?

src/utils/proposals.ts
```diff
@@ -0,0 +1,40 @@
...
+   * Proposals may store the applicant name in one of several fields.
+   * This utility returns iterates through the list of attributes where
+   * we expect to find a name, in a preferential order, and returns the
+   * first matching value.
+   *
+   * If none is found, falls back to a fallback value (if configured) or null.
+   *
+   * @param  {DataViewerProposal} proposal
+   * @param  {Boolean} fallback Whether to fall back to the applicant name fallback
+   *                            if no better name is found. Defaults to true.
```
jasonaowen, line 23: Under what circumstances would we set this to `false`?

jasonaowen (commented):
Thanks so much for working on this, @reefdog!

I have a few questions, then I'll test this later this afternoon.

src/components/ProposalListTable.tsx
reefdog: @jasonowens I'm specifically thinking this may be the entry point for doing formatting or linkifying URLs.

src/utils/proposals.ts
reefdog: 😆 An excellent question. I don't have a current need. The circumstances I imagined were checking "does this proposal have an applicant name?" (separate from "give me an applicant name, even if it's the fallback string"). But if you'd prefer I remove it until we need it, happy to.

jasonaowen: Yes, please! :upside_down_face:

src/components/ProposalListTable.tsx
jasonaowen: Aha! I had been thinking of introducing a separate component - `<FormattedValue>` or so - but this is interesting. Let me think about it!

src/utils/proposals.ts
reefdog: Done!

src/components/ProposalListTable.tsx
jasonaowen: I think for formatting values we'll want to make a decision on a not-yet-passed attribute of a base field, rather than on the short code.

What I thought you might be getting at is that there are other sets of columns that should be collapsed like this, in which case I might advocate for a more general approach, but we brainstormed for a bit the other day as we were discussing variants and couldn't come up with any other such sets.

So yes, would you mind switching (ha) to an `if` statement, please? Thank you! :blue_heart:

reefdog: > I think for formatting values we'll want to make a decision on a not-yet-passed attribute of a base field, rather than on the short code.

Ah yes, good point.

> What I thought you might be getting at is that there are other sets of columns that should be collapsed like this, in which case I might advocate for a more general approach, but we brainstormed for a bit the other day as we were discussing variants and couldn't come up with any other such sets.

We do have one more (which I should have mentioned above): Proposal Name. The table is simplistically only looking at `proposal_name`, but [on the detail page we cascade to `proposal_title`](https://github.com/PhilanthropyDataCommons/data-viewer/blob/475c87d5c0ef94ee19f2628d013ea885d299a6f3/src/pages/ProposalDetail.tsx#L122-L123). In fact… we should fix that both here on the table and in the [sidebar grid](https://github.com/PhilanthropyDataCommons/data-viewer/blob/475c87d5c0ef94ee19f2628d013ea885d299a6f3/src/components/ProposalListGridItem.tsx#L38).

…which is to say, can we keep this a `switch` after all and I either fix the proposal title as well, or add that as a separate followup task?

jasonaowen: Excellent point! Okay yes, let's keep the `switch`. Thanks for talking and thinking through this with me!

Once we have a resolution to https://github.com/PhilanthropyDataCommons/service/issues/284 we can follow up; let's not block this PR on that question.

reefdog: @jasonaowen Okay based on our conversation in Zulip just now, it seems like there is an unresolved question as to whether `project_name` and `proposal_project_title` are conceptually related enough that we should do the "org name"-style cascading among them. We're going to discuss it in tomorrow's meeting.

If they aren't related enough, we'll need to remove the Proposal Title cascade from the detail page. If they are related enough, we'll want to adopt the cascade on the two missing places (table and sidebar).

Since in either case that's out of scope of this PR, I'm going to reduce it to an `if` and we can switch (lol you're right that is funny) to a `switch` if we decide that path.

jasonaowen (approved):
This looks good to me! Thanks, @reefdog.

I was able to test this locally, and it looks great! Thank you, as always, for the carefully structured commits - makes reviewing a breeze. :blue_heart:

src/components/ProposalListTable.tsx
reefdog: @jasonaowen Ah I didn't see your reply as I was drafting mine… okay yes, it's easiest just to leave it in (and I sort of suspect we're going to have multiple proposal name field candidates). Will keep as-is, and thank _you_!

src/utils/proposals.ts
```diff
@@ -0,0 +1,38 @@
...
+  'organization_legal_name',
+  'proposal_primary_contact_name',
+  'proposal_submitter_name',
+];
+
+const PROPOSAL_APPLICANT_NAME_FALLBACK = 'Unknown Applicant';
+
+/**
+   * Proposals may store the applicant name in one of several fields.
+   * This utility returns iterates through the list of attributes where
```
reefdog, line 15: Bah and of course I would notice this just seconds after clicking merge. 🤦

Issue #71, opened 2023-04-07
Use PDC logo for favicon


reefdog:
We've been marketing React too long, it's time for our own brand in the tab!

It should be our little shield logo, naturally.

- https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs
- https://web.dev/building-an-adaptive-favicon/

jasonaowen:
Is this waiting on #26? Or, who can provide this asset?

reefdog:
@jasonaowen Me! I'll do this.

reefdog:
@jasonaowen (Our existing PNGs are plenty high-fidelity for me to generate favicons from.)

PR #116, opened 2023-05-15
Cleanup some React App boilerplate

reefdog:
While preparing to tackle #71, I discovered a few Create React App boilerplate bits and pieces that deserved a quick cleanup but that I didn't want to muss up that impending PR with. So here they are.

This primarily removes some HTML comments that we don't need to be shipping to users (although I'm miffed we don't have a good place in our README for the `%PUBLIC_URL%` explanation, though it's easily Googleable) and corrects names/description metadata.

I considered just removing the WebManifest file because we have no intention of maintaining this as a mobile web PWA, but I can't convince myself of that.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -10,34 +10,11 @@
       content="Web site created using create-react-app"
     />
     <link rel="apple-touch-icon" href="%PUBLIC_URL%/logo192.png" />
-    <!--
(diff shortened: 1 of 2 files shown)


jasonaowen (approved):
Looks good to me! Thanks for cleaning all this up, @reefdog.

PR #121, opened 2023-05-16
Don't cascade `proposal_name` to other fields

### LINKED ISSUE #120: Don't cascade `proposal_name` field to other fields
We pluck out the Proposal Name to display atop the Proposal Detail page and in the proposal lists (table, sidebar). Historically we had both a `proposal_name` and a `proposal_title`, and without clarity about the relationship between them, we [cascade between them](https://github.com/PhilanthropyDataCommons/data-viewer/blob/767ec2085ec8d7be10670e7130a85d9353203d97/src/pages/ProposalDetail.tsx#L128-L129). `proposal_title` has [since been renamed](https://github.com/PhilanthropyDataCommons/service/issues/284#issuecomment-1531799718) `proposal_project_title`.

After further discussion among the dev team just now, spawned by [this conversation](https://github.com/PhilanthropyDataCommons/service/issues/284), we determined `proposal_name` is the correct field to look at and we no longer need cascading.

reefdog:
This PR simplifies our display of `proposal_name` from a cascade to other fields to just a simple field display.

Because of this change, I went ahead and made the simplification @jasonaowen [suggested](https://github.com/PhilanthropyDataCommons/data-viewer/pull/114#discussion_r1192764099) in his review of #114.

**Testing:**

The cascade that this PR removes wasn't working as designed anyway, since the field that `proposal_name` was falling back to (`proposal_title`) [had already been renamed](https://github.com/PhilanthropyDataCommons/service/issues/284#issuecomment-1531799718) without the viewer knowing / being updated. (Oops.) Thus, there's no A/B (`main` / this branch) comparison you can make to prove anything.

So: just load up the proposal detail page for any proposal with a `proposal_name` (e.g., 424 from the production API) and make sure the proposal name is displayed in the panel header as shown below:

image

Closes #120

### DIFF
--- a/src/components/ProposalListTable.tsx
+++ b/src/components/ProposalListTable.tsx
@@ -26,19 +26,9 @@ const ProposalListTableRow = ({
     navigate(`/proposals/${proposal.id}`);
   };
 
-  const getProposalCellContents = (shortCode: string) => {
-    let proposalValue;
-
-    switch (shortCode) {
-      case 'organization_name':
-        proposalValue = getPreferredApplicantNameValues(proposal);
-        break;
-      default:
-        proposalValue = proposal.values[shortCode];
-    }
-
-    return proposalValue;
-  };
+  const getProposalCellContents = (shortCode: string) => (shortCode === 'organization_name'
+    ? getPreferredApplicantNameValues(proposal)
+    : proposal.values[shortCode]);
 
   return (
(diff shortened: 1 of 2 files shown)


jasonaowen (approved):
Looks good - thanks for the quick turnaround on this, @reefdog!

Issue #23, opened 2023-02-28
Load proposal's related org details in side panel


reefdog:
Users should be able to view a detail panel for related organization next to the proposal:

<img width="1512" alt="PDC Data Viewer 06 - Proposal with Applicant" src="https://user-images.githubusercontent.com/4731/220751461-105dd5bd-4f1d-45a9-be7f-6514d6452dba.png">

Note that because org detail pages have lists of proposals, which are opened in a side panel on _those_ detail pages, we'll need to figure out the right UX flow.

reefdog:
Heyyyy, I forgot about this old issue! This is all done thanks to #616. 🎉

reefdog:
Well, actually… this is more about opening the org detail page in a side panel from the proposal page, which means:

1. Linking proposal to org
2. Loading org detail in side panel
3. Being able to "expand/maximize" org detail

So that is new! I'll reopen and reframe the issue.

slifty:
Per #988 we aren't going to be supporting this functionality.

PR #166, opened 2023-05-26
Add Dropdown component and story

reefdog:
This PR adds a `Dropdown` component, used for generating dropdown menus from button-like triggers:

https://github.com/PhilanthropyDataCommons/data-viewer/assets/4731/2ab5c95a-0d36-4fbf-afe3-55897c2e5671

This implementation is specific to our current needs and doesn’t offer the flexibility one would want from a general-purpose dropdown library. That’s fine until we have more use cases.

**Testing:**

As this isn't yet implemented on any page, you'll need to run the Storybook to see it.

Issue #23 Create applicant detail page

### DIFF
--- a/src/components/Dropdown/Dropdown.css
+++ b/src/components/Dropdown/Dropdown.css
@@ -1,0 +1,115 @@
+:root {
+  --dropdown--menu--border-radius: 8px;
+  --dropdown--menu--spacing: var(--accessible-spacing--1x);
+  --dropdown--menu-item--icon-size: 1.5em;
+}
+
+.dropdown {
+  position: relative;
+  display: inline-block;
+}
+
+.dropdown-trigger {
+  /* Hide the disclosure triangle that is natively applied to `<summary>` elements.
+     This is already implicit because the dropdown-trigger has the `.button` class applied,
+     which class specifies the `all: unset` property that has this same effect.
+     However, we shouldn't rely on that implicit effect,
+     so here we make our intention explicit. */
+  list-style: none;
+}
+
+/* Safari hack: Safari doesn't respect the above method of hiding the disclosure triangle. */
+.dropdown-trigger::-webkit-details-marker {
+  display: none;
+}
+
+.dropdown[open] > .dropdown-trigger {
+  --button--primary-color: var(--color--blue);
+  --button--contrast-color: var(--color--gray--lighter);
+}
+
+/* This generates an invisible backdrop when the dropdown is open,
+   behind the dropdown but above all other page content,
+   that is technically part of the dropdown trigger.
+   Thus, clicking anywhere else on the page is like clicking the trigger,
+   and closes the dropdown. */
+.dropdown[open] > .dropdown-trigger::before {
+  position: fixed;
+  inset: 0;
+  z-index: 80;
+  content: ' ';
+  display: block;
+  cursor: default;
+}
+
+.dropdown-menu {
+  position: absolute;
+  z-index: 100;
+  width: 30ch;
+
+  background-color: white;
+  border-radius: 8px;
+  border: 1px solid var(--color--gray--medium);
+  overflow: hidden;
+
+  display: flex;
+  flex-direction: column;
+}
+
+.dropdown.dropdown--left > .dropdown-menu {
+  left: 0;
+}
+
+.dropdown.dropdown--right > .dropdown-menu {
+  right: 0;
+}
+
+.dropdown-menu > * {
+  border-bottom: 1px solid var(--color--gray--light);
+}
+
+.dropdown-menu > *:last-child {
+  border-bottom: none;
+}
+
+/* CSS `transition` doesn't work when disclosing `<details>` element contents,
+  thus we use `animation`. */
+@keyframes show-dropdown-menu {
+  from {
+    opacity: 0;
+    transform: translateY(0);
+  }
+}
+
+.dropdown[open] .dropdown-menu {
+  animation: show-dropdown-menu 200ms ease-in-out;
+
+  opacity: 1;
+  transform: translateY(10px);
+}
+
+.dropdown-menu-text {
+  font-size: 0.85rem;
+  padding: var(--dropdown--menu--spacing);
+}
+
+.dropdown-menu-link {
+  color: inherit;
+  font-weight: var(--font-weight--medium);
+  text-decoration: none;
+  padding: var(--dropdown--menu--spacing);
+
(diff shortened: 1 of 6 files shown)


src/components/Dropdown/Dropdown.css
```diff
@@ -0,0 +1,115 @@
...
+  --button--primary-color: var(--color--blue);
+  --button--contrast-color: var(--color--gray--lighter);
+}
+
+/* This generates an invisible backgrop when the dropdown is open,
+   behind the dropdown but above all other page content,
+   that is technically part of the dropdown trigger.
+   Thus, clicking anywhere else on the page is like clicking the trigger,
+   and closes the dropdown. */
+.dropdown[open] > .dropdown-trigger::before {
```
jasonaowen, line 36: Woah, fancy - thanks for the explanation!

src/components/Dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,30 @@
...
+  children: React.ReactNode;
+  /**
+   * Should the menu be aligned to the left or right edge of its toggle?
+   */
+  align?: 'left' | 'right';
+  className?: string;
+}
+
+/**
+ * Accessible, JavaScript-free dropdowns using the `<details>` element.
```
jasonaowen, line 20: :tada: but also this is rendered with JavaScript so :upside_down_face: . I'm still a fan, just enjoying the irony!

src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -0,0 +1,19 @@
...
+interface DropdownTriggerProps {
+  children: React.ReactNode;
+  className?: string;
+}
+
+/**
+ * Generates a `<summary>` element, styled as a button,
+ * to use within a `Dropdown` component,
+ */
+export const DropdownTrigger = ({
```
jasonaowen, line 12: As you noted elsewhere, this puts the onus on the consuming code to ensure the first child of a `<Dropdown>` is this `<DropdownTrigger>`, and is difficult to enforce - or at least, I'm not sure how to do so right now.

The stories all use basic text as the only children of a `<DropdownTrigger>`; do you intend to use anything else in production? How would you feel about inlining `<DropdownTrigger>` into `<Dropdown>` by adding a `title` prop (or `summary`, or `name`, or `trigger`, or...)?

src/stories/Dropdown.stories.tsx
```diff
@@ -0,0 +1,150 @@
...
+  DropdownMenuTitle,
+  DropdownMenuLink,
+} from '../components/Dropdown';
+
+const meta = {
+  component: Dropdown,
+  tags: ['autodocs'],
+  // Storybook elements are placed inside `overflow:hidden` containers,
+  // which in our case hide the dropdown menu. This lets us see them,
+  // at the expense of space.
```
jasonaowen, line 18: Did you mean to put this comment next to the `decorator`? Or what is the "this" that lets us seem them?

jasonaowen (commented):
Very cool! I have questions, though.

src/components/Dropdown/Dropdown.tsx
reefdog: 😝

src/components/Dropdown/DropdownTrigger.tsx
reefdog: Oof, I had forgotten the requirement that the first element of `<details>` must be `<summary>`, which means yes there is an unstated order requirement here. :/

So I do intend for `DropdownTrigger` to allow multiple children, e.g. an icon + text. Okay actually that's my only example so far. But I think your solution is still a good one. In fact, it may let me condense the hierarchy to simply:

```jsx
<Dropdown trigger={}>
  /* Dropdown items here */
</Dropdown>
```

Then there's no need for `DropdownTrigger` *or* `DropdownMenu` components. I kind of like this! What do you think?

src/stories/Dropdown.stories.tsx
reefdog: Oh, yes, thank you! This was leftover from when I originally defined `decorators` here. But then I realized each example needed its own `height`, and so had to define them individually. (I didn't try very hard to have one default decorator, over-rideable by individual examples; it seemed to just nest them, which didn't work.)

Let me take one more look at if I can override a default decorator, and if not, I'll move this comment to the `Default` example.

src/components/Dropdown/DropdownTrigger.tsx
jasonaowen: That looks great! Passing a component as a prop is a pattern I've seen before, particularly with react-router, so that would have been my next suggestion.

src/stories/Dropdown.stories.tsx
reefdog: I suspect there is a solution to what I want (default component decorator that individual stories can override) either through the use of custom story arguments or some sort of "story decorator overrides component decorator" feature. But I'm not finding it quickly enough and that's really just an optimization so I'm going to go with the latter solution above.

src/components/Dropdown/DropdownTrigger.tsx
reefdog: @jasonaowen Okay, updated! Ready for you check out again.

src/components/Dropdown/DropdownMenu.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+interface DropdownMenuProps {
+  children: React.ReactNode;
+  className?: string;
+}
+
+/**
+ * A menu that is only visible when its `Dropdown` parent has an `open` attribute.
+ */
+export const DropdownMenu = ({
```
jasonaowen, line 11: I think this could be inlined, too, since it's only used in a single place!

src/stories/Dropdown.stories.tsx
```diff
@@ -0,0 +1,128 @@
...
+    ),
+  ],
+};
+
+export const WithIconInTrigger: Story = {
+  args: {
+    trigger: [
+      <ArrowUpOnSquareIcon />,
+      'Toggle menu',
+    ],
```
jasonaowen, line 67: Ooh, love this!

jasonaowen (approved):
Awesome work, @reefdog!

src/components/Dropdown/DropdownMenu.tsx
reefdog: Good idea! Done.

PR #167, opened 2023-05-26
Condense next/previous proposal buttons

reefdog:
For speedy dev navigation between proposals, we offer next/previous buttons on the proposal detail panel. These take up quite a lot of space, however.

This change condenses them by hiding the labels and using the `FormElementGroup` wrapper.

**Before:**
Proposal navigation with labeled buttons

**After:**
Proposal navigation with icons only

### DIFF
--- a/src/components/ProposalDetailPanel.tsx
+++ b/src/components/ProposalDetailPanel.tsx
@@ -12,6 +12,7 @@ import {
   PanelTitleWrapper,
 } from './Panel';
 import { ProposalTable } from './ProposalTable';
+import { FormElementGroup } from './FormElementGroup';
 
 interface ProposalDetailPanelProps {
   title: string | undefined;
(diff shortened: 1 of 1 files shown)


reefdog:
@jasonaowen I did this because once we add the dropdown menu, particularly, it got noisy in that corner.

jasonaowen (approved):
LGTM! Tested locally and love the rounded outer corners / sharp inner corners.

I'd also be down with just deleting these buttons now that we have the list sidebar; they were a dev aide that is less relevant now. But if you're still getting value out of them, let's keep them!

reefdog:
@jasonaowen I'm actually *not* getting value out of them, but when I started to delete them there were some trickle-down (up?) effects, such as no longer needing the `proposalId` prop passed to the page, that I quickly decided "hey lemme just do what I need for the moment". I'll open a separate issue for removing them just in case it dominoes.

reefdog:
> love the rounded outer corners / sharp inner corners.

Yes! This is an artifact of the [`FormElementGroup`](https://pilot.philanthropydatacommons.org/storybook/?path=/docs/stories-formelementgroup--docs) component.

PR #169, opened 2023-05-26
Allow table row headers to wrap but with min-width

### LINKED ISSUE #168: Long field labels break table formatting
On proposal detail pages, when field labels are very long, the table looks terrible because we currently aren't letting the row header content wrap:

table-wide

This was an intentional choice because field **values** are usually quite long, and without that property, often took up all available room and made field labels look bad:

table-with-wrap

Obviously, we need a middle ground: to not let the values take up _all_ available space, but while still letting row headers wrap in case they are really really long.

reefdog:
This PR replaces the draconian `white-space: nowrap` on row headers with a more sensible min-width of 25%.

While this may still not look perfect in all scenarios, it looks natural in all scenarios I tested it against, and avoids the problem where very long row headers can generate undesired layout.

**Before:**
before

**After:**
after

Closes #168 Long field labels break table formatting

### DIFF
--- a/src/components/Table/Table.css
+++ b/src/components/Table/Table.css
@@ -23,7 +23,7 @@
 }
 
 .table tbody th {
-  white-space: nowrap;
+  min-width: 25%;
 }
 
 .table th,


jasonaowen (approved):
Nice and simple!

I tested this locally, and it looks good to me on the proposals I examined.

Thanks, @reefdog!

PR #171, opened 2023-05-26
Remove Storybook examples

### LINKED ISSUE #170: Remove Storybook examples
When initializing Storybook, it creates some example components, pages, and stories so you can see how it all works.

These have outlived their usefulness and now just get in the way when trying to open or import our similarly-named components. We should remove them.

reefdog:
When initializing Storybook, it creates some example components, pages, and stories so you can see how it all works.

These have outlived their usefulness and now just get in the way when trying to open or import our similarly-named components.

This PR deletes all the examples, and renames the path of the Storybook introduction document so it’s not nested in an “Example” section in the sidebar.

Closes #170 Remove Storybook examples


jasonaowen (approved):
Love it!

PR #176, opened 2023-05-27
Add data platform panel

reefdog:
_This commit ladders off #166._

This PR adds the components and page required to show data platform provider data:

providers

Some of these commits can be used as-is (pending review), but others are intended only to demonstrate patterns and usage. I expect @jasonaowen will want to take a slightly different tactic for generating the `ProposalDetailWithProvider` page, for instance.

**Testing:**
1. Go to any proposal detail page
2. Use the new "Data providers" dropdown to open a provider panel
3. Data is placeholder, but the close button should work and the link to Candid and Charity Navigator _might_ work. (The URL format is correct, but it's possible the platform doesn't have that EIN.)

**Notes:**
- This is quite messy but a start.
- Doesn't do anything conditional based on whether we have data from one of the platforms; we should add this.
- Hardcodes in the known provider names, and/but doesn't sanitize the URL to only allow known providers.
- In the course of this, "data platform providers" became a dizzyingly difficult name to understand. Are these platforms? Or providers? What is a data platform provider, anyway?
- Doesn't attempt the challenge of mapping keys to labels; assumes it is passed an array of value objects similar to the proposal table.

Issue #23

PR #191, opened 2023-05-30
Maintain consistent internal button positioning

reefdog:
This PR improves button display in two ways:

- Don’t allow the icon to shrink; it’s small, there’s no need
- Don’t allow button text to wrap; if we find ourself in this scenario, something else should probably change. (E.g., responsively hiding the text portion of the button, reducing wording, etc.)

**Before,** this was (incorrectly) possible:
Screenshot 2023-05-30 at 5 26 33 PM

**Now,** the interface insists upon maintaining this:
Screenshot 2023-05-30 at 5 26 44 PM

Why? Because too often buttons were in contexts (such as a flex layout) that _suggested_ they should reduce in size, when we didn't really want them to. Instead, we tend to want _other_ elements to get out of the way before we modify button layout. If we find some part of the interface where our buttons really are too big, we should address it some other way.

### DIFF
--- a/src/components/Button.css
+++ b/src/components/Button.css
@@ -25,6 +25,7 @@
   font-size: var(--button--font-size);
   line-height: 1;
   letter-spacing: normal;
+  white-space: nowrap;
   height: var(--button--height);
   padding: var(--button--padding-y) var(--button--padding-x);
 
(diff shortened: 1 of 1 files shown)


jasonaowen (approved):
These changes make sense!

I'm not sure how to test it; was this happening before on `main`? Or is this related to #176?

reefdog:
Yes, it was happening on #176 fairly often, though I thought I’d seen it before.

PR #193, opened 2023-05-31
Close dropdown when a child link is clicked

### LINKED ISSUE #190: Close dropdown when item is used
[Currently](https://pilot.philanthropydatacommons.org/storybook/?path=/docs/stories-dropdown--docs), when you click on a dropdown menu item, the dropdown itself doesn't respond. It was built to expect that the dropdown item clicked would be a link, resulting in a full-page navigation, which would naturally "reset" the dropdown menu. That was not a thorough expectation! A link may not necessarily lead to the menu being removed.

All we currently need:

- Clicking a dropdown menu item should close the dropdown (after / in addition to presumably performing its own action)

A more thorough design, which we won't implement yet because we have no need, but I want to document, would include:

- Configuring the menu to disable the close-on-click behavior (imagine a menu of radio-button-like items, where you want to continue viewing the menu / confirming the action was taken before closing)
- A non-link `DropdownMenuItem`, to support more form-like (checkbox, radio button) menu items that aren't actually links

reefdog:
This PR ensures a dropdown menu closes when a link within is clicked.

@jasonaowen rubber-ducked for a while on where this should live before settling on binding directly to the `Link` within each `DropdownMenuLink` component, and targeting the parent `Dropdown` component using native DOM traversal rather than refs or passing a handler in as a prop. This isn't the most React-idiomatic approach, but it avoids a larger set of complications.

It also fixes a missing key issue in the Dropdown Storybook that I hadn't noticed before.

**Testing:**
1. Go to [`main`'s Dropdown Storybook](https://pilot.philanthropydatacommons.org/storybook/?path=/docs/stories-dropdown--docs)
2. Open a dropdown and click an item inside
3. ❌ Note that the dropdown menu doesn't close!
4. Go to [this branch's Dropdown Storybook](https://deploy-preview-193--philanthropy-data-commons-viewer.netlify.app/storybook/?path=/docs/stories-dropdown--docs)
6. Open a dropdown and click an item inside
7. ✅ Note that the dropdown menu closes!

Closes #190

### DIFF
--- a/src/components/Dropdown/DropdownMenuLink.tsx
+++ b/src/components/Dropdown/DropdownMenuLink.tsx
@@ -17,8 +17,19 @@ export const DropdownMenuLink = ({
   to,
   children,
   className = '',
-}: DropdownMenuLinkProps) => (
-  <Link to={to} className={`dropdown-menu-link ${className}`.trim()}>
-    {children}
-  </Link>
-);
+}: DropdownMenuLinkProps) => {
+  const handleClick = (e: React.SyntheticEvent) => {
+    const $target = e.nativeEvent.target as HTMLLinkElement;
+    $target.closest('.dropdown')?.removeAttribute('open');
+  };
+
+  return (
+    <Link
+      to={to}
+      className={`dropdown-menu-link ${className}`.trim()}
+      onClick={handleClick}
+    >
+      {children}
+    </Link>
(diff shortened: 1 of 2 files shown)


jasonaowen (approved):
Thanks for talking through this with me, @reefdog!

PR #196, opened 2023-06-01
Add padding to panel bodies by default

### LINKED ISSUE #195: Panel bodies have no internal padding
By design, `PanelBody` component has no internal padding. That's because we _usually_ (but not always) use it with content such as tables, which we want to occupy the entire space. Really, this should be configurable, and in fact the default behavior should be to include padding.

reefdog:
This PR reformats our `PanelBody` components so that, by default, they have an internal padding that matches the `PanelHeader` padding. This makes them more usable as-is without any further styling. The padding can be removed with the new `padded` prop, which removes the padding for a more "full-panel" interface that we often use.

This required adding the `padded` prop to multiple components and changing the `ClosablePanel` component to pass it through as well.

**Testing:**
- Most panels should look the same as they did before
- The "Loading…" and "No data found" versions of the data platform provider panels should now appear with padding, so the messages are no longer flush with the edge

**Regression note:**

There is one big regression that I don't like and can't figure out: when the Dashboard panel re-renders from "Loading" to have content, we also flip from padded to not padded. I would expect this change to be instantaneous; that is, as soon as the table appears, it is already in an unpadded `PanelBody`. But it seems to render into the padded `PanelBody` and _then_ very quickly the padding is removed, causing an unsightly reformatting flash as the table repositions. I don't like this but couldn't solve it, and welcome suggestions.

Closes #195

### DIFF
--- a/src/components/Panel/Panel.css
+++ b/src/components/Panel/Panel.css
@@ -86,7 +86,3 @@
 .panel-body--padded {
   padding: var(--panel--padding);
 }
-
-.panel-message {
-  color: var(--color--gray--medium-dark);
-}
--- a/src/components/Panel/Panel.css
+++ b/src/components/Panel/Panel.css
@@ -83,9 +83,3 @@
   padding: var(--fixed-spacing--1x);
   color: var(--color--gray--medium-dark);
 }
-
-.panel.fill .panel-body {
-  position: absolute;
-  inset: 0;
-  overflow: auto;
-}
--- a/src/components/Panel/Panel.css
+++ b/src/components/Panel/Panel.css
@@ -83,7 +83,10 @@
   overflow: auto;
 }
 
+.panel-body--padded {
+  padding: var(--panel--padding);
+}
+
 .panel-message {
   color: var(--color--gray--medium-dark);
-  padding: var(--panel--padding);
 }
--- a/src/components/Panel/Panel.css
+++ b/src/components/Panel/Panel.css
@@ -1,3 +1,7 @@
+:root {
+  --panel--padding: var(--fixed-spacing--1x);
+}
+
 .panel {
   background-color: white;
   border-radius: 5px;
@@ -10,7 +14,7 @@
 
 .panel-header {
   grid-row: 1;
-  padding: var(--fixed-spacing--1x);
+  padding: var(--panel--padding);
 
   display: flex;
   justify-content: space-between;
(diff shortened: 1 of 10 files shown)


src/components/Panel/Panel.css
```diff
@@ -1,3 +1,7 @@
+:root {
+  --panel--padding: var(--fixed-spacing--1x);
```
jasonaowen, line 2: Commit 797259655cbdc53608ff896e9eb884fb685a8403 has a typo in its commit message: `weay`. (Picking a line introduced with that commit to comment on, but not related to this line specifically.)

jasonaowen (commented):
> There is one big regression that I don't like and can't figure out: when the Dashboard panel re-renders from "Loading" to have content, we also flip from padded to not padded. I would expect this change to be instantaneous; that is, as soon as the table appears, it is already in an unpadded `PanelBody`. But it seems to render into the padded `PanelBody` and _then_ very quickly the padding is removed, causing an unsightly reformatting flash as the table repositions. I don't like this but couldn't solve it, and welcome suggestions.

I believe this comes from the `transition-duration` property: https://github.com/PhilanthropyDataCommons/data-viewer/blob/22d2158ef59deb5b4cb1015bb3be126b67d275f7/src/index.css#L59-L64

Combined with the fact that that loading message previously was getting its padding via `.panel-message` on its enclosing `<div>`, rather than via the `padded` prop on the parent `<PanelBody>`.

I'm not sure what the best fix is here; maybe suppress the transition on `<PanelBody>`? Or maybe those alternate states should be pulled out into separate local components, like for data platform providers?

jasonaowen (approved):
Aside from the regression, :+1: to the rest! Thanks for cleaning this up, @reefdog.

I'll defer to you on whether the regression is more or less important than the fixes, and in particular what you'd rather demo.

reefdog:
> I believe this comes from the transition-duration property: 

@jasonaowen 😮 You're right! So, I _thought_ that was a clever way of setting a default `transition-duration` without also declaring what properties should transition. (This way, individual elements could still declare `transition-property` to opt into a particular property transitioning, with the default `transition-duration` automatically applying because of that global rule.) But it looks like setting `transition-duration` implicitly sets `transition-property: all`, so transitions now apply to all property changes! Booooo!

Excellent find, I'll fix ASAP.

PR #200, opened 2023-06-02
Highlight current proposal in scroll list panel

### LINKED ISSUE #174: Highlight current proposal in sidebar list
Currently, we do not indicate which proposal we are viewing in the proposal list sidebar of the proposal detail page. This makes it more difficult to browse around, as it's easier to lose your place.

reefdog:
This PR styles the current proposal in the sidebar list with a light blue highlight (and slightly darker text, for contrast). It also ensures the current proposal is scrolled into view when the panel first renders. It doesn't scroll the view every time the current proposal changes, because that feels chaotic. It also actually scrolls to the element just previous to the active one, as that signals "there are still more proposals above this one".

Composing a screenshot is burdensome because of having to redact the details, so best to just run it.

**Testing:**
1. Run branch and go to a proposal
2. The sidebar should render with that proposal highlighted in light blue and scrolled with the current proposal nearly to the top (unless the proposal is too near the bottom of the list for the sidebar to scroll up any farther)
3. As you click to other proposals, the highlight should change.
4. If you refresh the page from any of those proposals, the selected proposal should be positioned nearly at the top of the sidebar

Closes #174


src/components/ProposalListGrid.tsx
```diff
@@ -1,18 +1,35 @@
...
-    {proposals.map((proposal, index) => (
-      <ProposalListGridItem
-        key={index} // eslint-disable-line react/no-array-index-key
-        proposal={proposal}
-      />
-    ))}
-  </div>
-);
+export const ProposalListGrid = ({ proposals }: ProposalListGridProps) => {
+  const { proposalId } = useParams();
```
reefdog, line 11: @jasonaowen Is it okay for a component to reach into params like this? It made sense at the time so I wasn't passing the prop down through all the components from the page, but upon waking up this morning I realized maybe only pages should access params? I also realized I'm out of the loop on what Redux-like global stores React might have these days to get around prop-batons. Anyway, your thoughts here appreciated!

jasonaowen: Thanks for calling this out! I'd tend to prefer that the current proposal ID be passed in as a prop, primarily because the `<ProposalListGrid>` shouldn't know or care how it's rendered; the fact that this is a route parameter should only be the responsibility of the component responsible for that route. Plus, it'd let us show off the highlighted proposal row in storybook!

src/components/ProposalListGrid.tsx
```diff
@@ -10,8 +10,19 @@ interface ProposalListGridProps {
 export const ProposalListGrid = ({ proposals }: ProposalListGridProps) => {
   const { proposalId } = useParams();
 
+  const listGridRef = React.useRef<HTMLDivElement>(null);
+
+  useEffect(() => {
+    // Scroll to the previous sibling to emphasize that there are (probably)
+    // previous elements in the list out of scroll-view.
+    listGridRef.current?.querySelector('.active')?.previousElementSibling?.scrollIntoView();
+  }, []);
```
jasonaowen, line 19: As written, this is a one-shot. I think that's probably the right call, but I wanted to check: should this run again if the currently selected proposal changes? Probably not, because clicking on a proposal shouldn't scroll the list out from under the mouse.

jasonaowen (changes requested):
This looks amazing! Excellent work, @reefdog.

src/components/ProposalListGrid.tsx
reefdog: That's correct! This scroll ensures the current proposal is visible in the sidebar _when the page renders_ to give the user a sense of geography. After that the user should feel entirely in control. Clicking a new proposal doesn't need to re-scroll the sidebar, because the user's sense of geography isn't at risk.

src/components/ProposalListGrid.tsx
reefdog: Two great reasons, I'll add the prop-baton!

src/components/ProposalListGrid.tsx
jasonaowen: Well stated! Thanks for talking through that with me.

src/components/ProposalListGrid.tsx
reefdog: Done and ready for re-review!

I also added a story as you made reference to in your review, which necessitated one new prerequisite commit so that sample proposals data could be shared between the stories.

jasonaowen (approved):
Awesome! Thanks so much, @reefdog.

PR #221, opened 2023-06-15
Remove next/prev proposal buttons

### LINKED ISSUE #172: Remove next/previous proposal dev buttons
We devs gave ourselves next/previous buttons on a proposal page that let us quickly navigate among sequential proposal IDs to see different examples of data. This was never shown to end users and had expected edge bugs such as breaking when there are gaps in the ID sequence (or at the boundaries).

Now that we have the sidebar, which is the "correct" way to navigate to other proposals (since, e.g., we know they exist), we can remove these buttons.

https://github.com/PhilanthropyDataCommons/data-viewer/blob/9697f5dd9bfed939abf3df4a9e3de13f48d04b96/src/components/ProposalDetailPanel.tsx#L59-L76

reefdog:
_This PR ladders off #200; otherwise, removing this block of code would cause the linter to want me to remove the `proposalId` reference, but #200 uses that reference. It would cause unnecessary churn to have started this off `main`._

This PR removes the dev-only next/previous arrows in the top-right of the proposal detail panel. Now that we have a sidebar, we no longer need this convenience.

**Testing:**

Because this is a dev feature, testing must be done when running the app in dev mode (like, locally). The deploy preview isn't sufficient.

1. On `main`, go to a proposal detail page.
2. Note the left/right arrows in the top right of the detail panel.
3. Switch to this branch and check again.
4. ✅ Note the absence of the left/right arrows.

Closes #172

### DIFF
--- a/src/components/ProposalDetailPanel.tsx
+++ b/src/components/ProposalDetailPanel.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
-import { Link } from 'react-router-dom';
-import { BackwardIcon, DocumentTextIcon, ForwardIcon } from '@heroicons/react/24/solid';
+import { DocumentTextIcon } from '@heroicons/react/24/solid';
 import { ArrowRightOnRectangleIcon, CircleStackIcon } from '@heroicons/react/24/outline';
 import {
   Panel,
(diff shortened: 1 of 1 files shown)


jasonaowen (approved):
"The time has come," the Walrus said,

PR #241, opened 2023-06-23
Upgrade Storybook, and tell Dependabot not to

jasonaowen:
Storybook has a bunch of components to choose from. Those components should be updated alongside each other; mixing different versions is not a good practice, and ends up causing a lot of PR thrash and email spam. Additionally, since Dependabot will only open a certain number PRs at a time, these PRs can block other, more important updates.
    
The team working on Dependabot is considering a feature to allow for grouped updates, where we'd be able to specify that all of the Storybook components should be updated at the same time, rather than separate PRs. Until that feature is implemented, though, tell Dependabot to ignore Storybook.

Use the [upgrade script](https://storybook.js.org/docs/react/configure/upgrading#page-top) to update all its dependencies to [7.0.23](https://github.com/storybookjs/storybook/releases/tag/v7.0.23):
    
```
npx storybook@latest upgrade
```
    
This brings us back to a consistent version, after some dependabot thrash where it didn't notice a new version until after merging some of its old version PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,6 +9,10 @@ updates:
       - dependency-name: "@types/node"
         update-types:
           - "version-update:semver-major"
+      - dependency-name: "*storybook*"
+        # Dependabot is not yet able to upgrade all the dependencies together,
+        # which leads to mixed versions and quite a lot of PR thrash.
+        # See https://github.com/dependabot/dependabot-core/issues/1190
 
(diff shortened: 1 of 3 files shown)


reefdog (approved):
Your reasoning for ignoring updates sounds very sensible to me. ✅

And the Storybook upgrade works just fine! ✅✅

jasonaowen:
Thanks, @reefdog!

PR #248, opened 2023-06-24
Add RECENT_CHANGES document and link to it

### LINKED ISSUE #39: Add changelog / What's New page
Jim has requested a page on which we can communicate to Pilot participants what's new / what's changed.

Eventually, this will probably have to be a bbbbbblog, but it will start as a static page whose contents have to be updated manually (and as HTML unless Markdown comes "free").

reefdog:
This PR adds the `RECENT_CHANGES.md` document we discussed in #39, populates it with some suggested content, and modifies the landing page to link to it. (Note that this link won't actually working until this PR is merged in, but I don't think that weirdness should stop us from treating this as a single PR.)

Please feel free to suggest edits/additions to the recent changes! I'm tagging everyone on this review because suggestions for populating the change list are welcome from anyone. Please do read [this comment](https://github.com/PhilanthropyDataCommons/data-viewer/issues/39#issuecomment-1603439095) first, which explains what sort of changes we want to record here. (It's not a technical CHANGELOG.) 

Closes #39

### DIFF
--- a/RECENT_CHANGES.md
+++ b/RECENT_CHANGES.md
@@ -1,0 +1,26 @@
+# Recent Changes to the PDC Data Viewer
+
+## June 2023
+
+- Add the ability to show applicant data from data platform providers,
+  starting with Candid and Charity Navigator
+- Highlight current proposal in sidebar
+- Allow text selection on Dashboard table
+
+## May 2023
+
+- Add search capability to Dashboard
+- Add sidebar to proposal detail page
+- Improve performance using batched API queries
+- Improve selection of applicant name across varying base fields
+- Improve rendering of Dashboard table across viewport sizes
+
+## April 2023
+
+🚀 Pilot launched! The initial version featured:
+
+- Data from 100+ proposals imported into the PDC
+- Keycloak-based user system with two-factor authentication via SMS
+- Dashboard of all proposals
+- Ability to drill into each proposal to view all its fields
+- Ability to toggle between field names and keys/shortcodes
--- a/src/pages/Landing.tsx
+++ b/src/pages/Landing.tsx
@@ -69,6 +69,16 @@ const Landing = () => {
               />
             </dd>
           </Dli>
+          <Dli>
+            <dt>Recent improvements</dt>
+            <dd>
+              <OffsiteLink
+                to="https://github.com/PhilanthropyDataCommons/data-viewer/blob/main/RECENT_CHANGES.md"
+              >
+                Read a summary of recent changes here.
+              </OffsiteLink>
+            </dd>
+          </Dli>
         </dl>
       </PanelBody>
     </Panel>


reefdog:
We don't have a `DEVELOPMENT.md`, but if we did we should also note that when a PR includes a user-facing change, the PR submitter should describe their change in `RECENT_CHANGES.md` as part of that PR.

RECENT_CHANGES.md
```diff
@@ -0,0 +1,26 @@
...
+## June 2023
+
+- Add the ability to show applicant data from data platform providers,
+  starting with Candid and Charity Navigator
+- Highlight current proposal in sidebar
+- Allow text selection on Dashboard table
+
+## May 2023
+
+- Add search capability to Dashboard
```
kfogel, line 12: I know that we added search capability in two stages: an initial "basic" search (which I believe was actually running client-side, IIRC), and then later that was replaced/improved by putting a real search engine into the Data Viewer.

So I expected either to see two items (at different chronological locations) about search, OR if this is the only item about search then I expected it to say something like "Improved search capability (added phrase searches, stemming, etc)" or something like that.

src/pages/Landing.tsx
```diff
@@ -69,6 +69,16 @@ const Landing = () => {
               />
             </dd>
           </Dli>
+          <Dli>
+            <dt>Want to know what we’ve been up to?</dt>
```
kfogel, line 73: I think the thing we want to focus the reader's attention on is the changes, not the team.  Also the "we" carries a slight danger of exclusion -- "we" are in here doing things, while "you" are out there reading about what we're doing -- whereas we'd rather phrase everything in a way that leaves the door open to participation.

So maybe just "See recent improvements to the PDC data viewer" or something like that?  I can help wordsmith collaboratively if you'd like, but I also trust your judgement now that I've said my piece above!

kfogel:
> We don't have a `DEVELOPMENT.md`, but if we did we should also note that when a PR includes a user-facing change, the PR submitter should describe their change in `RECENT_CHANGES.md` as part of that PR.

We usually call that file `CONTRIBUTING.md` (not that there's anything wrong with `DEVELOPMENT.md`, it's just that the former has been our convention).

OTS projects usually start with a minimal `CONTRIBUTING.md` that just references our [standard, good-for-any-project CONTRIBUTING.md](https://code.librehq.com/ots/meta/-/blob/main/CONTRIBUTING.md) and then adds any project-specific guidance needed.  Your point above about documenting recent changes can be one of those pieces of specific guidance.

(It's quite possible that the whole `RECENT_CHANGES.md` convention that we've started here will eventually migrate to become standard at OTS projects too, in which case we'll move material from here to the ur-`CONTRIBUTING.md`.  But that's to be decided later.  For now, the recent changes thing is just a project-specific convention here.)

RECENT_CHANGES.md
reefdog: @kfogel Since we deployed both search versions in May, I wasn't quite sure how to handle this one. Is it weirder to have "Add search" + "Improve search" as two separate items in May, or to have "Add and improve search" as a single item? (Using condensed language here.)

We could arguably fudge this one a _little_ and put "Add search" in April, because that is when the PR was opened. It just wasn't closed/deployed until May 1.

kfogel: @reefdog Oh, I see.  Well, I'm not sure we want to organize by calendar months anyway -- like, there might be several months when nothing major happens.  Ah, but I see the problem: we're using continuous deployment, so it's not like there's a "release" event.

The real releases, in a sense, have been the occasional Friday live demonstrations to the Pilot Participants group (and later, our release-equivalent will be demos or announcements to that group or whatever group succeeds it).  So maybe those announcements/demos should be the milestones that the recent changes list is organized by?

reefdog: @kfogel Yeah, our continuous release makes this a tricky problem!

If we think Friday demonstrations are a good milestone, I can reconstruct around those. Then perhaps—as you suggested on the call—we could have a "Pending" section at the top where new items accumulate until they're demoed.

I'd like to hear more input on this strategy before implementing it, since it'll require another pass through the history to reconstruct. If we don't get any feedback here we'll discuss tomorrow.

bickelj: Suppose there is a new, proposed change that has yet to be deployed. It sits in a Pull Request during review. Part of the change is the update to `RECENT_CHANGES`. No assumption can be made about the date when that change will be visible to the user because it has yet to pass review. I'm not sure what to suggest as a solution. Some kind of date reference is important. Perhaps the date should only reference past months and the top group is merely "Recent"? That way the list at the top can be built up with no assumed date, and when weeks or months pass, the date can be inserted.

Even though the software is only useful to people who actually use the software, I think the "done" definition should be "when it is available in production" rather than "when users get notified" or "when it gets demonstrated." This idea is about a "done" definition so it's indirectly related to the RECENT_CHANGES doc. Suppose a scheduled demo gets skipped or two demos happen within two weeks. I think I would prefer a more even interval in this document if there is an interval chosen at all.

Perhaps the most recent change should have no date on it, while each and every previous change needs to have a date associated. So if I were adding "...ability to show applicant data..." it might look like this:
```
- Add the ability to show applicant data from data platform providers,
  starting with Candid and Charity Navigator
- 2023-06-23 Highlight current proposal in sidebar
- 2023-06-23 Allow text selection on Dashboard table
...
```

And now that I try to do it, I see it is pretty tedious. I wish I had a better suggestion.
...
```

kfogel: @reefdog Your solution sounds fine, including the Pending section.

@bickelj Eh, easy enough to deal with the problem of the floating change: just include it in the PR, and then adjust the location right before merging.

jasonaowen: I like the monthly approach! In general, I'd prefer to minimize the amount of friction involved in making changes, and the monthly cadence feels like a useful trade-off between sharing updates without too much rebasing / merge conflicts / date tracking.

> there might be several months when nothing major happens

That's true! But I think it affects each approach similarly. If anything, grouping by month smooths over some of the demo scheduling concerns.

> I think the "done" definition should be "when it is available in production"

I agree!

I'd also love to hear feedback from our intended audience on any of this. *Do* they need or want more than a monthly update? Are they even going to click on the new link? Can we start with the lower-friction approach first and then refine it when we get that feedback?

All that said, I'll defer to @kfogel and @jim-mcgowan on this.

kfogel: Let's please just do monthly then; I agree with your assertion @jasonaowen that it's a good trade-off point between time-based and demo-based tracking.

It's my fault for asking the question about the two stages of search implementation in the first place :-).  I should have left well enough alone and not asked a fairly unimportant question that interfered with actually shipping this new file.  My bad!  I'll try to consider more carefully next time whether my question is *really* contributing.

reefdog: @kfogel Conversation that reinforces a direction can still be useful, and in this case I think it was! I'm glad you raised it.

src/pages/Landing.tsx
reefdog: @kfogel How about simply "Recent improvements" as the title?

reefdog:
@kfogel `CONTRIBUTING.md` is great. I opened #251 to capture the need for that.

RECENT_CHANGES.md
jasonaowen: Yes, +1 to what @reefdog just said - I think that was a good conversation to be had, @kfogel! No fault to assign.

jasonaowen (approved):
Now that @kfogel has signed off on the content, I'll +1 the form! The link works and looks good. Thanks, @reefdog!

src/pages/Landing.tsx
kfogel: @reefdog Sounds good to me!

PR #249, opened 2023-06-26
Improve interactive button styling

### LINKED ISSUE #173: Improve interactive button styling
[Buttons](https://pilot.philanthropydatacommons.org/storybook/?path=/docs/stories-button--docs) are currently very static: there is no hover state to indicate their interactivity and no alternate "active" state when being depressed. There is only a very clumsy `:focus` ring for minimal accessibility, but it is a thick garish blue outline that rarely fits with the aesthetic of the particular button style.

We should at least add hover styling and theme-specific focus styling, plus consider an active style.

reefdog:
Prior to this PR, the only interaction that button/form elements had was to gain a garish blue outline when focused. This PR:

- Adds hover state (slight darkening of background)
- Uses the hover state as the focus state
- Adds the outline to the active state (when the button is being pressed)
  - It pointedly does _not_ use the outline for the focus state anymore, since that remains after click (until the element is "blurred" but clicking/interacting elsewhere), and the outline is too large of a UI element to leave vestigially.

It also removes the focus ring from the app navbar, which now just has a simple background color change for both hover and focus.

**Testing:**
- `npm run storybook` and tinker with the [Story](http://localhost:6006/?path=/docs/stories-button--docs)
- `npm start` and make sure everything looks nice

Closes #173

### DIFF
--- a/src/components/FormElementGroup.css
+++ b/src/components/FormElementGroup.css
@@ -3,7 +3,7 @@
   gap: 0;
 }
 
-.form-element-group > *:focus {
+.form-element-group > *:is(:focus, :active) {
   z-index: 1;
 }
 
--- a/src/App.css
+++ b/src/App.css
@@ -52,15 +52,11 @@ button.App-navbar__item {
   cursor: pointer;
 }
 
-.App-navbar__item:focus {
-  outline: 2px solid blue;
-}
-
 .App-navbar__item.active {
   color: var(--color--blue);
 }
 
-.App-navbar__item:is(a, button):hover {
+.App-navbar__item:is(a, button):is(:hover, :focus) {
   background-color: #0001;
 }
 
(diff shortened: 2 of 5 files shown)


src/components/FormElementGroup.css
```diff
@@ -3,7 +3,7 @@
   gap: 0;
 }
 
-.form-element-group > *:focus {
+.form-element-group > *:is(:focus, :active) {
```
jasonaowen, line 6: > In a group of form elements — where elements are immediately adjacent — we “lift” the z-index of the focused element so its focus ring displays above its siblings.

This is a great explanation - thank you for including it in the commit message!

jasonaowen (approved):
This is great! I was able to test all the changes, and they look good. Thanks, @reefdog!

PR #252, opened 2023-06-28
Add "improve button styling" to RECENT_CHANGES

reefdog:
PR #249, which improved button styling, was done simultaneously with the creation of the `RECENT_CHANGES.md` doc. Thus, we weren't able to add a note about it until after that doc existed.

Now it does, so retroactively adding it.


bickelj (approved):
I didn't test it, but it looks good!

Issue #251, opened 2023-06-27
Add CONTRIBUTING.md


reefdog:
We should have a `CONTRIBUTING.md` file that describes how to participate in the project. This came up when discussing how to make sure developers populated the `RECENT_CHANGES.md` doc as they added features that fit inclusion in #248.

@kfogel in https://github.com/PhilanthropyDataCommons/data-viewer/pull/248#issuecomment-1605839997:

> > We don't have a `DEVELOPMENT.md`, but if we did we should also note that when a PR includes a user-facing change, the PR submitter should describe their change in `RECENT_CHANGES.md` as part of that PR.
> 
> @reefdog We usually call that file `CONTRIBUTING.md` (not that there's anything wrong with `DEVELOPMENT.md`, it's just that the former has been our convention).
> 
> OTS projects usually start with a minimal `CONTRIBUTING.md` that just references our [standard, good-for-any-project CONTRIBUTING.md](https://code.librehq.com/ots/meta/-/blob/main/CONTRIBUTING.md) and then adds any project-specific guidance needed.  Your point above about documenting recent changes can be one of those pieces of specific guidance.

reefdog:
Per conversations in Zulip and on Jitsi, the approach we'll take:

1. We're going to copy OTS' [contributing guidelines](https://code.librehq.com/ots/meta/-/blob/main/CONTRIBUTING.md) into the PDC [meta repo](https://github.com/PhilanthropyDataCommons/meta/) rather than just reference them. We'll make a few edits to the guidelines that represent PDC team guidelines.
2. Individual PDC projects (like this one) will have `CONTRIBUTING.md` docs that point to the central PDC one, supplemented with any project-specific guidelines.

So essentially, just what @kfogel did in #287, except pointing to the PDC meta version of the file instead of the OTS one directly.

reefdog:
One change to the above: per discussion with @slifty about the magic of the `.github` repo, we're going to point there instead.

kfogel:
Thank you, @reefdog!

PR #287, opened 2023-08-09
Add contribution guidelines (issue #251)

kfogel:


jasonaowen (changes requested):
We discussed this as a group, and decided on a different approach. See https://github.com/PhilanthropyDataCommons/data-viewer/issues/251#issuecomment-1707168621 for more details.

Thanks for starting this conversation, @kfogel!

PR #310, opened 2023-09-06
Set Dependabot to open PRs at noon Pacific

jasonaowen:
Configure Dependabot to create pull requests during business hours.

https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduletime

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,6 +4,8 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+      timezone: "America/Los_Angeles"
+      time: "12:00"
     open-pull-requests-limit: 10
     ignore:
       - dependency-name: "@types/node"
@@ -18,3 +20,5 @@ updates:
     directory: ".github/workflows"
     schedule:
       interval: "daily"
+      timezone: "America/Los_Angeles"
+      time: "12:00"


reefdog (approved):

slifty:
I am not on the data-viewer part of the team so this doesn't impact me much, but I just wanted to mention that this change would be disruptive to my personal productivity flow if it expanded to other parts of the PDC.

I tend to want to cover dependabot notifications first thing in my day, and so having them opened ASAP is helpful as opposed to having them open mid-day.  I do think in general it'd be best for individual developers to configure their personal notification setup based on their own needs and preferences rather than codifying those preferences in the repo.

jasonaowen:
@slifty I would actually encourage you to set your dependabot configuration explicitly, although it sounds like not to the same time. From the [docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#scheduletime):

> By default, Dependabot checks for new versions at a random set time for the repository.

It might be the case that the current random times in the other repos are suitable for you, but in this repo it was opening PRs at around 11pm Pacific - early enough that my phone hasn't gone to quiet time for the night, but late enough to annoy me when my phone starts blowing up with work emails.

slifty:
🤯

PR #311, opened 2023-09-06
Group Dependabot updates to Storybook and React

jasonaowen:
We have two sets of dependencies that should only ever be updated together, and never separately. React updates rarely, but we'd previously disabled Dependabot updates to Storybook because there were so many components and Dependabot would open a PR for each.

However, GitHub [recently](https://github.blog/2023-08-24-a-faster-way-to-manage-version-updates-with-dependabot/) released support for [grouped updates](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups)! With this feature, we can define groups that Dependabot should open one collective PR for, rather than separate PRs.

Define groups for React and Storybook, and allow Dependabot to update Storybook again.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,10 +9,18 @@ updates:
       - dependency-name: "@types/node"
         update-types:
           - "version-update:semver-major"
-      - dependency-name: "*storybook*"
-        # Dependabot is not yet able to upgrade all the dependencies together,
-        # which leads to mixed versions and quite a lot of PR thrash.
-        # See https://github.com/dependabot/dependabot-core/issues/1190
+    groups:
+      storybook:
+        dependency-type: "development"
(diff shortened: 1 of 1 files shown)


slifty (approved):
Oh snap this is really great to know about.

As an aside... I wonder if it would be reasonable to group eslint updates similarly

jasonaowen:
@slifty it absolutely would, and is part of why I tagged you on this PR! In this repo, we're currently getting ESLint transitively, but in the service repo where you aren't - and here, when we eject from create-react-app - group ESLint updates together would be great.

jasonaowen:
Also, check out the first grouped PR: https://github.com/PhilanthropyDataCommons/data-viewer/pull/316. Very cool!

Issue #315, opened 2023-09-12
Choose new name for Data Viewer


reefdog:
The Data Viewer (project and application) is destined to allow data entry. Consequently, the name "Data Viewer" is not a fitting name. Either we need a read/write-focused name, or, this can be canonically referred to as something like "the front end" or "the PDC".

slifty:
Here's a Really Good Brainstormed List:

* Data Viewer, Bath, and Beyond
* Data Portal
* PDC Portal
* Data Viewer
* Data Hub / PDC Data Hub
* Data Viewdeditor
* Data Catalogue
* Data Catalog
* Data Cadalok
* Data Vuvuzeladlock
* Apple Inc.

I think I vote for PDC Portal

kfogel:
I think "PDC Portal" is fine (w/ `data-portal` as repository name).

I suspect that colloquially sometimes it will be referred to as the "PDC Data Portal" and sometimes as "the Portal", and we should just roll with that if so.

Let's just start using "PDC Portal" in writing, and any of the above in speech.  Meanwhile, Jim and I can mention it to the rest of the team and see if they have any strong opinions.

Okay to close this?

slifty:
*sigh* I'm instantly regretting not highlighting `Data Viewdeditor`

slifty:
UPDATE: others wanna weigh in so let's allow space for that!

jasonaowen:
Thanks for raising this, @reefdog! I feel like we've said it out loud a few times, but I appreciate you writing it in an issue. And thank you @slifty for starting the brainstorming process!

I feel like "portal" doesn't convey any useful information - it does not describe the repo, and if I were an outsider I don't think seeing a repo named "portal" would tell me much about its contents. (See also: [GitHub repo search for "portal"](https://github.com/search?q=portal&ref=cmdform&type=repositories), which contains a wide variety of projects.)

Likewise, including "data" in the name doesn't add much in the context of the "Philanthropy Data Commons".

(Separately, I'd prefer if the repo name did not share a prefix with any other repos in this org, as it makes command line tab-completion less convenient. That might mean we should rename `data-scripts` if we want this repo to be `data-something`.)

`front-end` seems like the most descriptive name so far, although it's also the most boring. I'm open to other ideas!

kfogel:
(Happy to continue thinking/discussing.)

[...]

PR #327, opened 2023-09-19
Rename repo from data-viewer to front-end

jasonaowen:
Update all the places that the repository name occurs. This does not update any in-app user-facing strings, per @kfogel's [suggestion](https://github.com/PhilanthropyDataCommons/data-viewer/issues/315#issuecomment-1716320859).

Issue #315 Choose new name for Data Viewer

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# data-viewer
+# front-end
 A UI for interacting with the data stored in the PDC service
 
 ## Dependencies
@@ -8,14 +8,14 @@ Open ID Connect ("OIDC")
 identity provider ("IdP")
 for authentication.
 It should be agnostic about what kind of IdP is used.
-To use the data-viewer application,
+To use the front-end application,
 you will need to be able to log in.
 
 ## Configuration
 
 ### Identity Provider
 
-The IdP needs a client for the data-viewer.
+The IdP needs a client for the front-end.
 The IdP should be configured for the client to use the
 [Authorization Code flow](https://oauth.net/2/grant-types/authorization-code/).
 This client is a public client,
--- a/src/pages/Landing.tsx
+++ b/src/pages/Landing.tsx
@@ -73,7 +73,7 @@ const Landing = () => {
             <dt>Recent improvements</dt>
             <dd>
               <OffsiteLink
-                to="https://github.com/PhilanthropyDataCommons/data-viewer/blob/main/RECENT_CHANGES.md"
+                to="https://github.com/PhilanthropyDataCommons/front-end/blob/main/RECENT_CHANGES.md"
               >
                 Read a summary of recent changes here.
               </OffsiteLink>
--- a/.env.example
+++ b/.env.example
@@ -16,7 +16,7 @@
 REACT_APP_OIDC_AUTHORITY=https://auth.example.com/realms/example
 
 # The client ID is assigned by the IdP
-REACT_APP_OIDC_CLIENT_ID=example-data-viewer
+REACT_APP_OIDC_CLIENT_ID=example-front-end
 
 # The API URL is the location of a running PhilanthropyDataCommons/service instance
 # https://github.com/PhilanthropyDataCommons/service
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "data-viewer",
+  "name": "front-end",
   "version": "0.1.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
-      "name": "data-viewer",
+      "name": "front-end",
       "version": "0.1.0",
       "dependencies": {
         "@axa-fr/react-oidc": "^7.4.1",
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "data-viewer",
+  "name": "front-end",
   "version": "0.1.0",
   "private": true,
   "dependencies": {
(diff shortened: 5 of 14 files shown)


jasonaowen:
We'll also need to rename the repo in the GitHub admin view; that can happen before or after this is merged, but should be close in time so that the "Recent Changes" URL continues to work.

README.md
```diff
@@ -1,4 +1,4 @@
-# data-viewer
+# front-end
```
slifty, line 1: This was auto-populated by github and while we're here changing it it might be nice to rename the title to something more human like "Philanthropy Data Commons Front End" (the service is currently titled "Philanthropy Data Commons Service")

slifty (changes requested):
I found a few more:

* https://github.com/PhilanthropyDataCommons/data-viewer/blob/2455150ec98224b655d95db4591d95edc71ddce7/public/index.html#L15
* https://github.com/PhilanthropyDataCommons/data-viewer/blob/2455150ec98224b655d95db4591d95edc71ddce7/public/manifest.json#L3
* https://github.com/PhilanthropyDataCommons/data-viewer/blob/2455150ec98224b655d95db4591d95edc71ddce7/src/pages/Landing.tsx#L23

Not sure if we care about this one:

* https://github.com/PhilanthropyDataCommons/data-viewer/blob/2455150ec98224b655d95db4591d95edc71ddce7/src/interfaces/DataViewerProposal.ts#L1

Also, a suggestion to give the ReadMe a more robust title rather than using the repo slug.

slifty:
Seeing now that you intentionally skipped those -- I thought Karl was suggesting we not get bogged down thinking about branding, but I didn't think he meant we should keep the name Data Viewer for the frontend.

What about renaming it to PDC Data Portal for user facing strings (or just calling it the Philanthropy Data Commons) as part of this PR?  That doesn't prevent us from being able to change it again later, and it would still accomplish karl's suggestion of decoupling the public name from the repository name.

reefdog:
@slifty I think it would be trivial to do any suggested branding/user-facing changes as a follow-up PR. Jason's changes are essentially non-controversial ("change IDs and repo references"), while your suggested changes warrant [even a tiny amount of] discussion. For that reason alone I suggest merging this one in and then following up right away with your suggestion!

reefdog (approved):
While acknowledging @slifty's desire to update user-facing names / branding, I do think this is a good PR as-is and am signaling _my_ vote by approving it. (That said, Dan if you feel stronger then by all means you can still formally request changes!)

README.md
reefdog: I do think (whether in this PR or a followup) that re-titling the README is a good idea.

reefdog:
@slifty:

> Not sure if we care about [`DataViewerProposal`]

Ooh, good catch. I actually *do* think we care about it and it should be renamed `FrontEndProposal`, but also, it's a small red flag that we have to rename an interface just because the project name changed. Though it does _kinda_ make sense, because this is the interface for "how this front end models a proposal" (as opposed to how the back-end service models it).

README.md
jasonaowen: Let's do that as part of the follow-up rebrand PR!

slifty:
For sure -- my `request changes` came before realizing it was intended - happy to mark as approved (though yeah the DataViewerProposal update makes sense to make if others agree!)

README.md
slifty: FWIW I don't think the markdown title would be part of the rebrand since it is a dev document.  If there's a reason we don't want to change this to "Philanthropy Data Commons Front End" now that's fine, but I don't think this file should ever say "Philanthropy Data Commons Data Portal" or whatever we end up landing on for the branded name.

slifty (approved):
I left my feedback!  Take it or leave it but I will not stand in the way of progress 😅

README.md
reefdog: I do agree this file's title will almost certainly differ from the landing page title, but I would reframe the follow-up PR from "branding" (even though it involves that) to "user-facing language", where users are both website visitors and devs reading documentation. In that case, both the language on the landing page and the README title are fair game (IMOoooocow).

reefdog: Whoops I edited out my soft-landing "What do you think?" when rearranging sentences, so…

What do you think?

slifty: > I feel like the common convention is to have the title match the repo name

Got it!  I had always interpreted that as more of "it's the default and nobody got around to making a real title" BUT I'm also fine just punting / leaving as proposed!

Carry forth!  The bike shed is beautiful regardless!

jasonaowen: :bike: :house: 

I did a little further research and didn't find anything particularly persuasive in one direction or the other! So, now that I have your blessing, I'm going to pull the rip cord on this and leave it alone for now. :blue_heart: :parachute:

src/pages/Landing.tsx
```diff
@@ -73,7 +73,7 @@ const Landing = () => {
             <dt>Recent improvements</dt>
             <dd>
               <OffsiteLink
-                to="https://github.com/PhilanthropyDataCommons/data-viewer/blob/main/RECENT_CHANGES.md"
+                to="https://github.com/PhilanthropyDataCommons/front-end/blob/main/RECENT_CHANGES.md"
```
jasonaowen, line 76: I just tested this by clicking through the deploy preview and the new URL is now live.

reefdog: Sigh how I wish we could wildcard the auth domain. :/

jasonaowen: Me, too. Someday...

Actually, this might be worth re-evaluating separately - it might be worthwhile to do some legwork to enable deploy preview logins automatedly! It won't be out-of-the-box, but I think it should be possible.

reefdog (approved):
The code scans cleanly and I pulled locally / logged in to be sure the interface change still worked. It does indeed!

README.md
kfogel: > Carry forth! The bike shed is beautiful regardless!

Obligatory https://fuchsia.bikeshed.com/ .

PR #366, opened 2023-10-11
Add Contributing and Code of Conduct files

### LINKED ISSUE #251: Add CONTRIBUTING.md

reefdog:
This PR adds the `CONTRIBUTING.md` and `CODE_OF_CONDUCT.md` files that govern contributions and participant conduct.

As per multiple discussions, individual repos will contain links to the shared versions in the PDC `.github` repository, plus any repo-specific instructions.

Closes #251

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,26 @@
 # Contributing
 
 If you're interested in contributing to this project,
-please read the
+please start with the
 [general PDC contribution guidelines](https://github.com/PhilanthropyDataCommons/.github/blob/main/CONTRIBUTING.md).
+
+## Guidelines specific to this project
+
+### Populating "Recent Changes"
+
+If your change affects the user experience,
+describe it in concise, non-technical language
+in the `RECENT_CHANGES.md` document
+under a heading for the current month.
+Include this description in the PR
+that contains your change.
+
+Consult the existing document
+for a sense of what sort of changes
+justify inclusion in `RECENT_CHANGES.md`,
+and for how to write a good description.
+
+As a reviewer,
+be sure to review `RECENT_CHANGES.md` additions as well —
+and if the submitter didn't write one,
+be sure to consider whether they should.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,5 @@
+# Contributing
+
+If you're interested in contributing to this project,
+please read the
+[general PDC contribution guidelines](https://github.com/PhilanthropyDataCommons/.github/blob/main/CONTRIBUTING.md).
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,5 @@
+# Code of Conduct
+
+If you're interested in contributing to this project,
+please read the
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,26 @@
...
+
+If you're interested in contributing to this project,
+please start with the
+[general PDC contribution guidelines](https://github.com/PhilanthropyDataCommons/.github/blob/main/CONTRIBUTING.md).
+
+## Guidelines specific to this project
+
+### Populating "Recent Changes"
+
+If your change affects the user experience,
```
slifty, line 11: You seem to be wrapping at ~40 characters which is fairly aggressive.

I think an 80 character wrap might be nicer?  It looks like the main CONTRIBUTING.md uses 72.

(I personally think wrapping in markdown is a losing battle but I know I'm in the minority here)

slifty (approved):
The line breaks make it a little tough to read in plaintext mode as written; suggesting you adopt the pattern of CONTRIBUTING and use a 72 line wrap (or, since it's just you and me... give into the chaos and don't manually enter line wraps at all!)

Feel free to ignore this comment and merge as is -- I'm marking as approved.

CONTRIBUTING.md
reefdog: I've adopted @jasonaowen's manner of semantic line breaks. Despite the semblance of a [spec](https://sembr.org/), IMO the core philosophy is to reduce each line to a discrete phrase that could be changed with minimal diff confusion on the surrounding text.

That said, maybe it's a losing battle and diff engines are good enough at highlighting mid-line changes.

If this is a strong objection I'll defer!

slifty: I do think that when using a text editor (as opposed to reading in a web browser that flattens the whitespace) that approach becomes harder to read -- though I guess it's just more like reading poetry, where you can truly appreciate the WORDS being used and the shape of the prose...  that counts for *something*!

Honestly, whatever floats yinz's boats is fine (but please note that my approval won't stop be from making *fun* of the approach!)

reefdog: Grokked and I see your point!

However, in the interest of velocity (and with your overall approval), I'll save this discussion for another day.

Issue #333, opened 2023-09-21
Remove "pilot" terminology


reefdog:
According to @jim-mcgowan via a Zoom meeting just now, we are good to remove the "pilot" terminology! 🎉 With the resolution of #315 (and once #332 is merged), this is just "the Philanthropy Data Commons", full stop.

This also means we should choose a new subdomain. The bare domain remains the [marketing site](https://philanthropydatacommons.org/), and `api` remains [the API](https://api.philanthropydatacommons.org/). It's not uncommon, when there are separate platforms for "the marketing site" and "the app itself" — and the app isn't namespaced under the user's organization or something (e.g., `ots.basecamp.com`) — to use `app` as the subdomain. E.g., https://clockify.me vs. https://app.clockify.me.

**So, my recommendation is** `app.philanthropydatacommons.org` with `pilot.philanthropydatacommons.org` redirecting permanently to it.

My only concern is that it's only one character away from `api.philanthropydatacommons.org`.

Thoughts?

cc @jasonaowen @bickelj @slifty @kfogel @jim-mcgowan @jmergy

reefdog:
The copy edits have been made in PR #334. I realized, happily, the subdomain isn't referenced anywhere in our front-end code. We can thus merge in that PR independent of all the switch-flipping that accompanies configuring a new domain.

kfogel:
I don't think the "app"<->"api" [Levenshtein Distance](https://en.wikipedia.org/wiki/Levenshtein_distance) of 1 is a problem.  People using the API are generally typing that address once into a script somewhere -- if they get it wrong, they'll find out quickly and correct it, and thereafter they'll never have to worry about it again.  Meanwhile, people using the app mostly won't know that "api" is a thing and it's an unlikely typo for them to make.  And for developers working on the project, well, developers already know that one character can make a huge fifference!

reefdog:
@jasonaowen Tag! Could you configure Netlify to accept `app` and do a manual deployment? (We should wait to configure it as the automated `production` deployment target until later, of course.) And could you drop in whatever DNS info @jmergy will need — assuming it's different from the `pilot` settings?

reefdog:
Per @jim-mcgowan in today's meeting, we actually _don't_ need to wait to purge pilot-only data before doing this, asserting users won't pay attention to the URL enough for it to be a stronger signal than what we just tell them.

[...]

PR #420, opened 2023-11-21
Redirect from pilot.pdco to app.pdco

jasonaowen:
Configure Netlify to redirect requests.

Netlify allows [redirects between the domains of an app](https://docs.netlify.com/domains-https/custom-domains/multiple-domains/#domain-redirects) by using the [`_redirects` file](https://docs.netlify.com/routing/redirects/redirect-options/#domain-level-redirects).

This should not be merged until DNS is configured and confirmed working!

Issue #333 Change subdomain from pilot to app

### DIFF
--- a/RECENT_CHANGES.md
+++ b/RECENT_CHANGES.md
@@ -1,5 +1,10 @@
 # Recent Changes to the PDC
 
+## November 2023
+
+- Renamed the domain from `pilot.philanthropydatacommons.org`
+  to `app.philanthropydatacommons.org`
+
 ## June 2023
 
 - Add the ability to show applicant data from data platform providers,
--- a/_redirects
+++ b/_redirects
@@ -1,0 +1,2 @@
+# Move from pilot.pdco to app.pdco
+https://pilot.philanthropydatacommons.org/* https://app.philanthropydatacommons.org/:splat 301!


jasonaowen:
Unfortunately, I don't think we'll be able to test this change! We can take some assurance from the [redirect rules status check](https://github.com/PhilanthropyDataCommons/front-end/pull/420/checks?check_run_id=18874585467), which says:

> Successful in 42s — 1 redirect rule processed

> All redirect rules processed without errors.

and the Netlify deploy log which says the same. Since only redirects to `pilot.pdco` will trigger the redirection, I think we'll just have to test it once it's been deployed and adjust from there.

jasonaowen:
The DNS changes are live, and I've marked `app.pdco` as the primary domain in Netlify. I believe this is ready to go!

reefdog (approved):
I do have one tiny request: can you add a line item to `RECENT_CHANGES.md`? It feels like a worthy bullet point! However, I also want to signal my approval otherwise, so I'm going to approve this… with my original approval message below.

-----

Huzzah! To reiterate your point, we kind of have to accept/merge this in order to see if it works… but it _looks_ correct and I'm cool with making it so.

On a recent call, I believe it was @jim-mcgowan who said most users won't even notice the change, so deploying without **first** notifying stakeholders doesn't feel like a violation of the spirit of https://github.com/PhilanthropyDataCommons/meta/issues/70. And since credentials are entered on a different domain entirely (`auth.pdco`), this won't even require people to fiddle with their password managers.

As far as I'm concerned, feel free to merge this when ready!

jasonaowen:
Great catch, @reefdog! Thank you, updated and rebased.

jasonaowen:
:facepalm: I did not read the [documentation on redirects](https://docs.netlify.com/routing/redirects/) correctly.

> Save a plain text file called `_redirects` without a file extension to the [publish directory](https://docs.netlify.com/configure-builds/overview/#definitions) of your site.

We had that already: https://github.com/PhilanthropyDataCommons/front-end/blob/c551844410d92a73507a76e30e9b68ec6ca90f6b/public/_redirects#L3

That's the rule that it was processing. The new file was ignored completely. Sorry, folks! I'll clean this up in a subsequent PR.

Issue #409, opened 2023-11-18
Create bulk upload route and page


reefdog:
We need an "Add Data" page that currently defaults to a bulk uploader in the sidebar and instructional copy in the main pane:

![Upload CSV](https://github.com/PhilanthropyDataCommons/front-end/assets/4731/da448919-e554-4a77-9c95-22cc2c2c938c)

This will use a [Panel Grid Plus Sidebar](https://pilot.philanthropydatacommons.org/storybook/?path=/story/stories-panelgrid--plus-sidebar).

Suggestions welcome for the path. This is easily changeable in the future and shouldn't tie us up. I recommend a simple `/add-data`.

jmergy:
Like the sidebar use

reefdog:
This has been done!

PR #421, opened 2023-11-21
Refactor to new react-router data router

jasonaowen:
The React Router documentation [recommends](https://reactrouter.com/en/main/routers/create-browser-router) using the new data router approach for all projects. This is a fairly faithful and simple refactoring following the [migration guide](https://reactrouter.com/en/main/upgrading/v6-data). It should have all the same behavior as the existing application, and would allow us to use the new APIs that are only available in data router apps.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Route, Routes } from 'react-router-dom';
+import { BrowserRouter, Route, Routes } from 'react-router-dom';
 import { Layout } from './Layout';
 import { NotFound } from './pages/NotFound';
 import { ProposalDetail } from './pages/ProposalDetail';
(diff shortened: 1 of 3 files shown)


jasonaowen:
I wanted to do this migration before adding any new pages (such as for #409), so as to reduce later rework and to hopefully be able to use the new react-router APIs on new pages.

Unfortunately, it doesn't appear that the new react-router data loading APIs are compatible with react-oidc, as there isn't - yet - a natural way to restrict data loading requests to a logged-in session. react-oidc wants to handle that with components and hooks, while the data loading APIs want to be static and outside the React component tree. I wasn't able to unify those two very different approaches.

Still, I think this is worthwhile on its own, even if we never use any of the new APIs, if only for future proofing.

reefdog (approved):
This code scans much better and testing in-browser works great. Thanks!

Issue #410, opened 2023-11-18
Add navbar link to bulk upload page


reefdog:
- Icon: Solid `document-plus` from Heroicons
- Text: "Add Data"

PR #422, opened 2023-11-21
Show Storybook icon on deploy previews

jasonaowen:
We are using Netlify to build and deploy the front-end, and it is able to deploy pull requests to temporary domains. Often, one of the things we want to check on a deploy preview is that our components look good in Storybook - and, indeed, that Storybook is working correctly.

Previously, after Netlify gave us a deploy preview, we would follow the link it posted in the PR and then manually add `/storybook` to the end. Now, there can be a link in the app!

The link only renders when the app is built with the environment variable `REACT_APP_SHOW_STORYBOOK` set to the value `true` - otherwise, and by default, the new navbar icon will not be shown. Netlify allows for deploy-preview-only environment variables, and I have configured it to activate this feature.

Note also this won't really work with `npm start`: the link will render, but since the dev server doesn't know about Storybook, you'll get the 404 handler.

### DIFF
--- a/src/components/AppNavbar.tsx
+++ b/src/components/AppNavbar.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { NavLink } from 'react-router-dom';
-import { EnvelopeIcon } from '@heroicons/react/24/outline';
+import { BookOpenIcon, EnvelopeIcon } from '@heroicons/react/24/outline';
 import { TableCellsIcon } from '@heroicons/react/24/solid';
 import { User } from './User';
 
@@ -13,6 +13,14 @@ const AppNavbar = () => (
           Dashboard
         </NavLink>
       </li>
+      {process.env.REACT_APP_SHOW_STORYBOOK === 'true' && (
+        <li>
+          <a href="/storybook" className="App-navbar__item">
+            <BookOpenIcon />
+            Storybook
+          </a>
+        </li>
+      )}
       <li>
         <a
           href="mailto:info@philanthropydatacommons.org?Subject=Feedback%20on%20the%20Philanthropy%20Data%20Commons"


jasonaowen:
Screen Shot 2023-11-20 at 20 30 28

src/components/AppNavbar.tsx
```diff
@@ -1,12 +1,20 @@
...
 import { NavLink } from 'react-router-dom';
-import { EnvelopeIcon } from '@heroicons/react/24/outline';
+import { BookOpenIcon, EnvelopeIcon } from '@heroicons/react/24/outline';
 import { TableCellsIcon } from '@heroicons/react/24/solid';
 import { User } from './User';
 
 const AppNavbar = () => (
   <nav className="App-navbar">
     <ul>
+      {process.env.REACT_APP_SHOW_STORYBOOK === 'true' && (
```
reefdog, line 10: Oof! `true` as a string gives me hives. (But, I know env vars are always strings.) No changes requested here, this is Correct, it's just also in a larger way Incorrect. 😆

reefdog (changes requested):
Great idea!

My only request is that you move it to the right of Dashboard (still left of Feedback). Even us devs/testers benefit from having that "first item == Dashboard/Home" usability (sort of but not quite Fitts' Law), which I expect to reference more often than Storybook.

jasonaowen:
@reefdog interesting! Yeah happy to move it, but I was actually thinking this would be the least disruptive place to add it, as I didn't want to conditionally change the location of icons. I think of those navbar items as being right-to-left, to the point where I was considering floating with you the idea that we should reverse the order of the list in the code and make it render right-to-left via CSS.

jasonaowen:
We talked about this today, and @reefdog was convincing! I've updated the PR to put the new icon in between Dashboard and Feedback, and when we work on #410 that'll go between Dashboard and Storybook, so that the full list from left to right is:

- Dashboard
- Add Data
- Storybook
- Feedback
- User / login

reefdog (approved):
Huzzah! Thank you @jasonaowen!

PR #424, opened 2023-11-21
Add initial outline of data page

jasonaowen:
Build out the route, an initial page, and a few placeholder components for the new Add Data page, and wire the placeholder components into Storybook.

There are still several open questions that this does not attempt to resolve, including:

- what information do we get from the API about previous bulk uploads?
- what exactly do we want the prose to say?
- how exactly will the upload component work?

and other such future problems. For now, lay a foundation for additional work to build on.

Note that this is built on top of #421, so this is marked as a draft PR until that is merged and the merge base is reset to `main`.

Issue #409 Create bulk upload route and page

### DIFF
--- a/src/components/BulkUploader.tsx
+++ b/src/components/BulkUploader.tsx
@@ -1,0 +1,22 @@
+import React from 'react';
+import './BulkUploader.css';
+import { Panel, PanelBody } from './Panel';
+
+interface BulkUploaderProps {
+  history: object[]; // todo
+  loading?: boolean;
+}
+
+/* eslint-disable @typescript-eslint/no-unused-vars, arrow-body-style */
+export const BulkUploader = ({
+  history,
+  loading = false,
+}: BulkUploaderProps) => {
+  return (
+    <Panel>
+      <PanelBody>
+        <p>TODO: bulk uploads list</p>
+      </PanelBody>
+    </Panel>
+  );
+};
--- a/src/pages/AddData.tsx
+++ b/src/pages/AddData.tsx
@@ -1,0 +1,49 @@
+import React, { useEffect } from 'react';
+import { withOidcSecure } from '@axa-fr/react-oidc';
+import { useBaseFields } from '../pdc-api';
+import { PanelGrid, PanelGridItem } from '../components/PanelGrid';
+import { AddDataInstructions } from '../components/AddDataInstructions';
+import { BaseFields } from '../components/BaseFields';
+import { BulkUploader } from '../components/BulkUploader';
+
+const BulkUploaderLoader = () => {
+  const history: object[] = []; // todo
+  return (
+    <PanelGridItem>
+      <BulkUploader history={history} loading />
+    </PanelGridItem>
+  );
+};
+
+const BaseFieldsLoader = () => {
+  const fields = useBaseFields();
+  if (fields === null) {
+    return (
+      <BaseFields fields={[]} loading />
+    );
+  }
+
+  return (
+    <BaseFields fields={fields} />
+  );
+};
+
+const AddDataLoader = () => {
+  useEffect(() => {
+    document.title = 'Add Data - Philanthropy Data Commons';
+  }, []);
+
+  return (
+    <PanelGrid sidebarred>
+      <BulkUploaderLoader />
+      <PanelGridItem>
+        <AddDataInstructions>
+          <BaseFieldsLoader />
+        </AddDataInstructions>
+      </PanelGridItem>
+    </PanelGrid>
+  );
+};
+
+const AddData = withOidcSecure(AddDataLoader);
+export { AddData };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { createBrowserRouter, RouterProvider } from 'react-router-dom';
 import { Layout } from './Layout';
+import { AddData } from './pages/AddData';
 import { NotFound } from './pages/NotFound';
 import { ProposalDetail } from './pages/ProposalDetail';
 import { ProposalList } from './pages/ProposalList';
@@ -12,6 +13,7 @@ const router = createBrowserRouter([
     Component: Layout,
     children: [
       { path: '/', Component: Landing },
+      { path: '/add-data', Component: AddData },
       { path: '/proposals/:proposalId/provider?/:provider?', Component: ProposalDetail },
       { path: '/proposals', Component: ProposalList },
       { path: '*', Component: NotFound },
--- a/src/components/AddDataInstructions.tsx
+++ b/src/components/AddDataInstructions.tsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import './AddDataInstructions.css';
+import { Panel, PanelBody } from './Panel';
+
+interface AddDataInstructionsProps {
+  children?: React.ReactNode;
+}
+
+export const AddDataInstructions = ({
+  children = null,
+}: AddDataInstructionsProps) => (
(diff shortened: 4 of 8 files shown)


src/pages/AddData.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { withOidcSecure } from '@axa-fr/react-oidc';
+import { useBaseFields } from '../pdc-api';
+import { PanelGrid } from '../components/PanelGrid';
+import { BaseFields } from '../components/BaseFields';
+import { BulkUploader } from '../components/BulkUploader';
+
+const BulkUploaderLoader = () => {
```
jasonaowen, line 8: ...in particular, `BulkUploaderLoader`

jasonaowen (commented):
@reefdog I'm not married to any of the names, so any renaming suggestions are very welcome!

jasonaowen:
This was a draft because it depended on the react-router changes from #421; that's merged now, so it's ready to review!

reefdog:
@jasonaowen Eww, I didn't realize Netlify only generates deploy previews when the base is `main` at PR creation time. Hmm. This does put a damper on PR-laddering.

jasonaowen:
Yeah, it's unfortunate. The only [discussion](https://answers.netlify.com/t/trigger-a-deploy-preview-on-a-pull-request-after-changing-base/53009/2) I've been able to find about it had no resolution.

I feel like maybe closing and reopening the PR could trigger it? But I'm not sure how important this particular deploy preview might be.

src/pages/AddData.tsx
```diff
@@ -0,0 +1,36 @@
...
+
+  return (
+    <BaseFields fields={fields} />
+  );
+};
+
+const AddDataLoader = () => (
+  <PanelGrid sidebarred>
+    <BulkUploaderLoader />
+    <BaseFieldsLoader />
```
reefdog, line 31: Two questions:

1. I was hoping you'd go ahead and generate the panels for these as part of the task (though perhaps not part of this PR, if you're multi-staging it). My apologies, that was only lightly implied through my linking to the [Panel Grid Plus Sidebar](https://pilot.philanthropydatacommons.org/storybook/?path=/story/stories-panelgrid--plus-sidebar) component in the task description. Do you plan to deliver that as a separate PR, or does it make sense as this one? (The content can still be all "TODO"!)
2. The main (right-hand) panel is actually more than just base fields description, it's the "instructional text panel". The base field table is the most important bulk of it, but I think the hierarchy should be an instructional panel at the top with the base field table as just one of the content children. With that in mind, I think perhaps we need a different component here, which then internally renders `<BaseFieldsLoader />`?

reefdog (changes requested):
Thanks for this, excellent scaffolding!

I have a few requested changes, probably most significantly the question about whether you want to render the (empty) panels themselves as part of this PR or a follow-up; either way, I thought they'd be part of the scope of #409 so that other work can just fill in the panels (rather than creating them). What do you think?

reefdog:
> Yeah, it's unfortunate. The only [discussion](https://answers.netlify.com/t/trigger-a-deploy-preview-on-a-pull-request-after-changing-base/53009/2) I've been able to find about it had no resolution.
> 
> I feel like maybe closing and reopening the PR could trigger it? But I'm not sure how important this particular deploy preview might be.

Don't worry about it! We can run this locally.

src/pages/AddData.tsx
reefdog: …or perhaps all that's needed to rename `BaseFieldsLoader` to `AddDataInstructionPanelLoader` or something? I assume we want the API call to be as close to the relevant child as possible. Does that mean the panel makes the call, or just the section of the panel that contains the base fields?

src/pages/AddData.tsx
```diff
@@ -0,0 +1,36 @@
...
+    );
+  }
+
+  return (
+    <BaseFields fields={fields} />
+  );
+};
+
+const AddDataLoader = () => (
+  <PanelGrid sidebarred>
```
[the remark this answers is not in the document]
jasonaowen: Oh, yes, definitely! Great catch.

src/components/BulkUploader.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import './BulkUploader.css';
+
+interface BulkUploaderProps {
+  history: object[]; // todo
```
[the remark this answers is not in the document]
jasonaowen: My thinking was that you could update it to be the shape you want it to be, and then that would help inform the work @slifty is doing on the API side and/or the loader could bridge the gap.

src/pages/AddData.tsx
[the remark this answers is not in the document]
reefdog: It's such a silly way to have to add a page title. 😆

src/pages/AddData.tsx
jasonaowen: > I was hoping you'd go ahead and generate the panels for these

Ah sure, I can handle that as part of this PR!

> With that in mind, I think perhaps we need a different component here, which then internally renders `<BaseFieldsLoader />`?

> …or perhaps all that's needed to rename `BaseFieldsLoader` to `AddDataInstructionPanelLoader` or something? I assume we want the API call to be as close to the relevant child as possible. Does that mean the panel makes the call, or just the section of the panel that contains the base fields?

The approach I've been taking - and we can certainly revisit it - is to separate all of the API calls out into the `<...Loader>`s which live in the `src/pages/` directory. Under that model, we'd pass in the base fields to the component that renders it. That allows us to use Storybook more easily, and hopefully makes it easier for you to create components as needed!

Perhaps these components should have a parent-child relationship?

```jsx
<AddDataInstructionPanel>
  <BaseFieldsLoader />
</AddDataInstructionPanel>
```

jasonaowen:
Oh hey, pushing gets us a deploy preview! Yay.

src/pages/AddData.tsx
jasonaowen: Alternatively, they could be peers, and maybe we could even have some visual separation between the instructions and the base field list? Although we don't have any pages with multiple panels in a vertical alignment, and in fact displaying multiple `<Panel>`s inside a single `<PanelGridItem>` causes them to stack on top of each other in a last-one-wins scenario.

What do you think, @reefdog?

reefdog:
```jsx
<AddDataInstructionPanel>
  <BaseFieldsLoader />
</AddDataInstructionPanel>
```

This is exactly how I was envisioning it! Albeit eventually with a sibling component to `<BaseFieldsLoader />` that contained the instructional text. Referencing the image:

Upload CSV

We'd end up with some static content above the BaseFieldsLoader, e.g. (in this very much pseudo-markup):

```
<AddDataInstructionPanel>
  <h1>Add data to the PDC</h1>
  <p>Adding data via the API…</p>
  <p>Adding data via the bulk uploader…</p>
  <h2>Base Fields</h2>
  <p>Base fields are the backbone…</p>
  <BaseFieldsLoader />
</AddDataInstructionPanel>
```

With the exception that maybe the base field static content would also be within the loader, I don't know. But I think you see what i mean here?

reefdog: Does that fit your model, too?

jasonaowen: That makes sense! I've updated the branch with these changes. Let me know if this is closer to what you need!

jasonaowen:
Screen Shot 2023-11-21 at 15 13 48

reefdog (approved):
Super, thank you!!

PR #443, opened 2023-11-30
Add BulkUpload history components

### LINKED ISSUE #413: Create upload history sidebar components
Users will be able to view a history of their bulk uploads (see #365 for full context):

Upload History

We will need three states: Processing, Error, and Complete.

- Processing's icon should be a rotating animation
- Relative times (e.g., "2 minutes ago") should be [actual `<time>` elements](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/time)
- File size should be reported in nearest sensible unit (no decimals for KB; one decimal for MB; two decimals for GB)
- Error state does not currently report detailed error state; I'm still mulling on that

For the scope of this issue, we can assume what's shown here captures the full component need. Future features (like the ability to click on components to load detail in a panel) will come in separate/later issues.

### LINKED ISSUE #414: Populate file history sidebar
Using the components from #413, populate the file history sidebar, sorted newest (top) to oldest:

Upload sidebar

### LINKED ISSUE #492: Add call-to-action to BulkUpload "failed" message
Until detailed error reporting is completed, supplement "Failed [time] ago" message in `BulkUpload` history sidebar component with a call to action: "Failed [time] ago — email info@philanthropydatacommons.org for help".

reefdog:
This PR adds the BulkUpload history components and wires up the history to the API.

This is an incomplete PR but I'm hopping on a plane. Hopefully this establishes enough patterns and styling for you to continue from here! Still clearly TODO:

- [x] Demonstrate this in a panel
- [x] Humanize the byte size (1KB, 2MB, 1.1GB, etc.)
- [x] Humanize the time format
   - [x] Wrap the time in a `<time>` element with `datetime` attribute
- [x] Adapt to the real BulkUpload object attributes

**Update, January 19:**

Since writing the above, here's what's happened: Jason took over this branch, loaded BulkUpload history from the API, and wired up the components. (He also added file uploading on top of it, but that will be in a separate PR.) I collected that work into commits and added the humanized times and file sizes.

**Testing:**
- Check out the Storybook components
- Locally, create some BulkUpload history in your database (apologies, uploading itself will be in a forthcoming PR)
- Make sure the sidebar displays them correctly!

Note that the uploader itself is not included in this PR, and the styling/behavior of the components does not yet include the changes suggested in #477.

Closes #413
Closes #414
Closes #492


reefdog:
@jasonaowen I’ve made some further refinements to these components but am having trouble getting my laptop on to the Istanbul airport WiFi, so I’ll have to wait to push them until I’m at my final spot late Thursday night your time.

jasonaowen:
@reefdog how's your internet connection looking? Should I hold off on this or go with what's here?

src/interfaces/BulkUpload.ts
```diff
@@ -0,0 +1,7 @@
+export interface BulkUpload {
+  id: string;
+  fileName: string;
+  fileSize: string;
```
jasonaowen, line 4: Hm, it looks like we are not storing the file size in the database at the moment, so the API is not returning it. That seems like something we could reasonably cache, but maybe not in this first iteration? cc @slifty

src/interfaces/BulkUpload.ts
```diff
@@ -0,0 +1,7 @@
+export interface BulkUpload {
```
jasonaowen, line 1: I think I'm going to move this into `src/pdc-api.ts` as `ApiBulkUpload`.

src/components/BulkUploadListItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const makeMetadata = () => {
+    const metadata = [];
+
+    if (isProcessing) {
+      metadata.push(createdAt, fileSize, 'Processing…');
+    }
+
+    if (isComplete) {
+      metadata.push(createdAt, fileSize);
```
jasonaowen, line 53: @slifty this makes me think we want the `bulk_uploads` table to have a `finished_at` or so!

src/components/BulkUploadListItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+interface BulkUploadListItemProps {
+  upload: BulkUpload;
+}
+
+export const BulkUploadListItem = ({ upload }: BulkUploadListItemProps) => {
+  const {
+    fileName, fileSize, status, createdAt,
+  } = upload;
+
+  // TODO: Normalize these against the statuses returned by the API.
```
jasonaowen, line 16: Here's the full list from https://github.com/PhilanthropyDataCommons/service/blob/eeb035f4a71cedf7fdeb5226b770349df853272a/src/types/BulkUpload.ts#L5-L9 :

- `pending`
- `in_progress`
- `completed`
- `failed`
- `canceled`

The `canceled` status is planned but not implemented, and the `pending` and `in_progress` can probably both be considered `processing` for the sake of the UI.

src/components/BulkUploadListItem.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const isComplete = status === 'complete';
+  const hasError = status === 'error';
+
+  const makeClassNames = () => {
+    const classNames = ['bulk-upload-list-item'];
+
+    classNames.push(status);
+
+    return classNames.join(' ');
+  };
```
jasonaowen, line 27: I'm probably going to inline this to `['bulk-upload-list-item', status].join(' ')` or so.

src/components/BulkUploadListItem.tsx
reefdog: Ah, that brings up a good note… I just YOLO'ed `createdAt` not knowing what the actual best timestamp attribute would be and meant to call attention to it. Thanks for noticing, and yes, any more appropriate stamp is fine.

I do think if we only show one timestamp, reporting to the user when _they_ took the action is preferable. But there are other good reasons for showing `finished_at`, including (probably) closer sync with the `created_at` timestamps of individual rows. In any case I doubt the differences will be material and am not at all insistent on one way or the other.

src/components/BulkUploadListItem.tsx
reefdog: Yeah, I was considering doing the same, but wasn't sure how the status normalizing would work out / if we'd need to do more `switch`ing here. Feel free to do that.

src/components/BulkUploadListItem.tsx
reefdog: Oh excellent, I almost guessed correctly! Thanks for surfacing those.

1. Is `canceled` planned for MVP? If so, I'll add styling for it. What does it mean? cc @slifty 
2. You're right that from our end user perspective, `pending` and `in_progress` are the same. (At least, going on my assumption about what they mean. Pending is received and in the queue, but not yet started; in progress has actually been started?)

reefdog: I'm gonna update based on this list and push again.

src/components/BulkUploadListItem.tsx
reefdog: Oh, wait. I just realized the "Failed" status is written as if it's reporting when the failure occurred. Hmm.

Okay I've thought myself into thinking that for pending/in_progress we should report the start time (or something close to upload time) and for the other statuses we should report the end time (or something close). I don't think this precision is vital to the UX, but if it's easy enough to accomplish, then it'd be good.

In any case, I don't want to the blocker on that, so go with whatever you and @slifty think makes sense!

reefdog:
@jasonaowen Thanks for the review notes! I just pushed my earlier changes from the airport (sorry!) and a few new ones based on your comments above, including the correct status values. I'll be offline for a while again, so please have full freedom over this!

src/interfaces/BulkUpload.ts
slifty: Got it -- I'd vote we make an issue for it so we can implement it the Right Way after the demo.

src/components/BulkUploadListItem.tsx
slifty: For the immediate term if we could work with just the `createdAt` value that is preferred, I agree we should aspire to include more information before we consider the feature complete.  (this also gives us enough time to think about whether it would be better to have that more robust "status reporting" / "job history" feature which would provide the information in a more flexible / granular way along with the other desired metadata)

src/components/BulkUploadListItem.tsx
slifty: Just confirmation that we can skip canceled (in future the front end could add a "cancel" button for items that are pending, if you wanted.)

src/interfaces/BulkUpload.ts
jasonaowen: https://github.com/PhilanthropyDataCommons/service/issues/651

src/components/BulkUploadListItem.tsx
jasonaowen: https://github.com/PhilanthropyDataCommons/service/issues/652

src/components/BulkUploadListItem.tsx
jasonaowen: > Pending is received and in the queue, but not yet started; in progress has actually been started

Correct!

jasonaowen:
Rebased onto `main` after #432 and to prompt a deploy preview now that the base branch is `main`.

jasonaowen:
Looking good! I still have a little bit left to clean up, but it's working locally:

Screen Shot 2023-12-04 at 16 10 28

reefdog:
Huzzah! Clearly I'll need to do a further styling pass for situations like the wrapping (where the bullet delimiter appears at the beginning of the last line, not the end of the previous), and getting the spinner to be the same size as the icon, etc. But this is great!

slifty:
I wanna give a plug for differentiating between "IN_PROGRESS" and "PENDING"

Something might be pending for a long time and that means something different from being in progress for a long time.  I think it would be helpful in debugging / user stories for there to be a distinction.

reefdog:
I would describe them in the UI as “Waiting to be processed” and then “Processing”. (I think you picked fine names for the backend — better for that purpose, even — but they don’t translate intuitively IMO.)

I don’t mind, though I’d probably style them exactly the same and just change the bit of text in that metadata section.

reefdog:
@slifty As is typical, I think I have an even **better** idea! Forthcoming…

reefdog:
@slifty Take a look! https://deploy-preview-443--philanthropy-data-commons-viewer.netlify.app/storybook/?path=/docs/stories-bulkuploadlistitem--docs#default

slifty:
@reefdog wow! It's perfect! And to think it's all thanks to me!

(this looks great -- I have no notes)

reefdog:
@jasonaowen Relative time added to bulk upload history components. Should be visible in the Storybook deploy preview shortly. With that, I hand this back to you!

reefdog:
@slifty This PR is ready for review! Very sorry that it basically includes bulk upload _history_ without actually supporting the uploading. 😬 So you'll have to review code + Storybook components, but if you want to see it in practice you'll need to locally stuff some bulk upload history into your database manually.

src/components/BulkUploadListItem.tsx
```diff
@@ -22,112 +22,95 @@ export const BulkUploadListItem = ({ upload }: BulkUploadListItemProps) => {
...
+          Email the dev team for help
+        </EmailLink>
+      )
+    );
+  }
 
-    return null;
-  };
+  metadata = metadata.filter((datum) => datum).map((datum, i) => (
+    // eslint-disable-next-line react/no-array-index-key
```
slifty, line 98: Another instance of array index being used as a key; does `datum` have any type of identifier?

src/utils/files.ts
```diff
@@ -0,0 +1,33 @@
+/**
+   * Convert byte counts to larger units like KB or MB.
+   *
+   * Includes one trailing digit (when needed).
+   *
+   * @param  {bytes} number
+   * @return {String} Bytes formatted to nearest reasonable number
+   */
+const humanizeByteCount = (bytes: number) => {
```
slifty, line 9: It might be good to introduce this method in the same commit it is used (as written this commit just adds dead code, although of course I assume it's about to get used somewhere)

slifty: Actually -- rather than write this ourselves, what do you think of using something like https://www.npmjs.com/package/pretty-bytes (11m weekly downloads)

src/components/BulkUploadListItem.tsx
```diff
@@ -19,7 +20,7 @@ interface BulkUploadListItemProps {
 
 export const BulkUploadListItem = ({ upload }: BulkUploadListItemProps) => {
   const {
-    id, fileName, status, createdAt,
+    id, fileName, fileSize, status, createdAt,
```
slifty, line 23: w00t file size is back!

Rather than having all these commits with thrash would it be easy to reorder / smoosh / remove things so that it's just introduced once?

slifty (changes requested):
Just did a first pass code review.  Happy to talk through any of this / not trying to blow up progress with some of the more nit picky comments.

src/components/BulkUploadListItem.tsx
reefdog: It does not, unfortunately. I could probably refactor it to have that by making `metadata` an object instead of an array, so we can use the object keys as part of the key construction… I'll try that!

src/pages/AddData.tsx
```diff
@@ -1,18 +1,24 @@
...
+import { BulkUploadList } from '../components/BulkUploadList';
 
-const BulkUploaderLoader = () => {
-  const history: object[] = []; // todo
-  return (
-    <PanelGridItem>
-      <BulkUploader history={history} loading />
-    </PanelGridItem>
-  );
+const BulkUploaderLoader = () => <p>TODO: bulk uploader</p>;
```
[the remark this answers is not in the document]
reefdog: I hear you, but this is acceptable in the history of the project. This is just moving a string that already existed in the `BulkUploader` component — which we are deleting — inline to this template. Jason's uploader work, which will be the next PR that I open, will replace it with the real uploader.

src/utils/datetime.ts
```diff
@@ -0,0 +1,35 @@
...
+   * but accepts an optional second parameter to override. See:
+   * https://day.js.org/docs/en/display/format#localized-formats
+   *
+   * @param  {timestamp} string Any datetime that DayJS understands
+   * @param  {format} string DayJS format code (see docs)
+   * @return {String} Local version of datetime in our preferred format
+   */
+const localizeDateTime = (timestamp: string, format = 'LLL') => dayjs(timestamp).format(format);
+
+export {
```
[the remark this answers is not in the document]
reefdog: I sometimes (though rarely) have utility functions (or other variable creation) inside these utility files that _aren't_ meant for export, just some internal syntactic sugar. So an advantage of the manual export at the end is a single manifest that lets the developer eyeball all the exports from a file. Just IMO.

src/components/BulkUploadList.tsx
```diff
@@ -0,0 +1,21 @@
...
+  uploads: BulkUpload[];
+}
+
+export const BulkUploadList = ({
+  uploads,
+}: BulkUploadListProps) => (
+  <div className="bulk-upload-list">
+    {uploads.map((upload, index) => (
+      <BulkUploadListItem
+        key={index} // eslint-disable-line react/no-array-index-key
```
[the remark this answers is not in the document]
reefdog: Done by a following commit, but history has now been rewritten regardless so this was never even there! 💨

src/components/BulkUploadListItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+  const makeHumanizedFileSize = () => {
+    // TODO: Humanize the file size as bytes, KB, MB, or GB.
+    //       Round KB to zero decimals, MB to one decimal, and GB to two decimals.
+    const humanizedFileSize = fileSize;
+
+    return humanizedFileSize;
+  };
+
+  const makeMetadata = () => {
+    const humanizedTimestamp = makeHumanizedTime();
```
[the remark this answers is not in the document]
reefdog: > I just realized these utilities don't take inputs / are not true functions.
>
> Can we adjust this and the FileSize function so they take input, humanize, and return the humanized output?

Refactor has removed the "not a true function" bit; they're both now just variable assignments (and any prerequisite do functions take input).

> Also, what do you think of removing humanized from the variable names (I'm not feeling strongly on this, but the prefix smelled a little of hungarian notation on first read)

If I removed the `humanized` prefix, `timestamp` will be fine but `fileSize` is already assigned from the prop destructuring, so we'd need to find some other way of indicating it. I could change the `humanized` prefix to something like `pretty` or `formatted`, but those feel a little like just changing the lipstick on the pig if you're hoping to fully avoid Hungarian.

src/components/BulkUploadListItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+          {humanizedTimestamp}
+          {METADATA_DELIMITER}
+          {humanizedFileSize}
+          {METADATA_DELIMITER}
+          Waiting to process…
+        </>
+      );
+    }
+
+    if (isInProgress) {
```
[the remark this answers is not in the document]
reefdog: I think that's a good idea! We iterated here from a previous version where there was less switch-able logic going on, but with where we've landed, a switch seems best.

src/utils/datetime.ts
```diff
@@ -0,0 +1,35 @@
+import dayjs from 'dayjs';
```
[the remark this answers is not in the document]
reefdog: I actually thought best practice was the opposite: add it in its own commit, but bundle them together in a PR. But after further thought I'm not sure why that really helps, and also it's more difficult to reconstruct history from the basic git log/blame tools.

So, wish granted!

src/components/BulkUploadListItem.tsx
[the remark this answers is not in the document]
slifty: @reefdog I'm OK with it as is, given the context!

src/utils/datetime.ts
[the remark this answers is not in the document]
slifty: I probably asked you to do it the other way in a past project; oh to have been so young and naive.

src/components/BulkUploadListItem.tsx
```diff
@@ -0,0 +1,140 @@
...
+interface BulkUploadListItemProps {
+  upload: BulkUpload;
+}
+
+export const BulkUploadListItem = ({ upload }: BulkUploadListItemProps) => {
+  const {
+    fileName, fileSize, status, createdAt,
+  } = upload;
+
+  // Don't let the bullet wrap to the beginning of a new line.
```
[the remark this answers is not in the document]
reefdog: Will do!

reefdog:
@slifty Success! I've grouped the commits more sensibly (thank you for that prompt) and addressed what I believe were all your concerns. Can you give it another review? (It won't let me re-request one from you. 😵.)

src/components/BulkUploadListItem.tsx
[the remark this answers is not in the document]
reefdog: @slifty I at least changed `humanized*` (which is weird, subjective) to `relativeTimestamp` and `formattedFileSize`.

src/utils/datetime.ts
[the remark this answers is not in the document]
reefdog: Probably, it does sound like you. With your endless requests and rowing.

package.json
```diff
@@ -6,8 +6,10 @@
     "@axa-fr/react-oidc": "^7.13.6",
     "@fontsource/source-sans-pro": "^5.0.8",
     "@heroicons/react": "^2.0.18",
+    "dayjs": "^1.11.10",
```
slifty, line 9: Could you put this in its own commit?

src/components/BulkUploadListItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+import {
+  localizeDateTime,
+  relativizeDateTime,
+} from '../utils/datetime';
+import type { ApiBulkUpload } from '../pdc-api';
+import { EmailLink } from './EmailLink';
+import './BulkUploadListItem.css';
+
+enum BulkUploadStatuses {
+  pending = 'pending',
```
slifty, line 18: Huh, I'm used to enum names (not values) being ALL_CAPS but I suppose if the linter doesn't care then why should we!  This is not a change request per se, just a musing.

src/components/BulkUploadListItem.tsx
```diff
@@ -0,0 +1,127 @@
...
+          >
+            Email the dev team for help
+          </EmailLink>
+        )
+      );
+      break;
+    default:
+  }
+
+  // We delimit metadata elements with a bullet padded by two spaces.
```
slifty, line 99: this raises a deep philosophical question: if something has one space on either side of it, is it padded by two spaces or is it padded by one space.  If two: can something be padded by one space?  If one: are we saying the spaces are the same space?

slifty (approved):
HOORAY!

I left you some really useful final feedback.

(please smash the merge button at your leisure)

package.json
reefdog: `:twitch:`

src/components/BulkUploadListItem.tsx
reefdog: Ooh, actually, I do like that. Linter doesn't care but I'll change it anyway.

src/components/BulkUploadListItem.tsx
reefdog: Superficially, you're right, and I can illustrate it with a CSS example: we'd never say "padded by 20px" to mean the sum of padding. We would mean 20px in each relevant dimension. E.g., a box with 5px of padding on each side would be "padded by 5px", not padded by 20px. (If it were padded in only one dimension, say, horizontally but not vertically, then either context or explanation would make that clear.)

However, this is a _bullet_, which has only one external surface. Therefore to describe its padding, we _do_ sum the values of "both sides" — that is, spacing off any axis. And so one space on either "side" is really two spaces distributed symmetrically off the horizontal poles of the unisurface. I thought it easier to say "padded by two spaces".

Thank you for coming to my TED talk.

reefdog: That was offensively clear. Or rather, so clear that your explanation is offensive to me.

PR #449, opened 2023-12-01
Add data instructions and base fields table

### LINKED ISSUE #417: Add bulk upload instructional copy and link CSV template
Once instructional copy has been approved (#416), populate the interface with it! This will also include a local link to the CSV template, which I _assume_ will be a statically-hosted copy of the [service repo's canonical version](https://github.com/PhilanthropyDataCommons/service/blob/main/src/public/bulkUpload.csv). (If we can point directly to that one… even better, maybe?)

### LINKED ISSUE #418: Add base field description table to bulk upload instructions
Once base field descriptions are written (https://github.com/PhilanthropyDataCommons/service/issues/613) and loaded (https://github.com/PhilanthropyDataCommons/service/issues/614), we will need to generate a table with them in the bulk upload instruction panel:

Instruction Panel

Wrapping should be enabled by default, with no need for a toggler.

I'm assigning this to @jasonaowen and @reefdog to tag-team; it is likely mostly @jasonaowen but may require some UI stubbing by @reefdog to be sure it behaves as desired (full panel scrolling, maybe full bleed table, fixed header, etc.; all to be determined during implementation playing).

jasonaowen:
Add the text that @reefdog wrote describing the use of the add data page, and show the list of base fields.

Screenshot of add data page with instructions at the top and the list of base fields below it and extending off the bottom

Unfortunately, we have some very long base field `shortCode`s, so the first column is wider than in the mockup in #418. As a follow-up, we may wish to introduce a truncation toggle as we did on the proposals list. Additionally, or possibly alternatively, we are meaning to do a base field review, so maybe we'll fix this display challenge by fixing the data.

Note that this code assumes the API is giving us descriptions, but that change has not yet been deployed to production. That works out because a missing key lookup evaluates to null, which is rendered just the same as the empty string the API would give us until descriptions have been written.

To review, manually edit your URL to `/add-data`, or visit the [add data page on the deploy preview](https://deploy-preview-449--philanthropy-data-commons-viewer.netlify.app/add-data); I've updated Keycloak to allow logging in from the deploy preview.

I believe this should not conflict with #443, but if it does I'll resolve any merge conflicts on whichever PR is merged second.

Resolves #417 
Resolves #418

### DIFF
--- a/src/components/AddDataInstructions.css
+++ b/src/components/AddDataInstructions.css
@@ -1,0 +1,3 @@
+.add-data-instructions {
+  max-width: 80ch;
+}
--- a/src/components/AddDataInstructions.tsx
+++ b/src/components/AddDataInstructions.tsx
@@ -2,6 +2,9 @@ import React from 'react';
 import './AddDataInstructions.css';
 import { Panel, PanelBody } from './Panel';
 
+const apiUrl = new URL('/', process.env.REACT_APP_API_URL);
+const bulkUploadTemplateUrl = new URL('/static/bulkUpload.csv', apiUrl);
+
 interface AddDataInstructionsProps {
   children?: React.ReactNode;
 }
@@ -11,7 +14,54 @@ export const AddDataInstructions = ({
 }: AddDataInstructionsProps) => (
   <Panel>
     <PanelBody>
-      <p>TODO: add data instructions</p>
+      <article className="add-data-instructions">
+        <h3>Add data to the PDC</h3>
+
+        <p>
+          <b>Adding data via the bulk uploader</b>
+          {' '}
+          is drag-and-drop easy:
+        </p>
+
+        <ol>
+          <li>
+            Create a CSV of your data with base field keys as the first row
+            (see table below). If you want, you can start with
+            {' '}
+            <a href={bulkUploadTemplateUrl.toString()}>
+              this template of recommended base fields
+            </a>
+            .
+          </li>
+          <li>
+            Upload your CSV in the sidebar to the left.
+          </li>
+        </ol>
+
+        <p>
+          Note that you cannot create new base fields via the bulk uploader.
+          Any columns in your file that don’t match an existing base field will
+          cause an error.
+        </p>
+
+        <p>
+          <b>Adding data via the API</b>
+          {' '}
+          allows the most control. You can create new opportunities,
+          add applications, register new base fields, and more.
+          To use the API, request an API key from
+          {' '}
+          <a href="mailto:info@philanthropydatacommons.org?Subject=PDC%20API%20key%20request">
+            info@philanthropydatacommons.org
+          </a>
+          {' '}
+          and then visit the documentation site at
+          {' '}
+          <a href={apiUrl.toString()}>{apiUrl.hostname}</a>
+          .
+        </p>
+      </article>
+
       {children}
     </PanelBody>
   </Panel>
--- a/src/components/AddDataInstructions.tsx
+++ b/src/components/AddDataInstructions.tsx
@@ -62,7 +62,21 @@ export const AddDataInstructions = ({
         </p>
       </article>
 
-      {children}
+      <article>
+        <header className="add-data-instructions">
+          <h3>Base Fields</h3>
+
+          <p>
+            Base fields are the backbone of the PDC. Your data must write to an
+            existing base field, unless you create a new one using the
+            {' '}
+            <a href={apiUrl.toString()}>API</a>
+            .
+          </p>
+        </header>
+
+        {children}
+      </article>
     </PanelBody>
   </Panel>
 );
--- a/src/components/BaseFields.tsx
+++ b/src/components/BaseFields.tsx
@@ -1,18 +1,59 @@
 import React from 'react';
 import './BaseFields.css';
 import { ApiBaseField } from '../pdc-api';
+import {
+  Table,
+  TableHead,
+  ColumnHead,
+  TableBody,
+  TableRow,
+  RowCell,
+} from './Table';
+
+interface BaseFieldListTableRowProps {
+  field: ApiBaseField;
+}
+
+const BaseFieldListTableRow = ({
+  field,
+}: BaseFieldListTableRowProps) => (
+  <TableRow>
+    <RowCell>
+      <code className="base-field-short-code">
+        {field.shortCode}
+      </code>
+    </RowCell>
+    <RowCell>
+      {field.label}
+    </RowCell>
+    <RowCell>
+      {field.description}
+    </RowCell>
+  </TableRow>
+);
 
 interface BaseFieldsProps {
   fields: ApiBaseField[];
-  loading?: boolean;
 }
 
-/* eslint-disable @typescript-eslint/no-unused-vars, arrow-body-style */
 export const BaseFields = ({
   fields,
-  loading = false,
-}: BaseFieldsProps) => {
-  return (
-    <p>TODO: base fields list</p>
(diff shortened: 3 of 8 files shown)


src/components/BaseFields.tsx
```diff
@@ -1,18 +1,59 @@
...
+interface BaseFieldListTableRowProps {
+  field: ApiBaseField;
+}
+
+const BaseFieldListTableRow = ({
+  field,
+}: BaseFieldListTableRowProps) => (
+  <TableRow>
+    <RowCell>
+      <code>
```
reefdog, line 22: Oof, those long shortCodes are a nightmare (and an oxymoron!).

We could add the Toggle Wrap button, but I'm also okay with Just Wrapping Them. Unfortunately I'm bumping up against table-column-calculation weirdness and so I think we'll have to be a bit heavy-handed with it. Here's what I suggest:

```suggestion
      <code className="base-field-short-code">
```

Then, to `BaseFields.css`, add:

```css
.base-field-short-code {
  display: block;
  max-width: 50ch;
  word-wrap: break-word;
}
```

reefdog (dismissed):
Yay! Fantastic. I definitely see some panel improvements I'll want to make when I get back, but I think this is a good enough MVP.

I left a few inline thoughts, but once those are resolved feel free to merge.

src/components/AddDataInstructions.tsx
```diff
@@ -11,8 +14,61 @@ export const AddDataInstructions = ({
 }: AddDataInstructionsProps) => (
   <Panel>
     <PanelBody>
-      <p>TODO: add data instructions</p>
-      {children}
+      <article className="add-data-instructions">
+        <h3>Add data to the PDC</h3>
+
+        <p>Adding data via the bulk uploader is drag-and-drop easy:</p>
```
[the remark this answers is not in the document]
slifty: @reefdog are `<strong>` and `<em>` considered just as bad as `<b>`? I had assumed that the issue with `<b>` was really more that "b" implies a visual style whereas strong / em semantically mean emphasized.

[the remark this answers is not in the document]
reefdog: You're correct, and that's exactly why I'm using it here! It's a visual emphasis for guiding the eye, like chapter headings. If it were being read aloud, I wouldn't want there to be any strong/emphasis on the words, so I'm intentionally avoiding those semantic tags.

Probably the right thing to do here would be wrap each section in `<section>` and indicate the parts in `<b>` are "headings" (without using block-level `h#` tags) with `aria-role` attributes or something… but I'd need to dig more into that than I can at the moment.

src/components/AddDataInstructions.css
```diff
@@ -0,0 +1,3 @@
+.add-data-instructions {
+  max-width: 40em;
```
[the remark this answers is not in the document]
jasonaowen: TIL!

src/components/AddDataInstructions.tsx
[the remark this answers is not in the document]
jasonaowen: Oops, yes, you had that in the wikitext and I lost it when I copied and pasted. Sorry about that!

src/components/BaseFields.tsx
jasonaowen: Nice solution! The wrapped short codes still copy fine, although selecting them is more difficult. I'm kinda okay with this subtle discouragement to using such long base fields, because that length seems like a bit of a red flag; the longer the "short" code for a base field is, the more likely we want to review it.

src/components/AddDataInstructions.tsx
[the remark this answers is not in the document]
slifty: @reefdog ohhh that makes sense (also I'm now seeing I had misread your original post and DID want screen readers to see it that way) -- thanks for spending the moment to enlighten me!

src/components/AddDataInstructions.tsx
```diff
@@ -11,8 +14,69 @@ export const AddDataInstructions = ({
...
+
+        <p>
+          <b>Adding data via the bulk uploader</b>
+          {' '}
+          is drag-and-drop easy:
+        </p>
+
+        <ol>
+          <li>
+            Create a CSV of your data with base field keys as the first row
```
jasonaowen, line 28: After talking with @slifty about implementing the CSV parser, it sounds like there are some additional requirements here: the first two fields _must_ be `proposal_external_id` and `applicant_external_id`, and those are not base fields. That unfortunately complicates the instructions. I'm okay with deferring those instruction changes for now, but we'll need to follow up on that.

slifty: +1 on the idea of not changing this PR based on that just yet; I don't know exactly what it's going to look like yet so we should wait until the PR is at least opened.

reefdog: Oof… I trust you guys, but to pretend I don't know how computers work and am just a "make UX nice" guy:

1. The positional requirement, while easiest for a user to grok/resolve, is weird to me. There are CSV parsers that create objects for each row, keyed by column name, rather than positional arrays; are we avoiding that for memory consumption reasons or a tooling preference or…?
2. I think saying "must be base fields PLUS these two fields" is fine, but are the values of those fields something they get to make up, or do they have to reference correct values from our database (like a foreign key reference or something)? If the former, fine. We can adapt the instructions I'm sure. But if the latter, that's a real wrinkle in the UX and cuts down the audience for this tool dramatically.

jasonaowen: > The positional requirement, while easiest for a user to grok/resolve, is weird to me.

As @slifty said, we're not exactly sure what it's going to look like yet, so what I wrote is what I imagined from our conversation. We could definitely make it so those special fields are allowed anywhere; I thought requiring them at the beginning would be easier to understand and easier to explain, but I don't think there are any technical requirements there, and I'm happy to defer to your UX expertise on that!

> are the values of those fields something they get to make up?

Yes! The applicants will be upserted (created or reused) based on the external ID; the proposal external ID must be unique within the CSV.

(In more detail: there is a `UNIQUE` constraint on `proposals (applicant_id, opportunity_id, external_id)`. I believe we'll be creating a new `opportunity_id` for each CSV, and applicants are upserted, so all that's left to satisfy that constraint is making sure there aren't two proposals with the same `external_id`.)

src/components/AddDataInstructions.tsx
```diff
@@ -11,8 +14,69 @@ export const AddDataInstructions = ({
...
+            </a>
+            .
+          </li>
+          <li>
+            Upload your CSV in the sidebar to the left.
+          </li>
+        </ol>
+
+        <p>
+          Note that you cannot create new base fields via the bulk uploader.
```
slifty, line 42: In the spirit of "humans are better at being told what to do than what not to do" I wonder if this would be better written as "you may only use the base fields listed below"

The idea of new base fields may or may not actually be on the mind of the user, and I think answering that might be cleaner as a totally independent use case some day in the future (and in either event, this section is not addressing that use case / explaining how to do it!)

slifty (approved):
Wordsmithing note -- feel free to ignore.

I was able to see this all working on the test deployment!

src/components/AddDataInstructions.tsx
jasonaowen: That makes sense to me, and I agree with it - good catch! Unfortunately, I think for now I'm going to merge as-is and follow up with @reefdog when he's back.

reefdog: Good thought. Opened as #464.

PR #438, opened 2023-11-29
Add link to Add Data page to navbar

reefdog:
This PR adds a link to the Add Data page to the application navbar, as well as the RECENT_CHANGES note about bulk upload.

Closes #410

PR #519, opened 2024-01-24
Upload bulk uploads to the API

### LINKED ISSUE #412: Upload file to service via file-uploader
After the components are available in #411, we need to actually wire up file uploads. This involves multiple API calls, but to the user it's a single front-end operation.

- [x] Generate pre-signed POST
- [x] Upload file to the storage service
- [x] Register bulk upload with the API

@jasonaowen feel free to split this issue into sub-issues if you want to tackle it that way!

### LINKED ISSUE #528: Handle pre-upload BulkUploadFilePicker errors
#519 adds basic bulk uploads, but it's lacking error handling when the upload fails during initial registration with the API (or any other error prior to the API receiving the upload).

In this case, we should set the `status` state of BulkUploadFilePicker to `error.{someErrorType}` and add the appropriate case and content to [this render block](https://github.com/PhilanthropyDataCommons/front-end/blob/d5529b4bcbc4c36dfaf3719c794030995e1827bf/src/components/BulkUploaderFilePicker.tsx#L275-L315).

reefdog:
This PR:
- Adds a file handler to the file picker component
- Creates a presigned POST request
- Uploads the bulk upload to the service
- Marks the upload as complete once it is
- Marks the file picker with an error if the upload fails

It does not:

- Refresh the BulkUpload history with the new upload; user must refresh.
  - Opened as: #527

TODO:
- [ ] Sort out inconsistent upload issue
- ~Add RECENT_CHANGES note~ Edit: Actually, we shouldn't do this until we actually want users to use it, I.e. #438

🚨 **Reminder:** The `service` repo will need a production deployment before this is merged in. The [currently-deployed API](https://api.philanthropydatacommons.org) doesn't have `/bulkUploads` routes. 🚨

Closes #412
Closes #528

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
--- a/src/components/BulkUploaderFilePicker.tsx
+++ b/src/components/BulkUploaderFilePicker.tsx
@@ -49,7 +49,13 @@ const FilePickerMessage = ({
   </div>
 );
 
-export const BulkUploaderFilePicker = () => {
+interface BulkUploaderFilePickerProps {
+  uploadFile: (file: File) => Promise<void>;
+}
+
+export const BulkUploaderFilePicker = ({
+  uploadFile,
+}: BulkUploaderFilePickerProps) => {
   const ACCEPTED_MIME_TYPES = ['text/csv'];
   const ACCEPTED_FILE_EXTENSIONS = ['csv'];
   const MAXIMUM_FILE_SIZE_BYTES = 1e+9; // 1GB
@@ -74,7 +80,7 @@ export const BulkUploaderFilePicker = () => {
    *
    * @param  {FileList | null} files
    */
-  const validateFiles = (files: FileList | null) => {
+  const validateFiles = (files: FileList | null): files is FileList & { 0: File } => {
     if (!files?.[0]) {
       setStatus('error.noFile');
       return false;
@@ -105,10 +111,8 @@ export const BulkUploaderFilePicker = () => {
   };
 
   /**
-   * Provided a list of files
-   * from either a native file picker
-   * or the Drag and Drop API,
-   * validates and uploads the files.
+   * Provided a list of files from either a native file picker
+   * or the Drag and Drop API, validates and uploads the files.
    *
    * @param  {FileList | null} files
    */
@@ -118,13 +122,18 @@ export const BulkUploaderFilePicker = () => {
     }
 
     // Remove focus from the button.
-    // We wait until after validation
-    // so focus is not wrenched away.
+    // We wait until after validation so focus is not wrenched away.
     buttonRef.current?.blur();
 
     setStatus('uploading');
 
-    // TODO: Begin uploading
+    uploadFile(files[0])
+      .then(() => {
+        setStatus('complete');
+      })
+      .catch(() => {
+        setStatus('error.uploadFailure');
+      });
   };
 
   /**
@@ -302,12 +311,12 @@ export const BulkUploaderFilePicker = () => {
               resetLink
             />
           );
-        case 'generic':
+        case 'uploadFailure':
         default:
           return (
             <FilePickerMessage
               icon={<XCircleIcon className="icon" />}
-              text="Error uploading your file"
+              text="Error uploading your file. Please try again."
               resetLink
             />
           );
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import { useCallback, useEffect, useState } from 'react';
 import { useOidcFetch } from '@axa-fr/react-oidc';
 import { getLogger } from './logger';
 
@@ -6,14 +6,14 @@ const logger = getLogger('pdc-api');
 
 const API_URL = process.env.REACT_APP_API_URL;
 
-const throwNotOk = (res: Response): Response => {
+const throwIfResponseIsNotOk = (res: Response): Response => {
   if (!res.ok) {
     throw new Error(`Response status: ${res.status} ${res.statusText}`);
   }
   return res;
 };
 
-const logError = (error: unknown, path: string, params: URLSearchParams) => {
+const logError = (error: unknown, path: string, params: object) => {
   logger.error({ error, params }, `Error fetching ${path}`);
 };
 
@@ -30,7 +30,7 @@ const usePdcApi = <T>(
     const url = new URL(path, API_URL);
     url.search = params.toString();
     fetch(url)
-      .then(throwNotOk)
+      .then(throwIfResponseIsNotOk)
       .then((res) => res.json())
       .then(setResponse)
       .catch((e) => logError(e, path, params));
@@ -48,6 +48,103 @@ const usePdcApi = <T>(
   }, [path, params.toString()]);
 
   return response;
+};
+
+const usePdcCallbackApi = <T>(
+  path: string,
+): (options: RequestInit) => Promise<T> => {
+  const { fetch } = useOidcFetch();
+
+  return useCallback((options: RequestInit) => {
+    const url = new URL(path, API_URL);
+    return fetch(url, options)
+      .then(throwIfResponseIsNotOk)
+      .then((res) => res.json())
+      .catch((e) => logError(e, path, options));
+
+    /* eslint-disable-next-line react-hooks/exhaustive-deps --
+     *
+     * fetch should not be a dependency, because although it or its internal
+     * state may change from render to render, such changes are not relevant:
+     * a change in the way we make a request should not trigger an API request
+     */
+  }, [path]);
+};
+
+interface ApiPresignedPostRequest {
+  fileType: string;
+  fileSize: number;
+}
+
+interface PresignedPost {
+  url: string;
+  fields: Record<string, string> & {
+    key: string,
+  };
+}
+
+interface ApiPresignedPostResponse {
+  fileType: string;
+  fileSize: number;
+  presignedPost: PresignedPost;
+}
+
+const usePresignedPostCallback = () => {
+  const api = usePdcCallbackApi<ApiPresignedPostResponse>('/presignedPostRequests');
+  return (params: ApiPresignedPostRequest) => api({
+    method: 'post',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(params),
+  });
+};
+
+const uploadUsingPresignedPost = async (
+  file: File,
+  presignedPost: PresignedPost,
+) => {
+  const formData = new FormData();
+  Object.entries(presignedPost.fields).forEach(([key, value]) => (
+    formData.append(key, value)
+  ));
+  formData.append('Content-Type', file.type || 'application/octet-stream');
+  formData.append('file', file);
+  return fetch(
+    presignedPost.url,
+    {
+      method: 'post',
+      body: formData,
+    },
+  )
+    .then(throwIfResponseIsNotOk);
+};
+
+interface ApiBulkUploadsPostRequest {
+  fileName: string;
+  sourceKey: string;
+}
+
+interface ApiBulkUploadsPostResponse {
+  id: number;
+  fileName: string;
+  fileSize: number;
+  sourceKey: string;
+  status: string;
+  createdAt: string;
+}
+
+const useRegisterBulkUploadCallback = () => {
+  const api = usePdcCallbackApi<ApiBulkUploadsPostResponse>('/bulkUploads');
+  return (params: ApiBulkUploadsPostRequest) => api({
+    method: 'post',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(params),
+  });
 };
 
 interface ApiBaseField {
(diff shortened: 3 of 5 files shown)


src/components/BulkUploaderFilePicker.tsx
```diff
@@ -124,7 +130,9 @@ export const BulkUploaderFilePicker = () => {
 
     setStatus('uploading');
 
-    // TODO: Begin uploading
+    uploadFile(files[0]!);
```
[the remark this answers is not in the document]
slifty: OK I understand now and see in the validator we do NOT allow multiple files (and error).

I'm going to see if there's a way for the validator to signal that `files` is a FileList of size 1, specifically, which would obviate the need for a `!`

slifty:
@reefdog this is your PR so I can't request your review

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+end_of_line = lf
+indent_size = 2
+indent_style = space
+insert_final_newline = true
+trim_trailing_whitespace = true
```
reefdog, line 9: Prefer that this would have been in a separate PR, but, no harm no foul!

src/components/BulkUploaderFilePicker.tsx
```diff
@@ -105,10 +111,8 @@ export const BulkUploaderFilePicker = () => {
   };
 
   /**
-   * Provided a list of files
-   * from either a native file picker
-   * or the Drag and Drop API,
-   * validates and uploads the files.
+   * Provided a list of files from either a native file picker
```
slifty, line 114: Whoops I didn't realize this changed the original file (but like, also, these are not README lines and so the concept of breaking so frequently is actually a bad UX!  I imagine prettier will update these in future so I'm happy to undo them for this PR)

src/pdc-api.ts
```diff
@@ -50,6 +50,101 @@ const usePdcApi = <T>(
...
+
+interface ApiBulkUploadsPostRequest {
+  fileName: string;
+  sourceKey: string;
+}
+
+interface ApiBulkUploadsPostResponse {
+  id: number;
+  fileName: string;
+  sourceKey: string;
```
reefdog, line 131: This interface initially existed prior to `fileSize` being added to the BulkUpload. Can you add it here, now that it's part of the response?

src/pdc-api.ts
```diff
@@ -50,6 +50,101 @@ const usePdcApi = <T>(
...
 
+const usePdcCallbackApi = <T>(
+  path: string,
+): (options: RequestInit) => Promise<T> => {
+  const { fetch } = useOidcFetch();
+
+  return useCallback((options: RequestInit) => {
+    const url = new URL(path, API_URL);
+    return fetch(url, options)
+      .then(throwNotOk)
```
reefdog, line 61: Wait why is this throwing not OK all the time?

src/pdc-api.ts
```diff
@@ -50,6 +50,101 @@ const usePdcApi = <T>(
...
+  formData.append('Content-Type', post.fileType);
+  formData.append('file', file);
+  return fetch(
+    post.presignedPost.url,
+    {
+      method: 'post',
+      body: formData,
+    },
+  )
+    .then(throwNotOk)
```
reefdog, line 119: Same here? Why `throwNotOk`?

src/pdc-api.ts
reefdog: Ohhh wait, `throwNotOk` is more like `throwIfNotOk`:

```ts
const throwNotOk = (res: Response): Response => {
  if (!res.ok) {
    throw new Error(`Response status: ${res.status} ${res.statusText}`);
  }
  return res;
};
```

src/pdc-api.ts
reefdog: nm, actually read the source for the function. I object to the unclear name!

slifty: I bet that is a confusingly named helper that means  "throw if not OK"

I'm gonna go ahead and confirm that bet, and then we should update it to something more clear.

slifty:
```
const throwNotOk = (res: Response): Response => {
  if (!res.ok) {
    throw new Error(`Response status: ${res.status} ${res.statusText}`);
  }
  return res;
};
```

My bet was right!  I'll change that in a separate commit.

src/pdc-api.ts
```diff
@@ -50,6 +50,101 @@ const usePdcApi = <T>(
...
+    method: 'post',
+    headers: {
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify(params),
+  });
+};
+
+const uploadPresignedPost = (
```
reefdog, line 102: I don't think this function name is clear. It took me several reads to understand that it's not "uploading a pre-signed post". It's uploading a file, using the pre-signed post. I think we should rename to be more legible? Perhaps `uploadFileWithPresignedPost`?

src/pdc-api.ts
reefdog: nm, I did it.

src/pdc-api.ts
```diff
@@ -50,6 +50,102 @@ const usePdcApi = <T>(
...
+  post: ApiPresignedPostResponse,
+  file: File,
+) => {
+  const formData = new FormData();
+  Object.entries(post.presignedPost.fields).forEach(([key, value]) => (
+    formData.append(key, value)
+  ));
+  formData.append('Content-Type', post.fileType);
```
reefdog, line 110: @slifty It strikes me that we're mildly missing an opportunity to verify file type here. We're setting content type to the presigned post's `fileType`, but shouldn't we be using the file's type instead, just to double-verify that the file matches the presigned post?

In that case, this should read like so instead:

```suggestion
  formData.append('Content-Type', file.type || 'application/octet-stream');
```

Thoughts?

src/components/BulkUploaderFilePicker.tsx
reefdog: Don't worry about it.

reefdog (commented):
When GitHub will and won't let me add single comments without "sUbMiTtiNg a ReViEW" baffles me.

reefdog:
So to summarize what @slifty and I have discovered and been discussing separately: the uploader only sometimes works. About half the time, it fails, with the API reporting a failure to load the object from S3.

<details>
<summary>Click to view example output from the API…</summary>

```
[2024-01-25T22:14:45.326Z] ERROR (66267): Failed to load an object from S3
    source: "/Users/justin/Projects/ots/clients/pdc/service/src/jobQueue.ts"
    key: "unprocessed/b5818b66-6447-4de7-9331-8d2dff7e3e25"
    scope: {
      "label": "job",
      "workerId": "worker-fcd17bfc1befc115b3",
      "taskIdentifier": "processBulkUpload",
      "jobId": "19"
    }
    err: {
      "type": "NoSuchKey",
      "message": "UnknownError",
      "stack":
          NoSuchKey: UnknownError
              at de_NoSuchKeyRes (/Users/justin/Projects/ots/clients/pdc/service/node_modules/@aws-sdk/client-s3/dist-cjs/index.js:6085:21)
              at de_GetObjectCommandError (/Users/justin/Projects/ots/clients/pdc/service/node_modules/@aws-sdk/client-s3/dist-cjs/index.js:4399:19)
              at processTicksAndRejections (node:internal/process/task_queues:95:5)
              at async /Users/justin/Projects/ots/clients/pdc/service/node_modules/@smithy/middleware-serde/dist-cjs/index.js:35:20
              at async /Users/justin/Projects/ots/clients/pdc/service/node_modules/@aws-sdk/middleware-signing/dist-cjs/index.js:184:18
              at async /Users/justin/Projects/ots/clients/pdc/service/node_modules/@smithy/middleware-retry/dist-cjs/index.js:320:38
              at async /Users/justin/Projects/ots/clients/pdc/service/node_modules/@aws-sdk/middleware-flexible-checksums/dist-cjs/index.js:173:18
              at async /Users/justin/Projects/ots/clients/pdc/service/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:97:20
              at async /Users/justin/Projects/ots/clients/pdc/service/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:120:14
              at async /Users/justin/Projects/ots/clients/pdc/service/node_modules/@aws-sdk/middleware-logger/dist-cjs/index.js:33:22
      "name": "NoSuchKey",
      "$fault": "client",
      "$metadata": {
        "httpStatusCode": 404,
        "requestId": "tx0000096e87feb18aca143-0065b2dd55-4b6a0-nyc3d",
        "attempts": 1,
        "totalRetryDelay": 0
      },
      "Code": "NoSuchKey",
      "BucketName": "pdc-service-local-justin",
      "RequestId": "tx0000096e87feb18aca143-0065b2dd55-4b6a0-nyc3d",
      "HostId": "4b6a0-nyc3d-nyc3-zg04"
    }
[2024-01-25T22:14:45.333Z] INFO (66267): Download of bulk upload file from S3 failed
    source: "/Users/justin/Projects/ots/clients/pdc/service/src/jobQueue.ts"
    scope: {
      "label": "job",
      "workerId": "worker-fcd17bfc1befc115b3",
      "taskIdentifier": "processBulkUpload",
      "jobId": "19"
    }
    err: {
      "type": "Error",
      "message": "Unable to load the s3 object",
      "stack":
          Error: Unable to load the s3 object
              at downloadS3ObjectToTemporaryStorage (/Users/justin/Projects/ots/clients/pdc/service/src/tasks/processBulkUpload.ts:104:9)
              at async processBulkUpload (/Users/justin/Projects/ots/clients/pdc/service/src/tasks/processBulkUpload.ts:380:20)
              at async doNext (/Users/justin/Projects/ots/clients/pdc/service/node_modules/graphile-worker/src/worker.ts:252:18)
    }
```

</details>

@slifty suggests a possible race condition. Despite the [seemingly correct use of `await`](https://github.com/PhilanthropyDataCommons/front-end/pull/519/commits/087b6e498e4a35a2743060d48d87bffb420f3dcd#diff-a76b3aff18029fc2ee297bf1a42bdc97eabafa946e1f897cc54ebd504c0122e7R20-R32), it seems that when we try to register the bulk upload, the key that was used to upload the file isn't yet… valid?

src/pdc-api.ts
```diff
@@ -50,6 +50,102 @@ const usePdcApi = <T>(
...
+  formData.append('file', file);
+  return fetch(
+    post.presignedPost.url,
+    {
+      method: 'post',
+      body: formData,
+    },
+  )
+    .then(throwIfResponseIsNotOk)
+    .catch((e) => logError(e, post.presignedPost.url, post));
```
slifty, line 120: I believe that by catching this here we hide failures, which in turn means if the upload fails we're unable to properly stop the flow at the next level.

If we want to catch the error here we need to at least re-throw.

It's possible this is the reason the bulk uploader is complaining about being unable to load a given key.

src/pdc-api.ts
slifty: I like it, but don't like having the "null type is overridden" in two places.
I'll probably go with something like `file.type || post.fileType`

slifty: Nevermind -- I don't even want to have access to `post.fileType` (I only want the presignedPost rather than the presignedPost request result, so I've refactored that function).  I'll just double up the `||` for now.

reefdog: That's a good concern. However what do you think about just making a constant out of the fallback type and referencing it both places that way? I do think following the same logic here as for the presigned-post-generation seems important, otherwise there's essentially no file type validation if the second file (used for upload) has a `null` type.

reefdog: @slifty Ugh I keep replying to old things because GH's interface isn't pulling in latest replies fast enough.

reefdog:
_Updating the PR with some additional detail:_

@slifty has been able to configure the Digital Ocean buckets with correct CORS settings. Now, I'm still getting inconsistent uploads, but at least we're seeing a response from the endpoint!

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Error>
    <Code>InvalidArgument</Code>
    <Message>Malformed JSON</Message>
    <BucketName>pdc-service-local-justin</BucketName>
    <RequestId>tx0000073953a1345dc82f0-0065b81232-50742-nyc3d</RequestId>
    <HostId>50742-nyc3d-nyc3-zg04</HostId>
</Error>
```

And things are catching correctly so that these errors — which are "upload failures" that should prevent the bulk upload from ever being registered — do indeed catch prior to registering the bulk upload, which means less nonsense in the sidebar.

slifty:
Hooray!

And yeah for the sake of history: what we thought was a race condition was actually just an error that was being masked by a CORS violation combined with a mistake that was causing errors to get caught / handled without propogating up to stop the upload flow from continuing.

I'll look into what this `malformed json` issue is at which point hopefully we'll be good to go!

slifty:
Uh oh -- https://www.digitalocean.com/community/questions/having-issues-uploading-images-using-pre-signed-post-urls -- we're not alone... could it be that there's a bug in digital ocean spaces?  OP says it began this past December 24th.

Why is it that we always seem to manage to find a way to trigger Broken Things from major companies?

slifty:
@reefdog I think since it looks like this is an upstream issue (Digital Ocean), then we should be good to merge this so long as it passes your review otherwise.  Let me know!

reefdog (commented):
Yaaaaaay! I mean, boo about Digital Ocean's bug, but yay that it's not currently on us.

@slifty As this is my PR, I can't approve it, but I do _in spirit_ approve it. Can you _actually_ approve it so we can merge?

slifty:
HOORAY!

PR #564, opened 2024-02-09
Refresh upload history

### LINKED ISSUE #527: Refresh bulk upload history after upload completes
#519 adds basic bulk upload capability, but crucially after the upload completes, the sidebar history doesn't show anything until the user refreshes the page. This is a showstopping limitation; the new upload should show up right away. (We should also poll the API for updates, but that's separately tracked as #415.)

slifty:
This PR changes the API hooks so that they expose the ability to re-run a given API call at a later time.  It also utilizes that new functionality to re-run the bulk upload list lookup every time the user creates a new bulk upload.

Resolves #527

### DIFF
--- a/src/pages/AddData.tsx
+++ b/src/pages/AddData.tsx
@@ -17,7 +17,7 @@ import { BulkUploaderFilePicker } from '../components/BulkUploaderFilePicker';
 import { BulkUploadList } from '../components/BulkUploadList';
 import './AddData.css';
 
-const BulkUploaderLoader = () => {
+const BulkUploader = () => {
   const createPresignedPost = usePresignedPostCallback();
   const registerBulkUpload = useRegisterBulkUploadCallback();
   const handleUpload = async (file: File) => {
@@ -84,7 +84,7 @@ const AddDataLoader = () => {
       <PanelGridItem>
         <NewBulkUploadPanel
           apiUrl={new URL('/', process.env.REACT_APP_API_URL)}
-          bulkUploaderLoader={<BulkUploaderLoader />}
+          bulkUploader={<BulkUploader />}
           baseFieldsLoader={<BaseFieldsLoader />}
         />
       </PanelGridItem>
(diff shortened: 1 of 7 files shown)


src/pdc-api.ts
```diff
@@ -47,7 +47,15 @@ const usePdcApi = <T>(
      */
   }, [path, params.toString()]);
 
-  return response;
+  useEffect(() => {
+    fetchData()
+      .catch(() => {});
+  }, [fetchData]);
+
+  return [
```
slifty, line 55: I wanted to point out that this is less explicit than an object with named attributes (i.e.  `{response: ..., refresh: ...}`).  That said, react already has this pattern of having hooks return tuples in an array.

I am comfortable with changing to an object if that would feel better.

src/pdc-api.ts
```diff
@@ -21,11 +21,11 @@ const logError = (error: unknown, path: string, params: object) => {
 const usePdcApi = <T>(
   path: string,
   params: URLSearchParams = new URLSearchParams(),
-): T | null => {
+): [T | null, () => void] => {
   const { fetch } = useOidcFetch();
   const [response, setResponse] = useState(null);
 
-  useEffect(() => {
+  const fetchData = useCallback(async () => {
```
slifty, line 28: This doesn't need to be async / it's not using await anywhere and it returns a void.

src/pages/AddData.tsx
```diff
@@ -32,36 +50,20 @@ const BulkUploaderLoader = () => {
       fileName: file.name,
       sourceKey: presignedPost.fields.key,
     });
+    onBulkUpload();
```
reefdog, line 53: Mind throwing an extra newline above this (for consistency with the rest of the function)?

reefdog (approved):
I left one minor requested whitespace change but otherwise this looks great.

src/pages/AddData.tsx
slifty: The 👎🏻 is meant as "no, I don't mind!"

reefdog: @slifty Dan I just burstlolled at how chaotic that was.

PR #582, opened 2024-02-20
Add download attr to CSV template link

reefdog:
On the bulk upload page, we link users to a CSV template hosted by the service repository. The service’s HTTP server gets to set the attachment disposition header that determines if the file is downloaded to disk or opened in-browser. Which is fine, and its prerogative! But we want *our* users to definitely download the file.

The `download` attribute to the rescue! We’ll also take this opportunity to set a value for the attribute, which will be the downloaded file’s filename.

There's no issue associated with this, as it's just a hotfix for something I noticed after #438 was merged and deployed.

**Testing:**
- Open deploy preview
- Log in (I will have added it to the Keycloak auth list)
- Go to the Add Data page
- Click the "Download CSV template of recommended base field keys" link
- On `main`/production, this will stream in-browser, but in this PR deploy preview, it will download to disk as `pdc-bulk-upload-template.csv`

### DIFF
--- a/src/components/NewBulkUploadPanel.tsx
+++ b/src/components/NewBulkUploadPanel.tsx
@@ -51,7 +51,11 @@ export const NewBulkUploadPanel = ({
             </ul>
 
             <h2 className="label">Optional</h2>
-            <a href={bulkUploadTemplateUrl.toString()} className="has-icon no-visited">
+            <a
+              href={bulkUploadTemplateUrl.toString()}
+              className="has-icon no-visited"
+              download="pdc-bulk-upload-template"
+            >
               <DocumentDuplicateIcon className="icon" />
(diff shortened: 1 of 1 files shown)


slifty (approved):
ugh, pipes in file names

looks good!

reefdog:
@slifty Thanks! But one reason I hadn't asked for your review yet is I wanted to test this on Netlify myself to make sure the behavior was as expected — and it turns out it's not, for [good reason](https://stackoverflow.com/questions/28318017/html5-download-attribute-not-working-when-downloading-from-another-server-even)! So I'm actually going to close this PR and *not* merge it in.

But… thanks for the fast review anyway! I hope I can take advantage of your good teamwork later, on a more worthy task.

PR #584, opened 2024-02-20
Fix Bulk Upload List Refresh on File Upload 

### LINKED ISSUE #572: Fix Bulk Upload list refresh "flash"
When the list of bulk upload items is updated it creates a "flash" as the entire list is re-rendered.

This *should* be something we can avoid, since only the bulk uploads that have been updated or added would need to be re-rendered.  It's possible we are using array-index keys instead of bulk upload ID keys, or maybe something else.

This is probably important to fix before addressing #415

hminsky2002:
On the bulk upload page, when a user uploads a new file, the entire BulkUploadList component re-renders. This was fixed by using react state for the BulkUploads array that contains the metadata of the uploaded files. Now, on component update, only changed/added list items are re-rendered. 

Closes #572 

Testing:

Open dev instance
Log in 
Go to the Add Data page
Inspect the Bulk Upload History panel, and select a pre-existing component in the react-dev-tools inspector
Upload a new file, and verify that the pre-existing component was not re-rendered

### DIFF
--- a/src/pages/AddData.tsx
+++ b/src/pages/AddData.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect } from 'react';
+import React, { useEffect, useState } from 'react';
 import { NavLink } from 'react-router-dom';
 import { withOidcSecure } from '@axa-fr/react-oidc';
 import { DocumentPlusIcon } from '@heroicons/react/24/outline';
@@ -63,11 +63,14 @@ const BulkUploader = ({
 
 const AddDataLoader = () => {
   const [bulkUploadResponse, refreshBulkUploads] = useBulkUploads();
-  const bulkUploads = bulkUploadResponse?.entries ?? [];
+  const [bulkUploads, setBulkUploads] = useState(bulkUploadResponse?.entries ?? []);
 
   useEffect(() => {
     document.title = 'Add Data - Philanthropy Data Commons';
-  }, []);
+    if (bulkUploadResponse?.entries) {
+      setBulkUploads(bulkUploadResponse.entries);
+    }
+  }, [bulkUploadResponse]);
 
   return (
     <PanelGrid sidebarred>


reefdog:
Huzzah! I'll review this promptly, thanks @hminsky2002!

One quick note: it looks like you merged `main` into your branch to get the latest commits. We use rebase for this so there aren't any unnecessary merge commits, reserving merge commits only for when we merge PRs back to `main`. How's your git-fu? Are you able to drop the merge commit from the history and force push?

BTW, if he hasn't shared with you already, @slifty has a collection of incredibly useful git aliases and functions that make doing history rewriting / recomposing super easy. It's worth spending a little time with him on that, or else some of the PR requests that we make will seem onerous — when, with the right tools, they're fairly trivial to implement.

Okay, with that out of the way, review… starting!

reefdog:
(Oh one more thing: GitHub supports a couple of "action keywords" on PRs, like saying `Closes #572` on a newline will actually close the referenced issue when the PR is merged. Of course we reserve those keywords for when they actually apply, but in this case, it does!)

.gitignore
```diff
@@ -118,3 +118,6 @@ dist
 # @axa-fr/react-oidc generated files
 /public/OidcServiceWorker.js
 /public/OidcTrustedDomains.js
+
+#.vscode
```
slifty, line 122: Will leave a full review later but short term: I'd encourage you to add this to your personal global git ignore (the philosophy here is that this ignore line is specific to you personally and shouldn't be reflected in the project at large / it's up to each dev to locally control the files that their IDEs would normally try to insert)

[Here's an example of what my global git ignore looks like](https://github.com/slifty/dotfiles/blob/main/git/gitignore.symlink).  You should be able to put it at ~/.gitignore.

And then you can update your git config to point to that file: https://github.com/slifty/dotfiles/blob/a8f026cd7ccb38b57f1fc35d09a8f0e05c997028/git/gitconfig.symlink#L12

reefdog (changes requested):
Harry, this is great! It solves the problem and appears to be the idiomatically correct solution. 👏

The only requested changes I have are hygienic.

- 05693bb52f53ddf51116e0f86239da2d8c414884:
   - Edit this commit to remove the checked-in `.vscode` so it's not even in history.
   - Rewrite the title to be in present tense.  The pattern we use is to have the commit title finish the statement "[If I apply this commit, it will…] Fix full reload of Upload Items on Add Data"
   - Rewrap the body at 72 chars, not 50. (50 is the limit for the title / first line, but body lines should be up to 72 chars.)
   - Add an issue reference to the bottom of the commit message, e.g. `Issue #572 Fix Bulk Upload list refresh "flash"`
- 7ec489383e08ea4c454538d7d43febea09747906: Drop entirely (since you'll have removed the `.vscode` from 05693bb52f53ddf51116e0f86239da2d8c414884 and moved the ignore setting to your global `.gitignore`

hminsky2002:
> Harry, this is great! It solves the problem and appears to be the idiomatically correct solution. 👏
> 
> The only requested changes I have are hygienic.
> 
> * [05693bb](https://github.com/PhilanthropyDataCommons/front-end/commit/05693bb52f53ddf51116e0f86239da2d8c414884):
>   
>   * Edit this commit to remove the checked-in `.vscode` so it's not even in history.
>   * Rewrite the title to be in present tense.  The pattern we use is to have the commit title finish the statement "[If I apply this commit, it will…] Fix full reload of Upload Items on Add Data"
>   * Rewrap the body at 72 chars, not 50. (50 is the limit for the title / first line, but body lines should be up to 72 chars.)
>   * Add an issue reference to the bottom of the commit message, e.g. `Issue #572 Fix Bulk Upload list refresh "flash"`
> * [7ec4893](https://github.com/PhilanthropyDataCommons/front-end/commit/7ec489383e08ea4c454538d7d43febea09747906): Drop entirely (since you'll have removed the `.vscode` from [05693bb](https://github.com/PhilanthropyDataCommons/front-end/commit/05693bb52f53ddf51116e0f86239da2d8c414884) and moved the ignore setting to your global `.gitignore`

Sorry about the delay on this, thank you for the feedback and compliment Justin! I have reset the branch to the first commit and made the changes. Let me know if anything else needs to be modified!

slifty (approved):
Magical!  Great job figuring out the issue.

I ran this locally and indeed there's no flash when a new bulk upload is added.

reefdog:
Looks like you'll need to rebase off `main` and force-push again. Or, you could give GitHub's built-in interface for this a try, just be sure to switch to the [non-default] rebase option:

image

Although when I tried the built-in rebase myself on my PR for the first time, it stripped the Verified status off my commits. I suspect that's because I sign with my SSH key and not GPG key? You could try and test my theory!

slifty:
@hminsky2002 Since it's good to know how to do it without github UX, the commands to fix this are:

`git rebase main` from the command line and then `git push -f` and you should be all set

PR #585, opened 2024-02-21
Add Developers and About menus to site nav

### LINKED ISSUE #557: Modify Dropdown component to work with the site nav
To support drop-down menus in the navbar, we need to add a generic drop-down menu component. These will be lightly-enhanced `<details>` elements. Initially, the menu items will all be "title/description"-style, but we can support other types in the future.

**Example:**
Drop-down menu

### LINKED ISSUE #558: Create About menu in navbar
We should add an About menu item to the navbar that provides all the links from the landing page, but in a universally-accessible way:

About menu

The destination for these links are the same destinations as on the landing page:

- [About the PDC project](https://philanthropydatacommons.org/)
- [Need an account?](mailto:jimmcgowan@opentechstrategies.com)
- [Questions or feedback?](mailto:info@philanthropydatacommons.org)
- [Recent improvements](https://github.com/PhilanthropyDataCommons/front-end/blob/main/RECENT_CHANGES.md)

### LINKED ISSUE #559: Create Developer menu in navbar
We need to add a Developer menu in the navbar with links to development-related PDC resources.

Developers

- [API Documentation](https://api.philanthropydatacommons.org)
- [TypeScript SDK](https://www.npmjs.com/package/@pdc/sdk)

We should also move the [dev-only Storybook link](https://github.com/PhilanthropyDataCommons/front-end/blob/24dd3d3e4fd9ae8d1f80d87d00a0544ed8a6fd66/src/components/AppNavbar.tsx#L16-L23) into this menu.

reefdog:
This PR reworks the `Dropdown` component to be useable with the site nav, and adds a **Developers** menu and an **About** menu:

https://github.com/PhilanthropyDataCommons/front-end/assets/4731/c4cfc590-13cc-4ee8-a7ab-75c63f98a806

**Note:** The PR appears large because of the refactor of `Dropdown`, and because of some seemingly-unrelated (but actually prerequisite) changes from the changes to the `<code>` element styling.

**Testing:**
- On `main`/[production](https://app.philanthropydatacommons.org), note the lack of dropdown menus and the presence of the "Feedback" element in the navbar.
- In the Netlify deploy preview, note the presence of the Developers and About menus; the presence of the Storybook link in the Developers menu; the presence of the "Need an account?" item in the About menu; and that the Feedback item has been moved into the About menu.
- Locally, note the absence of the Storybook item in the Developers menu. (It's only present in environments where `REACT_APP_SHOW_STORYBOOK=true`, which is intended only to be the deploy preview environments.)
- After logging in locally, note the absence of the "Need an account?" item in the About menu.
- Verify that drop-downs still work, by going to any proposal page and opening the "Data providers" menu. (You don't need to actually load the data providers panel; just seeing the menu is sufficient.)

Closes #557 
Closes #558 
Closes #559


src/components/Dropdown/DropdownMenu.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+
+interface DropdownMenuProps {
+  children: React.ReactNode;
+  /**
+   * Align the menu to the left or right edge of its toggle.
+   */
+  align?: 'left' | 'right';
```
slifty, line 8: since `'left'` and `'right'` are used in a bunch of places could these values be turned into an enum instead, that gets referenced in all those places?

src/components/Dropdown/DropdownTrigger.tsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import '../Button.css';
+
+interface DropdownTriggerProps {
+  children: React.ReactNode;
+  type?: 'button' | 'navbar-item';
```
slifty, line 6: same enum idea here, for your consideration

slifty (approved):
Neat!  Looks great.

I'm approving, but there are some small UX things that don't directly relate to lines of code and please take or leave:

1. What would you think of removing closing punctuation from the link descriptions/  It seems a little odd but maybe that's just me
2. It'd be cool to have a bolder visual hierarchy between the link titles and descritpions (making the descriptions slightly lighter, smaller, or making the titles slightly larger)  right now it's hard for me to know the actual links at a glance / the descriptions make it harder to satisfice.
3. When I have a drop down open would it be possible to allow me to swap to opening the other dropdown without having to click twice? e.g. right now:

(A) Click "About" (opens about dropdown)
(B) Click "Developers" (closes about dropdown)
(C) Click "Developers" again (opens developers dropdown)

Would be nice to have:

(A) Click "About" (opens `about` dropdown)
(B) Click "Developers" (closes `about` dropdown and opens `develoeprs` dropdown)

reefdog:
> Neat! Looks great.

🎉 

> 1. What would you think of removing closing punctuation from the link descriptions/

My rule about closing punctuation on repeated elements is to (1) be consistent across all items in the set and (2) omit them if possible, but (3) don't awkwardly contort language.

I'd be happy to omit them from the descriptions, but some worked better with multiple sentences, so rule ① meant they all got them.

I welcome edits to improve the language in a way that lets us omit them, but sounds like this is a light request anyway and not worthwhile!

> 2. It'd be cool to have a bolder visual hierarchy between the link titles and descritpions (making the descriptions slightly lighter, smaller, or making the titles slightly larger)  right now it's hard for me to know the actual links at a glance / the descriptions make it harder to satisfice.

Fair point! I intentionally mimicked the navbar link style for the titles, and I've lately been trying to minimize a visual hierarchy that ended up with inaccessible text sizes or contrasts. In this case my first pass actually did have the hierarchy you mentioned, but when I saw how relatively small the descriptions were, I reevaluated my biases to see if full-size/contrast text worked.

_I_ thought it actually ended up surprisingly fine… but I acknowledge that's visually subjective. Does the accessibility argument help?

> 3. When I have a drop down open would it be possible to allow me to swap to opening the other dropdown without having to click twice? e.g. right now:

I grok and agree with this desire. Initially I thought the implementation of this would be too difficult, but… I think I just realized how to do it fairly easily! Will implement…

reefdog:
@slifty Actually… the item 3 is indeed gonna be more difficult than I'd hoped. I'm going to lean on your approval and just put this off for another day!

Quick reason: our [Dropdown](https://app.philanthropydatacommons.org/storybook/?path=/docs/stories-dropdown--docs) component uses the neat trick of `<details>`/`<summary>` elements, which is actually what GitHub uses for some of its dropdowns (but — notably, I just discovered! — not those in its navbar). When the dropdown is open, the `<summary>` element (which is the trigger for opening/closing the dropdown) stretches an invisible overlay across the entire viewport, below the dropdown, so that clicking anywhere else on the page closes the dropdown. All sans JavaScript! (Yes, we are literally a JS-requiring React app, but I still believe in using native controls whenever possible, for further reasons I'm happy to discuss elsewhere.)

While there are all sorts of neat advantages to this implementation, one downside is you have to close the dropdown before using anything else.

I was hoping to change the behavior specifically for navbar dropdowns, and nearly got there, but started to domino issues and hit my timebox. Thus, will need to save it for complaints.

PR #590, opened 2024-02-23
Add Poll API for bulk upload processing

### LINKED ISSUE #415: Poll API for bulk upload processing status
Users shouldn't have to reload the page to check [processing status of bulk uploads](#414), but nor do we need the responsiveness of a WebSocket. The page (component?) should poll the API for status updates on some reasonable interval, perhaps every 10 seconds to start (we can tweak based on real-world tests).

hminsky2002:
On the bulk upload page, when a user uploads a new file, the BulkUploadListItem will not update it's status until the page is refreshed. I have added a poll in the form of an interval set inside of the AddData page's useEffect hook that will poll the API every 2 seconds(a time amount I am open to changing, depending on group thought) and rerender any BulkUploadListItem's with changed status. I also added a callback function to the hook that clears the interval, so as to end the interval if the component ever dismounts. Based on my research this seems to be the convention when using setInterval.

Closes #415

Testing:

Open dev instance
Log in
Go to the Add Data page
Check the network tab in browser dev tools to verify a call to the backend is happening every 2 seconds
Upload a new file, verify that after it finishes uploading that is reflected in the bulk upload pannel

### DIFF
--- a/src/pages/AddData.tsx
+++ b/src/pages/AddData.tsx
@@ -17,6 +17,8 @@ import { BulkUploaderFilePicker } from '../components/BulkUploaderFilePicker';
 import { BulkUploadList } from '../components/BulkUploadList';
 import './AddData.css';
 
+const REFRESH_BULK_UPLOADS_INTERVAL = 5000;
+
 const BaseFieldsLoader = () => {
   const [fields] = useBaseFields();
   if (fields === null) {
@@ -67,10 +69,19 @@ const AddDataLoader = () => {
 
   useEffect(() => {
     document.title = 'Add Data - Philanthropy Data Commons';
+
     if (bulkUploadResponse?.entries) {
       setBulkUploads(bulkUploadResponse.entries);
     }
-  }, [bulkUploadResponse]);
+
+    const intervalId = setInterval(() => {
+      refreshBulkUploads();
+    }, REFRESH_BULK_UPLOADS_INTERVAL);
+
+    return () => {
+      clearInterval(intervalId);
+    };
+  }, [bulkUploadResponse, refreshBulkUploads]);
 
   return (
     <PanelGrid sidebarred>


reefdog (changes requested):
This, too, is great. Funnily enough I tried to implement this myself and struggled, but… you make it look so clean and obvious! 😆 👏

I have a formatting suggestion and a cadence configuration change, but nothing significant.

Re: your commit message, over-communication generally beats under-communication, but you don't need to describe the internals to quite that extent! Your code was clear and legible and does most of the talking; a commit message can add necessary context and justify an abnormal solution. But your solution here was clear and idiomatic, and solved the problem in the title, so you probably don't even need a description (besides the issue tag). But consider this graf a suggestion, not a requirement!

I'll give this a re-review as soon as you're ready.

reefdog (approved):
Looks great, and thanks for implementing the feedback. Merge as soon as you'd like.

slifty:
Just to +1 Justin's note about commit messages: you should let the code describe itself, and the commit message can describe the *why* or the *what* or any other special considerations / resources that led you to implement the solution you went with that isn't clear in the code.

src/pages/AddData.tsx
```diff
@@ -67,10 +69,17 @@ const AddDataLoader = () => {
...
     if (bulkUploadResponse?.entries) {
       setBulkUploads(bulkUploadResponse.entries);
     }
-  }, [bulkUploadResponse]);
+
+    const intervalId = setInterval(() => {
+      refreshBulkUploads();
+    }, REFRESH_BULK_UPLOADS_INTERVAL);
+
+    return () => clearInterval(intervalId);
```
slifty, line 81: Super subtle point here but right now I think it is circumstantial that the return value of `clearInterval` happens to match the expected return value of the returned function of `useEffect`

Which is to say, this would be safer / more explicitly returning `void` if we wrote it as:

```
    return () => {
        clearInterval(intervalId);
    }
```
Comment

reefdog: (But truly great catch.)

hminsky2002: This makes sense! I was not thinking about the return value matching useEffect, seems like the right decision to add the block. I just added it!

PR #605, opened 2024-02-28
Add Prettier

### LINKED ISSUE #497: Set up prettier on the front-end
_Edited by @reefdog on 2023-02-23 to add:_ Here's the `service` repo's Prettier PR: https://github.com/PhilanthropyDataCommons/service/pull/695

hminsky2002:
This PR adds prettier and invokes it on the project.

It follows the same conventions as @slifty 's  [PR to add prettier to the service repo](https://github.com/PhilanthropyDataCommons/service/pull/695), including switching from spaces
to tabs. Thank you @slifty!

Resolves #497


src/pdc-api.ts
reefdog, line 1: I don't think we actually want this change! Unlike the [overrides removed in the service repo](https://github.com/PhilanthropyDataCommons/service/pull/695/commits/88f0bcb13b54b6507f7c7061b4be9397605bec90), these are truly lint-related, not formatting-related. Without them, the ESLint warnings are now back.

I think you can completely drop this commit from the PR.

reefdog (changes requested):
This is really good @hminsky2002, thank you! This will be so nice.

I have a handful of inline requested changes, including dropping the last commit (those ESLint overrides still matter here). And of course, you'll have to be sure to update the commit hash in `.git-blame-ignore-revs` and all that.

Should be a quick re-review!

package.json
```diff
@@ -47,7 +49,9 @@
     "build:storybook": "storybook build -o build/storybook",
     "storybook": "storybook dev -p 6006",
     "test": "react-scripts test",
-    "lint": "eslint src/",
+    "lint:prettier": "prettier . --check",
+    "lint": "npm run lint:prettier",
```
reefdog: Ooh wait, I had missed this! We don't want `lint` to just run Prettier, we want it to run ESLint still at a minimum. Do we _also_ want it to run Prettier? I… actually don't know. cc @slifty what do you think?

The path forward is probably either:

1. Change `lint` back to `eslint src/` (and… add `lint:prettier` to CI?)
2. Change `lint` to `eslint src/ && npm run lint:prettier`?

reefdog: Update: Ooh, I see [what `service` did](https://github.com/PhilanthropyDataCommons/service/pull/695/commits/f6c7b400920657e2198b138f22103bbb5c578782#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519), and I like it. Let's copy that! So we actually need to add a `lint:ts` action that runs ESLint, then have our `lint` action run both `lint:ts` and `lint:prettier`.

For front-end, I think we probably need to lint both `.ts` and `.tsx` files… and it may not hurt to include `.js` files? Again @slifty what do you think. (The only `.js` files we have were generated by our OIDC library and live in `public/`.)

src/pdc-api.ts
```diff
@@ -7,252 +7,245 @@ const logger = getLogger('pdc-api');
...
+				.catch((e) => logError(e, path, options));
+
+			/* eslint-disable-next-line react-hooks/exhaustive-deps --
+			 *
+			 * fetch should not be a dependency, because although it or its internal
+			 * state may change from render to render, such changes are not relevant:
+			 * a change in the way we make a request should not trigger an API request
+			 */
+		},
+		[path],
```
reefdog, line 77: After Prettifying this section, the `eslint-disable-next-line` command is now _two_ lines away from the target line (`[path],`), which means ESLint complains again.

Can you edit this commit to move the command to just above the `[path],` line? The entire return should look like:

```ts
	return useCallback(
		(options: RequestInit) => {
			const url = new URL(path, API_URL);
			return fetch(url, options)
				.then(throwIfResponseIsNotOk)
				.then((res) => res.json())
				.catch((e) => logError(e, path, options));
		},
		/* eslint-disable-next-line react-hooks/exhaustive-deps --
		 *
		 * fetch should not be a dependency, because although it or its internal
		 * state may change from render to render, such changes are not relevant:
		 * a change in the way we make a request should not trigger an API request
		 */
		[path],
	);
```

You'll once again have to update the following commit with the new hash, sorry!

I'd recommend doing an `npm run lint` before actually committing this to be sure it's all clean. 🙏

reefdog (changes requested):
One last thing! The Prettier reformatting actually moved an ESLint command block in such a way it no longer applies where intended. Dropped a note about where. I think that should be all! 🤞

hminsky2002 (commented):
changes made!

package.json
slifty: yes! where you landed!

package.json
```diff
@@ -1,131 +1,137 @@
 {
-  "name": "front-end",
-  "version": "0.1.0",
-  "private": true,
```
slifty, line 4: Whoops!  It looks like you included the invoked prettier changes on this file in the "add prettier" commit -- this makes it hard to see what changes you made to add prettier.

Sorry for this, but would it be possible to split that out?  This would involve reverting this file to the previous state, re-making the change in this commit, and then adding the prettified version to the  "invoke oprettier" commit

package.json
```diff
@@ -1,131 +1,137 @@
...
+	},
+	"scripts": {
+		"start": "react-scripts start",
+		"build": "npm run build:app && npm run build:storybook",
+		"build:app": "react-scripts build",
+		"build:storybook": "storybook build -o build/storybook",
+		"storybook": "storybook dev -p 6006",
+		"test": "react-scripts test",
+		"lint:prettier": "prettier . --check",
+        "lint:ts": "eslint ./src --ext .ts,.tsx,.js",
```
slifty: let's rename this from "lint:ts" since it does more than just `ts` (e.g. let's something like `lint:eslint`)

slifty (changes requested):
eep -- commit mistake; happy to work with you to fix if helpful

package.json
reefdog: Oops, I missed this. (Agree!)

reefdog: @hminsky2002 I'm not seeing this change in the most recent code.

So to be clear we're talking about git history hygiene: in 4fe57a3be50d2ab12b549fd079e7b0ffdfc7673b, it's impossible to tell which `package.json` lines are new, because the file has simultaneously been Prettified. The Prettification of the file shouldn't happen until c9b81c71974a2cf9d39bbd3bc8587469fba7daa1, which of course you will then flag as ignored in the very next commit.

Sorry to keep asking for changes, but could you edit these so that the `package.json` changes in the first commit are only what's needed to add Prettier (including the new `lint:` scripts, etc.), then reserve the file Prettification for the next commit? 🙏

reefdog (changes requested):
Thanks for changing the npm script name!

It looks like the requested change to Git history for the `package.json` didn't happen though, can you read [this comment](https://github.com/PhilanthropyDataCommons/front-end/pull/605#discussion_r1511886258) and see if it makes sense? Happy to hop on a call to talk through (especially if the git surgery feels torturous)!

hminsky2002:
> Thanks for changing the npm script name!
> 
> It looks like the requested change to Git history for the `package.json` didn't happen though, can you read [this comment](https://github.com/PhilanthropyDataCommons/front-end/pull/605#discussion_r1511886258) and see if it makes sense? Happy to hop on a call to talk through (especially if the git surgery feels torturous)!

Argh! I don't know how that happened again, let me try and fix it again one more time then I'd be happy to call! My apologies that this is taking so long. Completely understand the need to fix this

hminsky2002:
> Thanks for changing the npm script name!
> 
> It looks like the requested change to Git history for the `package.json` didn't happen though, can you read [this comment](https://github.com/PhilanthropyDataCommons/front-end/pull/605#discussion_r1511886258) and see if it makes sense? Happy to hop on a call to talk through (especially if the git surgery feels torturous)!

@reefdog I just verified in the commit history that now the only changes for the package.json in my first commit are the added packages, scripts and the prettier extends line! I think what was happening was my vscode prettier was automatically formatting the file and I wasn't even realizing it 😭

reefdog:
@hminsky2002 Oh man, what a trial. I'm so sorry! Taking a look at it now. Thanks for sticking with it.

reefdog (approved):
Fantastic! Thank you for all the iterations, this is great and I'm happy to have it. 👏  Merge at your leisure.

PR #616, opened 2024-03-05
Add Organization Detail pages

### LINKED ISSUE #598: Create organization detail page
After I create the necessary designs for the org detail pages in #597 — which will be significantly similar to the proposal detail page, but with a few differences — we'll need it to be implemented. They'll live at `/organizations/{id}` and will be publicly-accessible.

_Note: If you'd like to get started on this prior to #597's completion, you could go ahead and start building the routes and API calls._

hminsky2002:
This PR adds views for Individual organization pages. They are similar in appearance to the proposal views, but the data itself has only a few fields, so there is no additional data processing after it is gathered from the api. It also introduces two new interfaces, `ApiOrganization` and `ApiOrganizations`, that are defined in the pdc-api.ts file, along with two new exported constants "ORGANIZATIONS_DEFAULT_PAGE" and "ORGANIZATIONS_DEFAULT_COUNT" that are used for querying the service repo. 

As per @reefdog's extremely helpful mockups, 
Here is the requested design 
image

And what it looks like on my machine:
Screen Shot 2024-03-04 at 11 38 17 PM

### Testing
- run `npm run start` in the front-end repository
- run `npm run start` in the service repository
- if your local service does not currently have any organization data, it can be populated with a post request to the 
`organizations` endpoint, with a json body of the form
  `{
 "employerIdentificationNumber":"number",
 "name":"name"
}`
(if the relation does not yet exist, it can be created with the following sql code, thanks @slifty!)
 ```
  CREATE TABLE organizations (
  id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
  employer_identification_number VARCHAR NOT NULL,
  name VARCHAR NOT NULL,
  created_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
  UNIQUE(employer_identification_number, name)
);

COMMENT ON TABLE organizations IS
  'Organizations that have data registered in the PDC.';
```
- on the front-end in your browser, go to `http://localhost:3000:organizations/1`

Closes #598

### CODE REVIEW

File: src/components/OrganizationListGrid.tsx [hunk 1]
```diff
@@ -0,0 +1,46 @@
...
+		// If we just scrolled the active organizations into view, however,
+		// the user might not realize there are more organizations
+		// "above" the active one; thus, we scroll to the previous element.
+		// If it doesn't exist -- i.e., if the first organization is active --
+		// then we can noop since it's visible by default anyway.
+		//
+		// Note that we only run this on initial page render,
+		// not every time the active organization changes.
+		// This is because we only need to orient the user on load;
+		// after that, their clicks run the show.
```
slifty, line 29: This comment make sense but I wanted to offer a suggestion (side note -- you'll find I often make suggestions when I see comments!  It's a bit of a philosophy that a comment is USUALLY something that indicates code that isn't self-documenting)

1) Can we pull this one-liner into a named function such as `scrollToActiveItem`

2) Regarding the comment itself, though, if feels like this is a band aid (to be clear: animating the scroll sounds great / like the right UX regardless of reason!).  What I mean is: what if someone clicked and looked away, and didn't see the animation?  What if they scrolled, left the tab open, came back the next day and forgot the context?

It seems like the real solution to the problem described in the comment would be an additional UI element that indicates there is content "up".  I'm not suggesting that this get added in this PR, but *am* suggesting that we can remove the comment and instead make an issue for it.

I also think that the content of the comment probably belongs in the commit message instead.  Commit messages are contextually locked to a moment in time, whereas a comment may well get stale over time and become tech debt.

slifty (changes requested):
Let's use our SDK for this instead of defining a custom type for Organization and OrganizationBundle!

I also have a request to move a code comment into the content of the commit message, AND to open an issue related to the need for a UI component to solve the core problem that the comment described.

@reefdog I hope you can take a look at this PR too with a keener eye towards things like the CSS / react implementation!

src/pdc-api.ts
```diff
@@ -233,18 +233,51 @@ const useProviderData = (externalId: string) =>
 		}),
 	);
 
+interface ApiOrganization {
```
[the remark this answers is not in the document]
hminsky2002: Oh heck yes! That's so cool, I would gladly incorporate this, I'll limit it to the organizations for now but would this be something we'd use to replace all other types eventually?

src/components/OrganizationListGrid.tsx
hminsky2002: I will say that this is a shamelessly copied comment/functionality from the current ProposalListGrid component so it might be something that @reefdog can better respond to!

slifty: Ha I feel even stronger that it shouldn't be included then!  Strongly suggest you DRY this up and toss it into the commit message (or just don't include it)

I do think that the named helper function would be good either way though.

slifty: Oh, actually -- can we DRY the entire concept of a list grid somehow?

It sounds like there is some redundant code; would it make sense to have some kind of generic `ListGrid` component that provides the common functionality and somehow you can override the component that actually goes inside?

src/pdc-api.ts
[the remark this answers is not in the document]
reefdog: Big +1 to this, and also to (eventually, in another PR) replacing our API-linked interfaces with those from the SDK.

src/components/OrganizationListGrid.tsx
reefdog: I pretty much agree with all of the above, which pains me to say. (And yes, much of it applies to my original implementation that Harry was just re-using. Eep!)

- the scroll behavior would be more self-documenting as a named function
- the overly-large and peacocking comment belonged primarily in a commit message; feel free to remove it (except the bit that belongs as part of the jsdoc)
- abstracting out a generic `ListGrid` (now that we have a second version) is probably worthwhile

src/components/OrganizationListGridItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const OrganizationListGridItem = ({
+	organization,
+	active = false,
+}: OrganizationListGridItemProps) => (
+	<Link
+		to={`/organizations/${organization.id}`}
+		className={`organization-list-grid-item ${active ? 'active' : ''}`.trim()}
+	>
+		<div className="orginization-name">{organization.name}</div>
```
reefdog, line 19:
```suggestion
		<div className="organization-name">{organization.name}</div>
```

(Will need to update CSS accordingly.)

src/components/OrganizationListGridItem.tsx
```diff
@@ -0,0 +1,24 @@
...
+		<div className="organization-EIN">
```
reefdog, line 20: As a convention let's use all lowercase CSS class names:

```suggestion
		<div className="organization-ein">
```

(Will need to update CSS accordingly.)

src/pages/OrganizationDetail.tsx
```diff
@@ -0,0 +1,80 @@
...
+};
+
+const OrganizationDetailPanelLoader = () => {
+	const params = useParams();
+	const organizationId = params.organizationId ?? 'missing';
+	const [organization] = useOrganization(organizationId);
+
+	useEffect(() => {
+		if (organization === null) {
+			document.title = 'Organization Detail - Philanthropy Data Commons';
```
reefdog, line 43:
```suggestion
			document.title = 'Loading… - Philanthropy Data Commons';
```

(I know, this is a change from the proposal page. That's okay!)

src/pages/OrganizationDetail.tsx
```diff
@@ -0,0 +1,80 @@
...
+		} else {
+			document.title = `${organization.name} Organization Detail - Philanthropy Data Commons`;
```
reefdog, line 45:
```suggestion
			document.title = `${organization.name} - Philanthropy Data Commons`;
```

src/pages/OrganizationDetail.tsx
```diff
@@ -0,0 +1,80 @@
+import React, { useEffect } from 'react';
+import { useParams } from 'react-router-dom';
+import { withOidcSecure } from '@axa-fr/react-oidc';
+import { PanelGrid, PanelGridItem } from '../components/PanelGrid';
```
reefdog, line 4: This should be imported after the `pdc-api` imports, along with other component imports.

src/App.tsx
```diff
@@ -7,6 +7,7 @@ import { ProposalDetail } from './pages/ProposalDetail';
 import { ProposalList } from './pages/ProposalList';
 import { Landing } from './pages/Landing';
 import './App.css';
+import { OrganizationDetail } from './pages/OrganizationDetail';
```
reefdog, line 10: Move this import above the CSS import, with the other pages.

src/components/OrganizationListGridItem.css
```diff
@@ -0,0 +1,39 @@
...
+.organization-list-grid-item:last-child {
+	border-bottom: none;
+}
+
+.orginization-name {
+	grid-area: file-name;
+	font-weight: var(--font-weight--medium);
+	font-size: 1.1em;
+	word-wrap: break-word;
+}
```
reefdog, line 34: Looks like you snagged this styling from `BulkUploadListItem.css`. This should really mimic `.proposal--applicant-name` from `ProposalListGridItem.css`, which basically just means `font-weight`.

src/components/OrganizationListGridItem.css
```diff
@@ -0,0 +1,39 @@
+:root {
+	--organization-list-grid-item--details-color: white;
```
reefdog, line 2: Not sure where `white` came from 😆 should be `var(--color--gray--medium-dark)`

src/components/OrganizationListGridItem.css
```diff
@@ -0,0 +1,39 @@
...
+	grid-area: file-name;
+	font-weight: var(--font-weight--medium);
+	font-size: 1.1em;
+	word-wrap: break-word;
+}
+
+.organization-EIN {
+	font-weight: var(--font-weight--light);
+	color: var(--color--gray--medium-dark);
+}
```
reefdog, line 39: This should match the equivalent element from the `ProposalListGridItem.css` (and class name lowercased per other comment)

```css
.organization-ein {
	color: var(--proposal-list-grid-item--details-color);
	font-size: 0.9em;
}
```

…but really this is just proof that yes, we should abstract out a `ListGridItem` because this styling is exactly the same across both components. hmm.

reefdog (dismissed):
I'm really excited to have this functionality! 🎉 And it looks/works great, thanks!

I do have some inline changes requested, mostly small. And while the final PR ended up very clean, I noticed a few things along the way that we should avoid:

1. You referenced `useOrganization()` in a commit before implementing it. While noting that in the commit message was good, we still shouldn't use a function before defining it. (Linting each commit would catch this sort of thing too!)
2. You added the `baseField` logic in one commit (copied from the original proposal context), then removed it in the next. If this had happened between two different PRs as we're iterating toward something, I'd shrug it off. But since this is all in one branch/PR, we have the opportunity to keep a cleaner commit history (and make reviewing easier) by editing it out of the prior commit rather than removing it in a later commit. Your git-fu seems to be great, but if that sort of request gives you a headache, get with Dan or I to help set you up with some aliases we use to make it pretty darn straightforward!
3. Somewhat related to the above, it looks like some things that would've been caught by the linter or formatter were added in one commit then resolved by the next (e.g., whitespace in the App.tsx path definitions). It'd be a good idea to run eslint and Prettify on each commit individually so we don't end up with that sort of churn. (Sigh, definitely time to rethink my precommit hook stance.)

Lastly, and most pedantically: commit messages are good quality, but could you be sure to wrap them consistently at 72 chars, and do a spell check pass on them? That'll help our editors when we publish the great PDC Development Retrospective Book (coming this fall).

src/components/OrganizationListGrid.tsx
slifty: For help Harry, ListGrid, I imagine this would take two props:

```
   items T[]
   renderItem(item: T)
```

And then OrganizationListGrid would *use* ListGrid and implement the organization-specific rendering function.

You then can have all of the generic styling / listgrid functionality in `ListGrid` which gets re-used and then the organization-specific styling and functionality in OrganizationListGrid.

slifty (approved):
Woohoo! The code looks great -- there are two comments from @reefdog that I can't tell if they were addressed, but I defer to you on whether they're important.

Issue #498, opened 2024-01-09
Explore switching from Create React App to Vite for the front-end build framework


slifty:
We're using create react app for our build system; it is no longer maintained.

Vite would be a good replacement.

reefdog:
One consequence of being stuck on create-react-app: https://github.com/PhilanthropyDataCommons/front-end/pull/626#issuecomment-1981030769

hminsky2002:
I shall note that simply uninstalling react-scripts, installing the vite package, and moving the index.html file to the root directory does not work. Womp womp.

reefdog:
> I shall note that simply uninstalling react-scripts, installing the vite package, and moving the index.html file to the root directory does not work. Womp womp.

https://getyarn.io/yarn-clip/65f69bb2-1a3a-40b2-97c8-9b456c06ff82

reefdog:
We can call this task complete as we are now [actively moving to Vite](https://github.com/PhilanthropyDataCommons/front-end/issues/720).

PR #625, opened 2024-03-05
Group eslint dependencies in dependabot

slifty:
This PR adds an eslint grouping in dependabot so eslint related packages will get updated at the same time.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -24,6 +24,9 @@ updates:
           - 'react-dom'
           - '@types/react'
           - '@types/react-dom'
+      eslint:
+        patterns:
+          - '*eslint*'
 
   - package-ecosystem: 'github-actions'
     directory: '.github/workflows'


slifty:
I opened it because the eslint-typescript library recently updated and requires eslint 8, which means the PR breaks (#611).
 
I thought "oh clearly there must be an eslint 8 PR somewhere else, and we need these grouped for the PRs to not break" but then realized that no, it's locked in with react app!

Which is to say, this PR will be more useful after implementing #498.  It isn't causing any harm until that point, however, and so I'd suggest we should merge it regardless.

.github/dependabot.yml
```diff
@@ -24,6 +24,9 @@ updates:
           - 'react-dom'
           - '@types/react'
           - '@types/react-dom'
+      eslint:
+        patterns:
+          - '*eslint*'
```
reefdog, line 29: This worries me as potentially too broad. Can we just make it `eslint*` with the same practical results and less risk of sweeping up something unrelated?

slifty: I copied the decisions made in the `service` repo and don't think it makes sense to have different rules for the same intention.

See the conversation around this change from there: https://github.com/PhilanthropyDataCommons/service/pull/592#discussion_r1393432857

What do you think?

reefdog: Perfect! I was going to ask if it was too much trouble to do precisely that check. (I do still think `eslint*` is better than `*eslint*`, but we're arguing over trivial risk here. Still, that seems to be what you actually searched for, right?)

slifty: I don't remember (I think that's a typo and I checked for both) -- but keep in mind there are eslint packages that definitely *do not* start with eslint (e.g. the typescript-namespaced eslint packages) https://github.com/PhilanthropyDataCommons/service/blob/4e0aacb74207eecdf15c6b7869d379ca41ad17c0/package.json#L54

We'll be using those when we pull out of react app, so I do think there is a practical reason to include it.

hminsky2002 (approved):
Seems as though this has been worked out!

slifty:
Let the record show that JUSTIN I DO NOT NEED YOUR APPROVAL.

PR #629, opened 2024-03-08
Add Organization Table Pages

### LINKED ISSUE #596: Create organization table
We need to add an organization table to supplement the current proposals table (which is mapped to the Dashboard navbar item but routes to `/proposals`). It will live at `/organizations`, but for the moment it shouldn't be linked from the navbar; you'll just have to know the URL exists. It will be quite empty, just two columns (name and EIN).

Its interface/layout will be pretty much exactly the same as the proposals table (`/proposals`, linked from the navbar as Dashboard for now), except with different columns and no Toggle Wrapping button/feature.

Clicking on each row will go to the associated `/organizations/{id}` detail page. (That… probably argues for building the detail page first.)

The data structure for this table should be significantly different — much lighter — than the proposals table. A proposal is composed from a deep structure of base fields, mapped through proposal versions to application form fields. I believe this table only needs to read the `name` and employer_identification_number` fields directly off the Organization entity for now. The `/organizations` API endpoint is currently in an [open PR](https://github.com/PhilanthropyDataCommons/service/pull/794).

One big difference from `/proposals`: this page will be public, no auth required.

So essentially: this page, but routed to `/organizations`, populated from the `/organizations` API endpoint, with only columns (Organization Name and Organization EIN), and sans "Toggle Wrapping":

Dashboard

hminsky2002:
This PR adds a view for the Organizations table. It is similar in structure to the current view for the '/proposals' route, but does not employ base-fields. 

### Testing
- run `npm run start` in the front-end repository
- run `npm run start` in the service repository
- populate organizations table
- on the front-end in your browser, go to `http://localhost:3000:organizations`
- click on an organization and verify that it redirects to the chosen organization's detail page

Closes #596


src/pages/OrganizationList.tsx
```diff
@@ -0,0 +1,41 @@
...
+} from '../pdc-api';
+
+const OrganizationListLoader = () => {
+	const [organizations] = useOrganizations(
+		ORGANIZATIONS_DEFAULT_PAGE,
+		ORGANIZATIONS_DEFAULT_COUNT,
+	);
+
+	useEffect(() => {
+		document.title = 'Organization List - Philanthropy Data Commons';
```
slifty, line 18: I'd call this "Organizations" rather than "Organization List" on user-facing things.

src/components/OrganizationListTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+		</TableRow>
+	);
+};
+
+interface OrganizationListTableProps {
+	columns: string[];
+	organizations: OrganizationBundle;
+}
+
+export const OrganizationListTable = ({
```
slifty, line 41: Please don't boo me off the stage BUT... should we make a generic `ListTable` similar to what we did with the `ListGrid`?  I think this will become more obviously useful when we add base fields to the organizations entity but even now I think there is some rendering / UX logic that is being duplicated.

src/components/OrganizationListTablePanel.tsx
```diff
@@ -0,0 +1,36 @@
...
+	organizations = undefined,
+	loading = false,
+}: OrganizationListTablePanelProps) => {
+	const generateFallbackMessage = () =>
+		loading ? 'Loading…' : 'No data available.';
+
+	return (
+		<Panel>
+			<PanelBody
+				padded={organizations !== undefined && organizations.entries.length > 0}
```
slifty, line 23: Could you put this in a named utility method, that way it's clear from reading what it is checking for and why (this will also mean you can remove that portion of the commit message!)

Maybe something like `hasOrganizations()`

src/components/OrganizationListTablePanel.tsx
```diff
@@ -0,0 +1,36 @@
...
+import { OrganizationBundle } from '@pdc/sdk';
+import { Panel, PanelBody } from './Panel';
+import { OrganizationListTable } from './OrganizationListTable';
+
+interface OrganizationListTablePanelProps {
+	organizations?: OrganizationBundle;
+	loading?: boolean;
+}
+
+const DEFAULT_COLUMNS = ['Organization Name', 'Organization EIN'];
```
slifty, line 11: I think we can take out `Organization` from both column names (but at least the `EIN` column name)

src/components/OrganizationListTablePanel.tsx
```diff
@@ -0,0 +1,36 @@
...
+}: OrganizationListTablePanelProps) => {
+	const generateFallbackMessage = () =>
+		loading ? 'Loading…' : 'No data available.';
+
+	return (
+		<Panel>
+			<PanelBody
+				padded={organizations !== undefined && organizations.entries.length > 0}
+			>
+				{organizations !== undefined && organizations.entries.length > 0 ? (
```
slifty, line 25: Using the named utility function will also allow you to avoid this redundant code, since you can use it in both places.

slifty (changes requested):
Some small suggested changes, and then one larger "idea" if you're up for it.

src/components/OrganizationListTable.tsx
hminsky2002: Boo! Boo! No, i'm kidding. This does seem like the right idea, let me see how feasible it is really quickly.

hminsky2002:
@slifty I have refactored the code to introduce a new abstract ListTable component following the conventions from PR #616. Furthermore, it's been rebased to the main branch, and your other comments have been implemented!

src/components/OrganizationListTable.tsx
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const OrganizationListTableRow = ({
+	organization,
+}: OrganizationListTableRowProps) => {
+	const navigate = useNavigate();
+
+	const handleRowClick = () => {
+		// Only navigate if we didn't just finish selecting text.
+		if (window.getSelection()?.isCollapsed) {
```
slifty, line 18: A COMMENT?!?!!  

Just kidding -- but also, if we put this check in a named function could we eliminate the comment?  For instance something like `userHasNotSelectedText()`

That way the code itself reads "if the user has not selected text, then navigate" and you don't need the comment 🎉

PR #630, opened 2024-03-09
Implement proposal UI changes

hminsky2002:
This PR updates the UI for the ProposalDetail page and corresponding ProposalDetailPanel and ProposalGridList components, such that the proposal title is displayed first followed by organization name, and the data provider panel is removed. Furthermore, it adds default behavior to the proposal title such that if the proposal_name base field is empty,
it will use a truncated value from proposal_summary, and if both are empty a placeholder string is used.

Testing
- run `npm run start` in the front-end repository
- run `npm run start` in the service repository
- Create and upload three proposals with the following fields defined:
   ```csv
   proposal_name,proposal_summary,organization_name,proposal_submitter_email,organization_tax_id,organization_city,organization_state_province,organization_country
   ```
   ```csv
   proposal_summary,organization_name,proposal_submitter_email,organization_tax_id,organization_city,organization_state_province,organization_country
   ```
   ```csv
   organization_name,proposal_submitter_email,organization_tax_id,organization_city,organization_state_province,organization_country
   ```
- on the front-end in your browser, go to http://localhost:3000:proposals and click on one of the proposals
- verify the title and layout as accurate for each proposal

Closes #602

Issue #601, opened 2024-02-28
Design UI changes for proposal detail page (in an org world)


reefdog:
The current proposal detail page sort of pretends to be an org detail page. Once we _actually_ have an org detail page (#598), we will want to make several changes to the proposal detail page. Capture those changes here.

reefdog:
![Proposal](https://github.com/PhilanthropyDataCommons/front-end/assets/4731/cbd8bdc4-e959-438d-874f-e995a00cc33f)

**Changes from current:**
- Sidebar data order should be: proposal title, then org name, then location
- Main panel title should be proposal title
- Main panel tags should be org name, then EIN; remove proposal title tag
- Remove Data Providers button

Note that not all proposals have titles. We should fall back to truncated proposal summary, then… "Untitled Proposal" until testing reveals a better name.

PR #658, opened 2024-04-02
Implement proposal detail page UI changes

### LINKED ISSUE #602: Implement UI changes for org-world proposal detail page
Implement the UI changes captured in #601.

reefdog:
_With gratitude, this PR replaces #630. I realized while reviewing @hminsky2002's efficient work that some larger refactoring was justified, as well as a feature I _didn't_ mention in the original issue: moving Data Provider buttons to the org page rather than removing it altogether._

This PR implements the desired UI changes for the proposal page now that we have proper organization pages.

- Sidebar and header are now proposal-centric instead of org-centric
- Data Providers button (Candid / Charity Navigator) moved to org page
- `ListGridItem` component added

**Testing:**
1. On `main`, go to any proposal page (e.g. http://localhost:3000/proposals/464)
2. Note the contents of the sidebar and panel header
3. Switch to this branch and reload that page
4. Note that the sidebar and panel header are quite different! (See #601 for the specific differences.) Rather than elevating organization-related fields in order to make the page look like an org page, we now elevate proposal-specific fields (name or summary). We've also moved the Data Providers button to the org page. Speaking of which…
5. Go to http://localhost:3000/organizations. Assuming you've added some, if you load an org page, you'll see the Data Providers button in the top right of the panel.

Closes #602


src/stories/ListGrid.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+		location: 'Arizona',
+		size: 'Small',
+		weight: '50lbs',
+	},
+];
+
+export const Default: Story = {
+	args: {
+		items: sampleGridItems,
+		renderItem: (item: SampleGridItem) => (
```
slifty, line 67: Hmmm, do you think there's any way for this to automatically detect that `sampleGridItems` is of type `SampleGridItem` so that you don't have to explicitly specify the type?

reefdog: I have not yet begun to "explicitly" specify the type, just you wait.

src/components/ListGrid.tsx
hminsky2002, line 1: Extremely elegant! When I was making the ListGrid component it felt weird to define the listgriditem twice but I couldn't grasp a way to abstract it because of the differences between the proposals and organizations. Very cool!

src/utils/proposals.ts
```diff
@@ -1,5 +1,29 @@
 import { FrontEndProposal } from '../interfaces/FrontEndProposal';
 
+/**
+ * Iterates through a list of fields on the proposal,
+ * in preferential order (most ot least), looking for desired data.
```
hminsky2002, line 5: typo in the comment here, i imagine it's supposed to say "most or least"?

hminsky2002 (approved):
Looks great! Much more sensible to abstract the ListGridItem i think. Tested locally and everything is functional and fluid. The only thing I could find was a typo in a comment, but I'm gonna approve this!

src/components/ListGrid.tsx
reefdog: Thanks!

I'm a little worried about some of the implied internal structure of `ListGridItemDetails`, which require understanding how CSS flexbox behaves to understand why you have to wrap multiple children in elements. I realize that's not documented well enough… _and_ just discovered that I duplicated some of the comments unnecessarily. Will fix!

src/utils/proposals.ts
reefdog: Ah, good eye! I meant `most to least`, and will fix.

PR #659, opened 2024-04-02
Migrate to Storybook v8

### LINKED ISSUE #645: Migrate Storybook to v8
Storybook has been upgraded to v8, which has a few [breaking changes](https://storybook.js.org/docs/migration-guide#major-breaking-changes). We need to run through the [migration guide](https://storybook.js.org/docs/migration-guide).

https://github.com/PhilanthropyDataCommons/front-end/pull/657

reefdog:
This PR upgrades us to Storybook v8, including some required configuration changes from the migration.

**Testing on deploy:**
1. Go to https://deploy-preview-659--philanthropy-data-commons-viewer.netlify.app/storybook/?path=/docs/introduction--docs
2. Poke around!

**Testing locally:**
1. `npm i`
4. `npm run storybook`
5. Go to http://localhost:6006 in your browser and poke around

Closes #645


.storybook/preview.tsx
```diff
@@ -11,7 +11,6 @@ const preview: Preview = {
 		backgrounds: {
 			default: 'light',
 		},
-		actions: { argTypesRegex: '^on[A-Z].*' },
```
hminsky2002, line 14: Everything else looks great, was wondering why we are removing this line? I didn't see anything in the v8 docs about this keyword being deprecated so just wanted to check!

reefdog: Good eye!

That line was boilerplate from the initial setup, and is for a feature we aren't using ([Actions](https://storybook.js.org/docs/essentials/actions)). The v8 migration script flagged it for suggested removal, since it's no longer the [recommended](https://storybook.js.org/docs/essentials/actions#via-storybooktest-fn-spy-function) way of mocking component methods. It's still _supported_, but not recommended:

> However, [[automatically matching args](https://storybook.js.org/docs/essentials/actions#automatically-matching-args)] is not the recommended way of writing actions. That's because automatically inferred args are not available as spies in your play function. If you use argTypesRegex and your stories have play functions, you will need to also define args with the fn utility to test them in your play function.

hminsky2002: Roger that! Sounds like it's approval time 😎

hminsky2002 (approved):

PR #661, opened 2024-04-02
Improve navbar icons and order

### LINKED ISSUE #606: Change Developers/About menu order

### LINKED ISSUE #660: Toggle between outline and solid icons in navbar
Right now, icons in the navbar are a bit chaotic. We use outlined variants for some and solid variants for others:

Currently chaotic

We should instead consistently use outline for inactive icons and solid for active/activated, e.g.:

Consistent

Consistent (open dropdown)

reefdog:
This PR changes our navbar icons to be outlines when inactive and solid when active. It also moves the Developer menu to "last" (farthest right) on the navbar, except that the user menu remains the _farthest_ right (as per standard design).

**Before:** Logged out, on the Landing page:
1-before-loggedout

**After:** Logged out, on the Landing page:
2-after-logged-out

**Before:** Logged in, on the Dashboard page:
1-before

**After:** Logged in, on the Dashboard page:
2-after

**Testing:**

Navbar icons should be outlined by default. They should turn solid when:

- **Dashboard** and **Add Data**: solid when you are actually on those page
  - ✋ These will also be blue
- **About** dropdown, **Developers** dropdown: solid when the menu is open
- **User** icon: solid when you are logged in

Closes #606
Closes #660

### DIFF
--- a/src/components/AppNavbar.tsx
+++ b/src/components/AppNavbar.tsx
@@ -56,39 +56,6 @@ const AppNavbar = () => {
 				<li>
 					<Dropdown>
 						<DropdownTrigger type="navbar-item">
-							<CommandLineIconOutline className="closed-only" />
-							<CommandLineIconSolid className="open-only" />
-							Developers
-						</DropdownTrigger>
-						<DropdownMenu align="right">
-							<DropdownMenuLink to="https://api.philanthropydatacommons.org">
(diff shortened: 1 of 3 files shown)


hminsky2002 (approved):
Looks good, I think this is a sensible change!

PR #674, opened 2024-04-10
Fail eslint on warning

### LINKED ISSUE #523: Lint warnings should be a showstopper
As noted by @slifty, our ESLint warnings aren't currently causing CI to block merge, but probably should!

slifty:
This PR updates our lint command so that it will return an error code if there are any warnings.  Prior to this commit our CI would pass even if there were warnings (e.g. console.log statements)

Resolves #523


reefdog (approved):
Makes sense, it works, as much as it pains me to approve of something you did, here you go.

Issue #618, opened 2024-03-05
Add user menu with link to account console


reefdog:
Currently, when the user is logged in, their name shows in the navbar, but it's dead text.

This should be a menu that spawns a link to the account console (TBD https://github.com/PhilanthropyDataCommons/auth/issues/32) and a log out button (TBD in forthcoming issues).

For this issue, let's just add the first one, along with some explanation text about the whole Keycloak thing, since it's not exactly a frictionless UX.

reefdog:
Well, turns out that logging out came first. See #696. So this menu exists and will just need the Account Settings work.

PR #676, opened 2024-04-10
Improve display of navbar dropdowns

### LINKED ISSUE #667: Change formatting of navbar dropdowns
Tinkering with a user menu for #618 and #619, I realized that… sigh… @slifty was right when he said that our navbar dropdowns don't need the "small all caps title" styling that I gave them, and once we have the user menu (where I definitely prefer *not* to have that styling), it looks weird going back and forth.

So let's turn these:
Developer menu

Into something more like these:
User menu

reefdog:
This PR improves the look of our navbar dropdowns in preparation for the user menu dropdown (#618).

Before:
dropdown-before

After:
dropdown-after

<details>
<summary>If you think that's a stylistic regression…</summary>

Maybe some of you will think "Ah, but, that's less attractive!" And you're not wrong, but, the new format looks better when interleaved with more traditional dropdown items (which we don't want to always force into the "small uppercase" format):

https://github.com/PhilanthropyDataCommons/front-end/assets/4731/7097f3dc-e41b-4ed1-95e1-9700e8c5ccef

</details>

Along for the ride come some text accessibility improvements and an improvement to the `:focus` outline when keyboard-navigating the menu.

**Testing:**

You can test this via the [deploy preview](https://deploy-preview-676--philanthropy-data-commons-viewer.netlify.app), since the affected menus are available while logged-out. You can also hit the [Storybook deployment](https://deploy-preview-676--philanthropy-data-commons-viewer.netlify.app/storybook/?path=/docs/stories-dropdown--docs) to see the updated Dropdown stories. Just make sure dropdowns are legible and not a regression from earlier forms.

Resolves #667

### DIFF
--- a/src/components/Dropdown/Dropdown.css
+++ b/src/components/Dropdown/Dropdown.css
@@ -47,9 +47,9 @@
 	width: var(--dropdown--menu--width);
 
 	background-color: white;
-	border-radius: 8px;
+	border-radius: var(--dropdown--menu--border-radius);
 	border: 1px solid var(--color--gray--medium);
-	overflow: hidden;
+	overflow: visible;
 
 	display: flex;
 	flex-direction: column;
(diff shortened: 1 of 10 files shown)


src/components/Dropdown/Dropdown.css
```diff
@@ -67,8 +67,15 @@
 	border-bottom: 1px solid var(--color--gray--light);
 }
 
+.dropdown-menu > *:first-child {
+	border-top-left-radius: var(--dropdown--menu--border-radius);
```
hminsky2002, line 71: I did not know that this was a thing you could set with css, very cool!

hminsky2002 (approved):
This looks idiomatically appropriate, and style wise looks great. Approved from me!

src/components/Dropdown/Dropdown.css
reefdog: Do you mean CSS variables, or setting the border radius?

PR #680, opened 2024-04-11
Limit navbar to only one open dropdown at a time

### LINKED ISSUE #679: Only one navbar dropdown should be open at once
Using the keyboard, it's possible to open multiple navbar dropdowns simultaneously:

https://github.com/PhilanthropyDataCommons/front-end/assets/4731/5c5f82da-abda-4487-a842-4451debd6e5a

This shouldn't be possible.

reefdog:
Previously, you could "trick" the navbar into having multiple dropdowns open at a once by keyboard-navigating from an open menu to an adjacent trigger. (See video in #679.)

Since our Dropdowns use the HTML `<details>` element behind the scenes, and that element supports a [`name` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details#name) precisely for adding accordion-like "just one open item at a time" behavior, this PR adds our own `name` prop to the `Dropdown` component to support the same thing.

**Testing:**

Preface: Firefox doesn't yet support this attribute, so you'll need to test this in Safari or Chrome. Also, the only way to really trigger this "bug" (and thus the only way to confirm it's fixed) is if your browser/OS combo allow tabbing through links. For macOS, this is disabled by default. Enable it in _System Settings_ → _Keyboard_ → _Keyboard navigation_.

1. Go to the `main`/[production site](https://app.philanthropydatacommons.org/)
2. Open the About menu in the navbar
3. Use the Tab key to tab through the items until the Developers navbar item is highlighted
4. Hit space/return to open it
5. :x: Note that both menus are open
6. Switch to the [deploy preview](https://deploy-preview-680--philanthropy-data-commons-viewer.netlify.app/) and repeat the steps
7. ✅ Note that About menu closes when the Developers menu opens

Resolves #679

### DIFF
--- a/src/components/AppNavbar.tsx
+++ b/src/components/AppNavbar.tsx
@@ -55,7 +55,7 @@ const AppNavbar = () => {
 					</li>
 				)}
 				<li>
-					<Dropdown>
+					<Dropdown name="navbar-dropdown">
 						<DropdownTrigger type="navbar-item">
 							<InformationCircleIconOutline className="closed-only" />
 							<InformationCircleIconSolid className="open-only" />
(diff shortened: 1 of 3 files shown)


hminsky2002 (approved):
Looks great to me!

PR #689, opened 2024-04-15
Fix display issues with organizations tables

### LINKED ISSUE #688: Organization table isn't padded correctly
The organization table isn't supposed to have padding around it, but it does:

Org table

reefdog:
Prior to this PR, the org table was incorrectly padded.

This PR fixes the padding, and resolves a display issue that is only noticeable once the padding was fixed.

**Testing:**
- In production, to go https://app.philanthropydatacommons.org/organizations
- ❌ Note the lack of padding around the "No data available" message
- On a local `main` with organizations data, go to `/organizations`
- ❌ Note the padding around the table
- Now, on this branch, repeat the above — well, however you can to trigger both states
- ✅ Note that the "No data available" message has padding, while the table doesn't

Closes #688

### DIFF
--- a/src/components/Table/Table.css
+++ b/src/components/Table/Table.css
@@ -24,6 +24,14 @@
 	top: calc(-1 * var(--table--border-width));
 }
 
+.panel > .panel-body:first-child > .table:first-child thead.fixed {
+	top: 0;
+}
+
+.panel > .panel-body:first-child > .table:first-child thead.fixed th {
+	border-top: 0;
+}
+
 .table thead th {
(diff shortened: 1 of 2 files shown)


src/components/Table/Table.css
```diff
@@ -24,6 +24,14 @@
 	top: calc(-1 * var(--table--border-width));
 }
 
+.panel > .panel-body:first-child > .table:first-child thead.fixed {
```
hminsky2002, line 27: Neat! Checked on my local branch and it was working, approved!

hminsky2002 (approved):
This was a blunder of mine own creation, thank you for fixing it!

PR #694, opened 2024-04-16
Relocate page security to router

### LINKED ISSUE #693: Secure pages at route definition
Previously, we intentionally decided not to secure page components in the router. See conversation here: https://github.com/PhilanthropyDataCommons/front-end/pull/106#discussion_r1186069693

However, the math on that has changed, particularly because of #566. That page has multiple URLs that lead to the same page component. We want to not require authentication for one of the routes (`/organizations/{id}`), but we _do_ want to authenticate its `provider/` and `proposal/` sub-routes. The best way to do that is in the router definition rather than within the page component itself.

While we can solve this for just this page without affecting the others, that would mean route protection was defined multiple ways — in the routes file for some pages (or versions of pages), in the page component for others.

Let's choose consistency and move all page securing to the routes file. We can still protect individual components by wrapping them in `withOidcSecure` (or `<OidcSecure>`) as desired.

reefdog:
This PR moves securing pages from the components themselves to the app router. There should be no functional changes from the current system, it's just prerequisite work for the public pages — but a significant enough change to justify its own PR and review.

This is a syntax [explicitly demonstrated by our authentication library](https://github.com/AxaFrance/oidc-client/tree/main/packages/react-oidc#how-to-secure-a-component--hoc-method), which we did originally [consider but discarded](https://github.com/PhilanthropyDataCommons/front-end/pull/106#discussion_r1186521329) for reasons that I _think_ no longer apply. (We weren't using `<RouteProvider>` at the time.)

This will let us selectively secure routes that lead to the same page component that unprotected routes lead to. Specifically, we will [later, not in this PR] want public users to be able to access `/organizations/{id}` but not `/organizations/2/proposals/100` or `/organizations/2/providers/candid`. This will now be possible, even though both lead to the same page component. (Within that page component, we will still conditionally secure some child components, which we planned to do anyway.)

We're getting into the weeds outside my area of expertise, so please weigh in if this is a Bad Idea™!

**Testing:**

Test everything around authentication: accessing pages while logged out should force you to authenticate. After authenticating, you should be able to access all pages, etc. This should have zero behavioral/functional changes from the current system.

Resolves #693

### DIFF
--- a/src/pages/AddData.tsx
+++ b/src/pages/AddData.tsx
@@ -1,6 +1,5 @@
 import React, { useEffect, useState } from 'react';
 import { NavLink } from 'react-router-dom';
-import { withOidcSecure } from '@axa-fr/react-oidc';
 import { DocumentPlusIcon } from '@heroicons/react/24/outline';
 import {
 	uploadUsingPresignedPost,
@@ -101,5 +100,4 @@ const AddDataLoader = () => {
 	);
 };
 
-const AddData = withOidcSecure(AddDataLoader);
-export { AddData };
+export { AddDataLoader as AddData };
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { createBrowserRouter, RouterProvider } from 'react-router-dom';
+import { withOidcSecure } from '@axa-fr/react-oidc';
 import { Layout } from './Layout';
 import { AddData } from './pages/AddData';
 import { NotFound } from './pages/NotFound';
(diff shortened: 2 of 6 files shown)


src/pages/AddData.tsx
```diff
@@ -101,5 +100,4 @@ const AddDataLoader = () => {
 	);
 };
 
-const AddData = withOidcSecure(AddDataLoader);
-export { AddData };
+export { AddDataLoader as AddData };
```
reefdog, line 103: I had two choices:

1. Keep the external surface area of these files the same by rewriting the exports like this
2. Export the `Loader`-suffixed versions and change all the imports

I chose this lighter-touch version. @jasonaowen and I had a very long discussion [here](https://github.com/PhilanthropyDataCommons/front-end/pull/106#discussion_r1186069693) about the `Loader` naming convention, which neither of us _love_ but which we both decided wasn't worth belaboring at the time.

I will note that we now stepped straight into this concern of Jason's:

> One of the reasons we have the no-default-export rule enabled is to avoid that kind of acrobatics, because calling something one name in one place and another name in another place makes it more difficult to understand the code.

However, this part is no longer the case:

> And it's not exactly that it's insecure, it's that it's meant to be used inside a secured component.

Because we actually use some of these `Loader`s on public pages. And we we've been inconsistent with our use of `Loader`s anyway.

Really the whole component naming convention is due a rethink in light of our current complexity, but I think we should punt on that until Phase 3.

slifty (approved):
This is a bit outside of my expertise as well so take my approval with a grain of salt, but this change makes sense to me!

PR #695, opened 2024-04-16
Add public org views

### LINKED ISSUE #560: Allow unauthenticated users to view organizations table
User story: https://github.com/PhilanthropyDataCommons/meta/issues/7
API dependency: https://github.com/PhilanthropyDataCommons/service/issues/912

To support public users viewing limited data, we will open the organizations table up to public access.

For now, this will just render the attributes on the Organization model itself, which are de facto public. Once we have Organization base fields, it will be the API's responsibility to choose (based on user -- or lackthereof) what fields to return. In views like this table, where we compose the columns from a template, we'll want some way of hiding the columns that the user _can't_ see; the data will just be missing from the API response, but it's not ideal to have columns that will _always_ be empty for the user.

### LINKED ISSUE #566: Allow unauthenticated users to view organization detail page
As a companion to #560, public visitors should be able to drill into organization detail pages.

- [x] Allow public visitors to load the page @reefdog 
- [ ] Only fetch/render proposals for authenticated users — hide entire section for public visitors @hminsky2002 
- [ ] Only fetch/render provider data for authenticated users — hide buttons for public visitors @hminsky2002 

API dependency: https://github.com/PhilanthropyDataCommons/service/issues/912

### LINKED ISSUE #571: Add "you're seeing limited data" banner to public views
In #562, I presented some options for drawing user attention to how we communicate to unauthenticated/public visitors that they're only seeing a subset of pages. After some lively debate on Jitsi, we settled on a standard fixed banner atop the site (except on the landing page; that is, only once you're "inside" the site):

Text banner

This banner is low-impact enough that we don't need to allow visitors to dismiss it. (Other reasons: visitors might forget they've done so; it's a good reminder to people shoulder-surfing the site or looking at screenshots; adding dismissability is more work.)

I'll refine the design and language during implementation, but the starting language will be:

> You're only viewing limited public data. [Log in to view the full dataset.] (Need an account? [Email Jim McGowan to request access].)

reefdog:
This PR adds publicly-accessible organization views (list and detail pages). It customizes the detail pages so that parts of the page that public visitors shouldn't see or use — data provider buttons/panels, proposal table/panels — are hidden. It also adds a banner atop the site that alerts public visitors that they may only be seeing a subset of the available data. This banner doesn't appear on the home page, just internal pages.

There are also a couple of minor clean-up commits to related code that are minor (and related) enough that I'd like to include them.

**Testing:**
1. On `main`/production, try to access `/organizations` or an organization detail page while logged out.
2. ❌ Note that you are redirected to the login form.
3. On this branch, do the same.
4. ✅ Note that despite not being logged-in, you are able to view the pages. 🎉
5. Verify that the public banner only appears for logged-out users on those pages (and not the home page).
6. For sanity, verify that you are still unable to access `/proposals`, the proposal detail pages, the Add Data page, and the two organization detail sub-pages (`/organizations/{id}/providers/{provider}` and `/organizations/{id}/proposals/{proposalId}`) while logged-out.

Resolves #571 
Resolves #566 
Resolves #560

### DIFF
--- a/src/components/AppBanner.tsx
+++ b/src/components/AppBanner.tsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import { useLocation } from 'react-router-dom';
+import { useOidc } from '@axa-fr/react-oidc';
+import { Button } from './Button';
+import { getLogger } from '../logger';
+
+const logger = getLogger('<AppBanner>');
+
+const AppBanner = () => {
+	const location = useLocation();
+	const { isAuthenticated, login } = useOidc();
+
+	const isHomePage = location.pathname === '/';
+	const showBanner = !isAuthenticated && !isHomePage;
+
+	return showBanner ? (
+		<div className="App-banner">
+			You are viewing limited public data.{' '}
+			<Button
+				linkStyle
+				onClick={() => {
+					login(location.pathname).catch(logger.error);
+				}}
+			>
+				Log in to view more.
+			</Button>
+		</div>
+	) : null;
+};
+
+export { AppBanner };
--- a/src/App.css
+++ b/src/App.css
@@ -1,16 +1,35 @@
+:root {
+	--app-banner--space-within: var(--fixed-spacing--halfx)
+		var(--fixed-spacing--1x);
+	--app-header--space-around: var(--fixed-spacing--2x) var(--fixed-spacing--2x)
+		var(--fixed-spacing--1x);
+	--app-main--space-around: var(--fixed-spacing--1x);
+}
+
 .App {
 	position: absolute;
 	inset: 0;
 
-	padding: var(--fixed-spacing--1x);
-
 	display: grid;
-	grid: min-content 1fr / auto;
-	gap: var(--fixed-spacing--1x);
+	grid-template:
+		'banner' min-content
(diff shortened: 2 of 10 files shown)


src/components/AppBanner.tsx
```diff
@@ -0,0 +1,31 @@
...
+const AppBanner = () => {
+	const location = useLocation();
+	const { isAuthenticated, login } = useOidc();
+
+	const isHomePage = location.pathname === '/';
+	const showBanner = !isAuthenticated && !isHomePage;
+
+	return showBanner ? (
+		<div className="App-banner">
+			You are viewing limited public data.{' '}
```
hminsky2002, line 18: Love how this looks on local!

hminsky2002 (approved):
This all looks good to me, thanks for fixing some of my blunders(specifically needless panel grid item). I think @slifty 's opinion is probably a good thing to have considering we were both working on this but approved from me!

slifty (approved):
for the record, I only approve of @hminsky2002's work

reefdog:
@hminsky2002 Quick shuffle the commits and rebase the whole branch

PR #696, opened 2024-04-19
Add user logout

### LINKED ISSUE #619: Add ability to log out from the front end
We need to be able to log out from the front end. I'm leaving this intentionally vague because I don't know if this is a route, a form POST, or what, but it'll be triggered by a "Logout" button in the user menu (#618) once that exists.

reefdog:
This PR adds a user menu with a lone "Log out" link. That part was simple, but there were three prerequisite changes:

- Refactoring our `Button` component to extend native buttons for a cleaner prop definition
- Cleaning up our focus styling on interactive elements
- Adding a new `DropdownMenuButton` component so that dropdowns can contain links _or_ buttons

**Testing:**
1. Log in
2. Click your username in the top right and then the "Log out" button
3. ✅ You should be logged out and sent to the home page
4. Click "Log in"
5. ✅ You should be prompted to log in; if you aren't, and you're just round-tripped back to the PDC with a valid session, ❌ that's bad because it means we didn't kill the Keycloak session.

Resolves #619

### DIFF
--- a/src/components/User.tsx
+++ b/src/components/User.tsx
@@ -1,14 +1,23 @@
 import React from 'react';
 import { useOidc, useOidcUser, OidcUserStatus } from '@axa-fr/react-oidc';
-import { UserIcon as UserIconOutline } from '@heroicons/react/24/outline';
+import {
+	PowerIcon,
+	UserIcon as UserIconOutline,
+} from '@heroicons/react/24/outline';
 import { UserIcon as UserIconSolid } from '@heroicons/react/24/solid';
+import {
+	Dropdown,
+	DropdownTrigger,
+	DropdownMenu,
+	DropdownMenuButton,
+} from './Dropdown';
 import { getLogger } from '../logger';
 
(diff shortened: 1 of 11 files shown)


src/components/User.tsx
```diff
@@ -40,10 +49,24 @@ const User = () => {
...
+					<DropdownTrigger type="navbar-item">
+						<UserIconSolid />
+						{oidcUser.name}
+					</DropdownTrigger>
+					<DropdownMenu align="right">
+						<DropdownMenuButton
+							icon={<PowerIcon />}
+							className="logout"
+							onClick={() => {
+								logout('/').catch(logger.error);
```
hminsky2002, line 62: This is great, works exactly as expected on my local branch, whatever page I may be on when I click logout i get redirected to the login screen

hminsky2002 (approved):
This seems like the exact right implementation of logout, and the design choices are all clear and followable. Approved from me!

Issue #622, opened 2024-03-05
Implement ability to filter proposals by "mine"


reefdog:
_Implementation task for #620, details TBD once design is complete._

reefdog:
As with #697, this requires the ability for the front-end to either know the current user ID, or otherwise send a magic keyword to the API to say "give me my stuff". See https://github.com/PhilanthropyDataCommons/service/issues/939.

reefdog:
This is dependent on #622.

slifty:
Per #988 we aren't going to be supporting this functionality.

PR #704, opened 2024-04-23
Convert Proposal's text-wrapper to a ToggleSwitch

### LINKED ISSUE #620: Design ability to filter proposals by "mine"
From the proposal list table, users who have access to proposals beyond just their own should be able to toggle between "all proposals I'm allowed to see" and "only my proposals".

Note: This is currently only an admin need. Unauthenticated visitors won't have permission to see any proposals, and non-admin users will only see their own anyway.

reefdog:
This PR is the output of #620. It adds a new `ToggleSwitch` component in the style of iOS toggle switches (which have since broken out of iOS-land to be a pretty widely-used UI component, e.g. in [Tailwind](https://tailwindui.com/components/application-ui/forms/toggles) or [Bootstrap](https://getbootstrap.com/docs/5.0/forms/checks-radios/#switches)), and it converts our button-style wrapping toggle for the proposals table to use it, with the eventual aim of using it for "Show only my proposals" in #622.

**Before:**

https://github.com/PhilanthropyDataCommons/front-end/assets/4731/64d40b45-1732-4cb0-b5a5-0c8405e50ed8

**After:**

https://github.com/PhilanthropyDataCommons/front-end/assets/4731/76ddc449-067c-4052-a366-b1a60c4aa92b

### Testing

- Examine the [Storybook](https://deploy-preview-704--philanthropy-data-commons-viewer.netlify.app/storybook/?path=/docs/stories-toggleswitch--docs)
- Compare the "Toggle wrapping" functionality on `main`/production's `/proposals` table with this one

Resolves #620 (in that by accepting this change, we're committing to this UI for the proposal-filter too)

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,6 +84,12 @@
 					"optionalDependencies": false
 				}
 			],
+			"jsx-a11y/label-has-associated-control": [
+				"error",
+				{
+					"assert": "either"
+				}
+			],
 			"react/function-component-definition": [
 				"error",
 				{
--- a/src/components/Panel/Panel.css
+++ b/src/components/Panel/Panel.css
@@ -85,6 +85,12 @@
 .panel-actions {
 	display: flex;
 	gap: var(--fixed-spacing--1x);
+	align-items: center;
+
+	&.switches {
+		padding-inline-end: var(--button--padding-x);
+		gap: var(--fixed-spacing--2x);
+	}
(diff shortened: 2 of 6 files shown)


hminsky2002 (approved):
Tested locally and functions as expected, agree with the design choice here! Approved

PR #706, opened 2024-04-24
Make user IDs available for use in dependent API calls

### LINKED ISSUE #702: Get the user's ID for use in dependent API calls
Several of our issues (#697, #622) require the front-end to know the currently-logged-in-user's ID. Not their Keycloak authentication ID, but their actual ID in the PDC which gets passed in to various calls as the `createdBy` param.

(The API may in the future accept our authentication ID — that is, the `sub` value from our JWT token — in lieu of this, but for now we'll need to know the actual numeric ID.)

Once https://github.com/PhilanthropyDataCommons/service/pull/946 is merged in, we (the front end) can get this from `/users?authenticationId={sub}`, where the value of `sub` is available from within a component as:

```js
const { oidcUser } = useOidcUser();
// Sub now vailable in `oidcUser.sub`
```

We don't want to have to fetch this every time we need it for an API call, so what we should do is either:

1. Fetch and cache it during login: this could turn into a yak shave since we need to insert ourselves into [our OIDC client library](https://github.com/AxaFrance/oidc-client)'s `login()` and `logout()` methods somehow.
2. Fetch and cache it the first time we need it: this may be easier since we control the entire mechanism for, well, when we call the API.

In either case we should be sure to remove it during logout.

_This smells like opening ourselves up to a security or cache-invalidation Problem so I do hope the API lets us pass in the `{sub}` soon (🤞 @slifty since you have nothing else to do right?)._

hminsky2002:
**This is a draft PR, it brings with it a pretty significant design choice that I'd like to review before moving forward. ~~Also, it contains an update to the pdc-sdk which should be a separate dependabot PR.~~**

This PR adds an additional context provider to the front-end application that provides the User(retrieved from the API by the subject of the OIDC session in the front-end) as global context to the rest of the application, which can be retrieved via the 'useContext' hook. There is currently no code in the commits that use this context, but it can be tested locally, e.g.
```ts
import React, {useContext} from 'React'
import {UserContext} from '../UserContext'


const MyComponent = () => {
          const user = useContext(UserContext);
          console.log(user) //or any other operation
```

[Context is an officially recommended approach to persisting user data as per the React Docs](https://react.dev/learn/passing-data-deeply-with-context#use-cases-for-context), and furthermore is the approach used, at least in part, by our current auth provider. However, there is something definitely awry about wrapping our application in two different User contexts. 

There is certainly also a viable approach in 'preflight'ing our api calls, but either way (as discussed in my first commit) I believe we will at least need to employ the new 'preload_user_info' feature of the 'react-oidc' package, in order to both acquire the oidcUser subject and the User object from the database. 

 
Closes #702

### DIFF
--- a/src/UserContext.ts
+++ b/src/UserContext.ts
@@ -1,0 +1,10 @@
+import { createContext } from 'react';
+import type { User } from '@pdc/sdk';
+
+const emptyUser = {
+	id: 0,
+	authenticationId: '',
+	createdAt: new Date(0),
+};
+
+export const UserContext = createContext<User>(emptyUser);
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
 			"name": "front-end",
 			"version": "0.1.0",
 			"dependencies": {
-				"@axa-fr/react-oidc": "^7.22.3",
+				"@axa-fr/react-oidc": "^7.22.4",
 				"@fontsource/source-sans-pro": "^5.0.8",
 				"@heroicons/react": "^2.1.3",
 				"@pdc/sdk": "^0.7.1",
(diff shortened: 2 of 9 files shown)


src/UserContext.ts
```diff
@@ -0,0 +1,10 @@
+import { createContext } from 'react';
+import type { User } from '@pdc/sdk';
+
+const emptyUser = {
+	id: 0,
+	authenticationId: '',
+	createdAt: new Date(0),
+};
```
reefdog, line 8: One small change I'd probably make is export this empty user so we can reuse it in App.tsx instead of redeclaring it.

src/App.tsx
```diff
@@ -43,6 +45,26 @@ const router = createBrowserRouter([
...
 
-const App = () => <RouterProvider router={router} />;
+const App = () => {
+	const { oidcUser } = useOidcUser();
+	const emptyUser = {
+		id: 0,
+		authenticationId: '',
+		createdAt: new Date(0),
+	};
+	const [currentUserState, setCurrentUserState] = useState(emptyUser);
```
slifty, line 55: I would rather `currentUserState` be potentially `undefined` than have a placeholder empty user.  Ultimately a user may or may not exist, and using the built in `undefined` instead of a specialized empty user object will make it easier to determine that state down the line.

reefdog: Good point!

slifty:
For context on the closure: we decided we don't need user context right now, and with the addition of `createdBy=me` support we may never need it

reefdog:
Co-sign! But thanks for the exploratory work @hminsky2002! If you think we may use it in the future, or otherwise need to reference it, you can rename the origin branch to `reference/{whatever}`. That way it's still accessible, but namespaced out of view.

slifty:
Agreed, that said I'd encourage that we NOT keep dead branches (I believe that it's all still available via this PR though)

reefdog:
@slifty if you delete the branch (without merging) you can still see the commits attached to the PR??

slifty:
@reefdog I believe so, yes -- it just is no longer part of the actual repository / not available to fetch.

reefdog:
@slifty Let's test! I have the branch saved locally so can delete it from here, we'll see if commits hang around in PR view. Because that would be the best outcome:

1. Closed PR branch deleted
2. PR still has commits linked/available for reference
3. Optionally, we can save the branch on origin under `reference/*` namespace, but only when we're _very_ likely to want the code in the future (and in an easier format than copying from the PR diff), not just for kicks

Issue #697, opened 2024-04-19
Filter bulk uploads to current user


reefdog:
The API will soon change so that super-admin users will see all entities, not just their own. That means in any views where we don't want this behavior, we'll have to add a filter.

For now, bulk uploads should filter by the current user. Maybe in the future we'll have a view where admins can see all bulk uploads, but that's not today, so this filter can just happen invisibly (i.e. no UI control over it).

reefdog:
Note that at the moment, the front-end has no idea what the current user's ID is, so they can't call `createdBy={my user ID}` as needed. I've opened a discussion-that-may-become-a-task in the service repo to address this: https://github.com/PhilanthropyDataCommons/service/issues/939

reefdog:
Delivered by #707.

PR #707, opened 2024-04-24
Only show the user their own bulk uploads

reefdog:
This PR scopes the bulk upload lists to only those created by the requesting user. It does so statically — i.e., without allowing the user to remove the scope and view bulk uploads from other people, even if they're authorized to. This is an intentional front-end UX choice until we build out a fuller interface for admins to see other people’s bulk uploads (if ever needed). It’s not a security concern; the API handles making sure users only see what they’re authorized to.

The `createdBy=me` value is a magic value that tells the API that the scope should be “the user ID of the requesting user”. That value doesn't work _yet_, but will once https://github.com/PhilanthropyDataCommons/service/pull/961 is merged.

**Testing:**
1. On `main`, on a multi-user system where multiple users have added bulk uploads, go to the Add Data page
2. ❌ No matter who you are logged-in as, all bulk uploads are show in the history sidebar
3. Switch to this branch and check again
4. ✅ You should only see your current user's uploads

🤚 This only affects the uploads, not any proposals or other objects inserted by that upload.

Delivers #697

### DIFF
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -176,7 +176,11 @@ interface ApiBulkUploads {
 	total: number;
 }
 
-const useBulkUploads = () => usePdcApi<ApiBulkUploads>('/bulkUploads');
+const useBulkUploads = () =>
+	usePdcApi<ApiBulkUploads>(
+		'/bulkUploads',
+		new URLSearchParams({ createdBy: 'me' }),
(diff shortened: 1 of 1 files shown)


slifty (approved):
This is *just* a code review (not functionality) but it looks good to me!

PR #708, opened 2024-04-25
Make some UI improvements

reefdog:
Two quick UI improvements that fix some rough edges:

- Toggle switches have heavier-weight labels to better distinguish them from content text
- Panel actions remain positioned in the top right of the panel even when the title wraps (with an exception for switches, which I'd like to improve in the future)

**Testing:**
- The "Wrap long text" toggle switch on `/proposals` was normal-weight, now it's heavy
- When drilling into an organization detail page at `/organization/:id`, the Data Provider button wasn't tucked neatly in the top right corner, it was down a bit; now it's neat

No associated issues, these are just little things noticed while testing.

### DIFF
--- a/src/components/Panel/Panel.css
+++ b/src/components/Panel/Panel.css
@@ -17,6 +17,7 @@
 
 	display: flex;
 	justify-content: space-between;
+	align-items: flex-start;
 	gap: var(--fixed-spacing--2x);
 }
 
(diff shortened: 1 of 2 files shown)


slifty (approved):
Code review: looks good!

hminsky2002:
Late but agreed, looks grand!

PR #710, opened 2024-04-25
Switch spaces to tabs in editorconfig

### LINKED ISSUE #709: Update `.editorconfig` with our tabs-over-spaces preference
When we added Prettier and prettified all files in #605, we forgot to update `.editorconfig` to actually tell our editors / IDEs to preemptively format using tabs. Consequently, they obediently continued using spaces, leaving it up to Prettier to fix before commit. Whoops!

We should update this to save ourselves some troubles.

reefdog:
We switched to tabs months ago in #605, but missed this! Consequently, our editors/IDEs have been dutifully spitting out spaces, leaving it up to Prettier to fix before commit. Now, they do the Right Thing by default.

**Testing:**

This test will only be evident if you _don't_ have Prettier formatting on save or some such thing. (Which, is good to have! But would render this comparison invisible.)

- Using an IDE that supports `.editorconfig`, open the front end on `main` and write a few new indented lines of code
- ❌ Note that they're indented with spaces
- Switch to this branch and try again
- ✅ Note that they're indented with tabs

Resolves #709


.editorconfig
```diff
@@ -4,6 +4,6 @@ root = true
 charset = utf-8
 end_of_line = lf
 indent_size = tab
-indent_style = space
+indent_style = tab
```
hminsky2002, line 7: I had not noticed this but now that you mention it this is a pertinent fix. Thanks for catching it!

hminsky2002 (approved):
Space ❌
Tab ✅

Issue #705, opened 2024-04-23
Use types from SDK when available


reefdog:
The PDC SDK provides some of the types that we're manually defining, e.g. our [`ApiProposal`](https://github.com/PhilanthropyDataCommons/front-end/blob/3a3ca478660406aedc416ed183ef2317f555627a/src/pdc-api.ts#L181-L193). These are brittle and in some cases already out of date.

We should rely on the SDK for these, now that we can!

reefdog:
@hminsky2002 I went ahead and got started on this because I realized so much of our queued work requires knowledge of the latest types, and it makes more sense to use the SDK than update our local stand-ins. But I haven't finished yet; if you end up being blocked by this, feel free to pick up the branch (`705-use-sdk-for-interfaces`) and finish it up.

slifty:
@reefdog I should probably bump the SDK version so you have the `Users` type.

slifty:
New SDK version is released!

PR #719, opened 2024-05-01
Use SDK for Interfaces

hminsky2002:
Major credits to @reefdog for the bulk of the work here, git lies and says that I contributed to the first two because I rebased (I think), but the first two commits(BulkUpload and BaseField) were entirely Justin and I only helped finish the final commit(Proposals)

This PR dumps ApiBulkUpload, ApiProposal, and ApiBaseField, three interfaces that were defined and exported from ```pdc-api.ts``` and used throughout the code base and replaces them with their official SDK counterpart. 

Closes Issue #705 

Reviewing Request: Please take a look at the changes to the time component in ```src/components/BulkUploadListItem.tsx```. I'm not sure that coalescing the date values into empty strings when they are undefined is the right approach but I struggled to think of what a desirable placeholder would be.


hminsky2002:
Leaving this as a Draft PR because unsure about the scope of this issue:

@reefdog Is this issue just to adapt the Exported interfaces(ApiBulkUpload, ApiBaseField, ApiProposal)? Or do we want every interface defined in pdc-api to be tossed out and replaced with the pdc/sdk equivalent?

src/pdc-api.ts
```diff
@@ -132,23 +135,9 @@ const uploadUsingPresignedPost = async (
...
-	sourceKey: string;
-	status: string;
-	createdAt: string;
-}
-
 const useRegisterBulkUploadCallback = () => {
-	const api = usePdcCallbackApi<ApiBulkUploadsPostResponse>('/bulkUploads');
-	return (params: ApiBulkUploadsPostRequest) =>
+	const api = usePdcCallbackApi<BulkUpload>('/bulkUploads');
+	return (params: WritableBulkUpload) =>
```
reefdog, line 140: While we do want this change, do we have to make it as part of this same commit? I.e., is the proposal work broken if we don't also convert the writable bulk uploads here? If not, please move this to a separate commit.

reefdog: (It arguably should have been part of my `BulkUpload` type conversion — and if it can be, great, but it's also fine as a separate commit.)

src/components/BulkUploadListItem.tsx
```diff
@@ -6,29 +6,24 @@ import {
...
 
 export const BulkUploadListItem = ({ upload }: BulkUploadListItemProps) => {
 	const { id, fileName, fileSize, status, createdAt } = upload;
 
 	const relativeTimestamp = (
-		<time dateTime={createdAt} title={localizeDateTime(createdAt)}>
-			{relativizeDateTime(createdAt)}
+		<time
+			dateTime={createdAt?.toString()}
+			title={localizeDateTime(createdAt ?? '')}
```
slifty, line 24: Innnnteresting -- I'm actually surprised that `createdAt` is nullable; it ought not be!

slifty: Looks like several fields are incorrectly marked as optional in our swagger spec -- I'm opening an issue for this.

slifty: Issue opened: https://github.com/PhilanthropyDataCommons/service/issues/987

I would actually wait for this to get merged / the new SDK to be released, since it will mean you don't have to make these checks.

src/map-proposals.ts
```diff
@@ -8,14 +8,22 @@ const extendMultimapReducer = (
 	...multimap,
 });
 
-const mapProposals = (fields: ApiBaseField[], proposals: ApiProposal[]) =>
+const mapProposals = (fields: BaseField[], proposals: Proposal[]) =>
```
slifty, line 11: This doesn't need to be part of this PR (but you might consider making this change in a separate PR before this one)

I don't think you need to do this mapping any longer since base fields are now provided as part of the application form fields directly.

src/components/BulkUploadListItem.tsx
hminsky2002: This is fantastic news! Thank you for checking, yeah I had felt that it was odd that so many of these fields which, if they were not defined, we would want to ignore the object entirely, would be left as optional. Excited to refactor this PR with these changes

src/map-proposals.ts
reefdog: You're right that we should've documented this function and behavior better back in the day! I believe what it's doing is coercing the data structure into a format that's easier to iterate through for building the interface tables. So this is a bad function name but basically `mapApiProposalDataForInterface` would be more explicitly correct.

We can revisit the data structure to see if that's necessary any more, but the intention was that we would reduce the proposal data to only those that matched base fields we knew about, and flattened it to `key: value` as `baseFieldKey: proposalValue` (for the most recent version only), and also removed any blank values from the set.

Which is to say: the output of this process is still desirable, but maybe there's a better modern path to get there, and/or maybe it's a good time to revisit how we present this overall! (Since, for instance, we weren't yet tackling versions.)

src/components/BulkUploadListItem.tsx
reefdog: @hminsky2002 `main` now has the updated SDK with the improved types. 🎉 Feel free to rebase and carry on.

hminsky2002:
@reefdog and @slifty: thank you very much for the sdk update, things are looking much much smoother now! 
I ended up just redoing the whole branch but I think things are basically where they need to be EXCEPT an eslint issue regarding the ProposalFieldValue's ApplicationFormField field in the schema. 

In the final commit you'll see(and I've commented on them for reference) some explicit eslint ignore lines in 'utils/proposalFields.ts' and 'map-proposal-fields-tests.ts' because it is unhappy that we are potentially accessing a field of type 'any.' 

As per our typescript conversations my inkling here is that we should be using a typeguard to verify the type of the ApplicationFormField as such, but I wanted to verify that with you guys before making any changes!

src/utils/proposalFields.ts
```diff
@@ -32,12 +30,13 @@ const getValueOfBaseField = (
 		return undefined;
 	}
 	const fieldValue = proposal.versions[0]?.fieldValues.find(
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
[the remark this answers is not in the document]
hminsky2002: applicationFormField is potentially of type 'any'

src/utils/proposalFields.ts
```diff
@@ -1,19 +1,17 @@
...
 
-const mapProposalBaseFields = (
-	baseFields: BaseField[],
-	proposal: ApiProposal,
-) =>
+const mapProposalBaseFields = (baseFields: BaseField[], proposal: Proposal) =>
 	(proposal.versions[0]?.fieldValues ?? []).map(
 		({ applicationFormField, value }) => {
 			const baseField = baseFields.find(
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
hminsky2002, line 8: more explicit ignore lines

hminsky2002: applicationFormField is potentially of type 'any'

slifty: When you come across things like this please consider opening an issue in the sdk repository (our SDK shouldn't be resulting in eslint warnings!)

I'm not sure why `any` is listed as a possible value but that's definitely a bug!

src/utils/proposalFields.ts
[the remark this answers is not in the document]
slifty: If we do end up needing to keep this ignore, you should add a comment explaning why the ignore is justified (normally eslint-disable is a no no, so it warrants a comment!)

src/utils/proposalFields.ts
hminsky2002: Heard loud and clear, opening an issue now! To that end then I will also remove the draft status to this PR(as I believe this was the only reason I had left as draft)

slifty: Reminder to add a comment explaining the reason for the eslint disable / linking to the opened issue

src/utils/proposalFields.ts
```diff
@@ -1,27 +1,26 @@
...
 				({ id }) => id === applicationFormField.baseFieldId,
 			);
 			return {
 				shortCode: baseField?.shortCode ?? 'missing',
 				fieldName: baseField?.label ?? 'missing',
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment,@typescript-eslint/no-unsafe-member-access
```
slifty, line 14: Reminder to add a comment explaining the reason for the eslint disable / linking to the opened issue (I won't tag all of them but you get the idea!)

slifty (changes requested):
Wooo!  So clean -- the ONLY request is to add that documentation around the eslint-disable lines so we have a better chance at clearing them away later.

hminsky2002:
> Wooo! So clean -- the ONLY request is to add that documentation around the eslint-disable lines so we have a better chance at clearing them away later.

@slifty thank you thank you, i strive for sanitation! Just force pushed with a comment for each file, tried to follow the convention of other eslint disable comments but if you have any notes on it happy to refactor!

src/utils/proposalFields.ts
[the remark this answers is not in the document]
reefdog: @hminsky2002 Bump to add the explanation here too. (Same explanation as the other spot, but want to have it in both places.)

reefdog (commented):
@hminsky2002 This is great! Sorry it's taken my this long to really re-review.

I did find additional opportunities to lean on the SDK. I'm going to commit them, and then ask YOU to review MY commits. I'm also going to tweak the commit messages from your commits to correct the co-author credit syntax; you need a newline above it for GitHub to recognize it, and you can credit my name name rather than nickname. E.g.:

```
Yadda yadda

Issue #120 Yadda yadda

Co-authored-by: Justin Reese <justin@justinreese.com>
```

But I think that when I edit that, I may end up stealing credit for the commit itself, so I may end up inverting the co-author credit to you. lol. We'll see.

Anyway, new work incoming shortly, and yay for this change!

src/utils/proposalFields.ts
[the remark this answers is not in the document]
reefdog: Ohh, wait, I see that you already explained it earlier in the file. Carry on!

reefdog:
@hminsky2002 Ahh, since you opened the PR, it won't let me request you to review. So could you (and/or @slifty) eyeball 75271b74b374bbbfc51d99fe800f861e32b5de69 and let me know if it's GTG?

hminsky2002:
> @hminsky2002 Ahh, since you opened the PR, it won't let me request you to review. So could you (and/or @slifty) eyeball [75271b7](https://github.com/PhilanthropyDataCommons/front-end/commit/75271b74b374bbbfc51d99fe800f861e32b5de69) and let me know if it's GTG?

Just eyeballed and looks gorgeous! Thank you for the assist @reefdog

reefdog (approved):

PR #721, opened 2024-05-01
WIP: Move to vite

slifty:
This PR transitions away from Create React App and over to Vite.

The main work that remains:

1) Updating to vite's equivalent of web vitals
2) Updating any documentation (since env vars are now different)
3) Updating our secrets
4) Ensuring build works
5) Going through and manually testing the heck out of everything!
6) Removing dead dependencies


slifty:
As of right now I believe @hminsky2002 wants to pick this up and run with it; that said Harry let me know if you end up diving into other paths in which case I can do so!

PR #724, opened 2024-05-02
Update EIN to Tax ID

### LINKED ISSUE #723: Update references to `employerIdentificationNumber` to `taxId`
The API is going to be returning a `taxId` instead of an `employerIdentificationNumber` for organizations, and once that starts we should move over to it.

See https://github.com/PhilanthropyDataCommons/service/issues/931

Also this PR: https://github.com/PhilanthropyDataCommons/service/pull/985

slifty:
This PR uses the new `taxId` attribute instead of the deprecated `employerIdentificationNumber`.

Resolves #723

### DIFF
--- a/src/components/OrganizationListGrid.tsx
+++ b/src/components/OrganizationListGrid.tsx
@@ -25,9 +25,7 @@ export const OrganizationListGrid = ({
 				key={organization.id}
 			>
 				<ListGridItemTitle>{organization.name}</ListGridItemTitle>
-				<ListGridItemDetails>
-					{organization.employerIdentificationNumber}
-				</ListGridItemDetails>
+				<ListGridItemDetails>{organization.taxId}</ListGridItemDetails>
 			</ListGridItem>
 		)}
 	/>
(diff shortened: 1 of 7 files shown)


reefdog (changes requested):
One instance was missed somehow:

https://github.com/PhilanthropyDataCommons/front-end/blob/0b60d3831f092c0ac35493db284edc950371fb70/src/components/OrganizationListGrid.tsx#L29

Update that reference and this is good to go! 🙏

src/components/OrganizationListGrid.tsx
```diff
@@ -26,7 +26,7 @@ export const OrganizationListGrid = ({
 			>
 				<ListGridItemTitle>{organization.name}</ListGridItemTitle>
 				<ListGridItemDetails>
-					{organization.employerIdentificationNumber}
+					{organization.taxId}
```
reefdog, line 29: Prettier is tripping on this because with the shorter attribute name, it no longer needs to be on its own line. Can you run Prettier on this and force-push again?

slifty: Argh sorry about that -- this is what I get for trying to sneak in fixes after hours 😅

reefdog: No it's fine, I realize what I should have done (since CI caught this too) is actually approve it with a note, because as soon as it passed then you could merge. Sorry about that!

reefdog (approved):

PR #741, opened 2024-05-15
Guard presigned post field values

### LINKED ISSUE #739: Remove ESLint overrides once SDK fixes `any` issues
In #719, we convert some locally-defined types to use those provided by the SDK. Upon doing so, we discovered some shortcomings of the SDK (https://github.com/PhilanthropyDataCommons/service/issues/1004, ~https://github.com/PhilanthropyDataCommons/service/issues/1011~) that required inline ESLint disabling.

Once those issues are resolved, let's be sure to remove the overrides.

🛑 This issue is currently blocked by:
- [ ] https://github.com/PhilanthropyDataCommons/service/issues/1004
- ~https://github.com/PhilanthropyDataCommons/service/issues/1011~ Scratch this; it's unrelated and a wontfix.

slifty:
This PR fixes a type safety issue when building presigned post headers.

Related to #739

### DIFF
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -110,20 +110,19 @@ const usePresignedPostCallback = () => {
 			body: JSON.stringify(params),
 		});
 };
+const isStringOrBlob = (value: unknown): value is string | Blob =>
+	typeof value === 'string' || value instanceof Blob;
 
 const uploadUsingPresignedPost = async (
 	file: File,
 	presignedPost: PresignedPostRequestPresignedPost,
 ) => {
 	const formData = new FormData();
-	Object.entries(presignedPost.fields).forEach(([key, value]) =>
-		/* eslint-disable-next-line @typescript-eslint/no-unsafe-argument --
-		 *
-		 * SDK incorrectly defines this type.
-		 * https://github.com/PhilanthropyDataCommons/service/issues/1011
-		 */
-		formData.append(key, value),
-	);
+	Object.entries(presignedPost.fields).forEach(([key, value]) => {
+		if (isStringOrBlob(value)) {
+			formData.append(key, value);
+		}
+	});
 	formData.append('Content-Type', file.type || 'application/octet-stream');
 	formData.append('file', file);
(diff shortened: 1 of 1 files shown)


reefdog (approved):
Nice!

So to be sure I understand: the SDK is properly describing the expected `PresignedPostRequestPresignedPost` type (or, I guess, the type of `fields` values); the problem is that we're setting it based on an incoming object which can have values of type `any`. So we need a type guard to ensure only strings or blobs get through. Yeah?

reefdog:
@slifty I just edited your PR description to change `Resolves #739` to `Addresses #739` because that issue is actually a placeholder for removing all `any`-related ESLint overrides. (I would have asked you to edit it — I don't like to edit people's messages — but I wasn't sure you'd see it before merging and triggering the auto-close.)

I incorrectly thought both https://github.com/PhilanthropyDataCommons/service/issues/1004 and https://github.com/PhilanthropyDataCommons/service/issues/1011 were related. I'd like to keep #739 open until https://github.com/PhilanthropyDataCommons/service/issues/1004 is also resolved and we can remove the related overrides.

slifty:
> the problem is that we're setting it based on an incoming object which can have values of type any. So we need a type guard to ensure only strings or blobs get through. Yeah?

Sort of!  The issue is that `Object.entities` returns an array of tuples of type `[string, any]`

Even if we did this:

```
Object.entries({ foo: 'bar' }).forEach([key, value]) => { ... }
```

value would be considered an `any` -- that's because the signature of the javascript function `Object.entries` is "this returns values of the any type"

reefdog:
@slifty I've nodded my way through enough of our conversations that it later turns out I didn't actually understand that I feel the need to confirm that yes, I actually understood this one, thank you! 🙏

slifty:
> @slifty I've nodded my way through enough of our conversations that it later turns out I didn't actually understand that I feel the need to confirm that yes, I actually understood this one, thank you! 🙏

look, only time will tell for sure

reefdog:
@slifty All I know is a type guard is a sort of filter

PR #742, opened 2024-05-17
Use `PlatformProviderResponse` from SDK

### LINKED ISSUE #737: Replace `PlatformProviderResponse` type with SDK version
We should use the `PlatformProviderResponse` type provided by the SDK rather than [defining our own](https://github.com/PhilanthropyDataCommons/front-end/blob/3e39eb2f5815dd0de2caf03733d94242475465ef/src/pdc-api.ts#L240-L245).

🛑 **This issue is currently blocked by https://github.com/PhilanthropyDataCommons/service/issues/1013.**

reefdog:
This PR uses the `PlatformProviderResponse` type from the SDK instead of defining our own.

In order to do this, we needed to disable the ESLint complaint about the `data` property being an `any`. ~This is allowable because in this ersatz implementation of a provider response, we don’t know the shape of the data.~ Update: This is not actually quite accurate. See updated context: https://github.com/PhilanthropyDataCommons/service/issues/1013#issuecomment-2120839690

**Testing:**

Just linting should do it. We actually aren't _using_ this type in a visible way, since the current batch of visible platform data is no longer available to the front-end.

Resolves #737

### DIFF
--- a/src/components/DataPlatformProvider/DataPlatformProviderLoader.tsx
+++ b/src/components/DataPlatformProvider/DataPlatformProviderLoader.tsx
@@ -187,6 +187,14 @@ const ProviderPanelLoader = ({
 		return <NoData onClose={onClose} provider={provider} />;
 	}
 	const { parse, name } = providers[provider];
+	/* eslint-disable-next-line @typescript-eslint/no-unsafe-argument --
+	 *
+	 * The `data` property on a `PlatformProviderResponse` gets typed as `any`
+	 * by the SDK generator, but this is a bug in the generator. We can re-enable
+	 * this rule once the upstream bug is fixed.
+	 *
+	 * See: https://github.com/PhilanthropyDataCommons/service/issues/1013#issuecomment-2120839690
+	 */
 	const { applicant, url, values } = parse(selectedProviderData.data);
 	return (
 		<DataPlatformProviderPanel
--- a/src/pdc-api.ts
+++ b/src/pdc-api.ts
@@ -6,6 +6,7 @@ import type {
 	WritableBulkUpload,
 	Organization,
 	OrganizationBundle,
+	PlatformProviderResponse,
 	PresignedPostRequest,
 	WritablePresignedPostRequest,
 	PresignedPostRequestPresignedPost,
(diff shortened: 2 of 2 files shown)


src/components/DataPlatformProvider/DataPlatformProviderLoader.tsx
```diff
@@ -187,6 +187,11 @@ const ProviderPanelLoader = ({
 		return <NoData onClose={onClose} provider={provider} />;
 	}
 	const { parse, name } = providers[provider];
+	/* eslint-disable-next-line @typescript-eslint/no-unsafe-argument --
+	 *
+	 * The `data` property on a `PlatformProviderResponse` is a true `any`,
```
slifty, line 192: What's genuinely annoying about this situation is that our spec *does* specify that it is an object (which is all that `parse` wants).  The codegen, however, incorrectly translates `object` to `any`:

https://github.com/swagger-api/swagger-codegen/blob/be7735b8c32130a9ca317a955fde3116c9df4507/modules/swagger-codegen/src/main/java/io/swagger/codegen/languages/AbstractTypeScriptClientCodegen.java#L98

I'll open an issue because why not.

reefdog: Ah, interesting. In any case, of course I don't want to wait for that upstream to get resolved before this is merged in.

reefdog: Should I reword this comment, then, so it's more fitting / includes a link to the issue you're about to make?

slifty: Wait, this commit from 5 years ago indicates they fixed it.

I wonder how old our codegen references are -- https://github.com/OpenAPITools/openapi-generator/commit/9368bea130c1bc5a0d039e228954c3e66ce89cc1

slifty: @reefdog I think rewording would be reasonable; tbh the issue that I closed as wontfix probably should be reopened and you should link to that

Issue #720, opened 2024-05-01
Move to Vite


slifty:
We are going to just do the full thing: move to [Vite](https://vitejs.dev/)!

We looked around a bit, NextJS is heavier than this project needs, especially because this is specifically NOT a back-end repo (we have the `service` for that).

#692 #498

PR #748, opened 2024-05-24
Migrate to Vite

hminsky2002:
HARK! HARK! THE END IS NIGH!

This PR contains the work done by @slifty and myself(but mostly @slifty) to migrate our build system from the now-deprecated Create React App to Vite. Principle changes include bringing us to the latest version of Typescript and Version 8.57 of Eslint. 

Notably, by removing the react-scripts package, we now no longer extend the Create React App default eslint rules. Two eslint disables in the 'pdc-api.ts' file have been removed for the time being, as they reference rules that don't exist in the package. A significant point of discussion before merging this PR should be whether we want to reimplement some or all of the linting rules provided by CRA. There exists this package which may be something worth looking into, but it hasn't been updated in 2 years: https://www.npmjs.com/package/eslint-config-react-app.

Another significant change is switching from jest to the 'vitest' framework, which is the testing framework officially used by the project Vite[1], and is designed to be natively compatible with vite.config.js files. Note the additions of the vitest and jsdom packages. The thought process behind this was that since we only have three testing files, it wouldn't change our testing design flow very much, and while vitest isn't native to the vite package it is supported by the project

It modifies Storybook to use the storybook/react-vite package, as this is the storybook-recommended plugin[2]

It also disables the 'react/react-in-jsx-scope' rule from our eslint, as it is used by our eslint-plugin-react package, but no longer necessary as of React 19 [3]. This removes the need for the 'import React' statement above all components. **This makes the commit much much longer but unfortunately is a breaking change :( If there is a better way to format this please let me know I will gladly refactor!**

[1] https://vitest.dev/guide/#projects-using-vitest 
[2] https://storybook.js.org/docs/get-started/react-vite 
[3] https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#eslint 

Closes Issue #720


reefdog (commented):
Awesome! I can't wait to try this out. I'm leaving the initial review up to @slifty as I work on other things, but two quick comments:

- Is this really ready for review? If so, go ahead and move it out of draft status, even if you expect changes.
- Can you go through the README and see if all the setup instructions (and other info) are up to date? I imagine some of it -- e.g. the section on `eject` -- will need some changes.

Otherwise, woohoo and looking forward to playing with this / being unblocked from forward progress!

hminsky2002:
> Awesome! I can't wait to try this out. I'm leaving the initial review up to @slifty as I work on other things, but two quick comments:
> 
> * Is this really ready for review? If so, go ahead and move it out of draft status, even if you expect changes.
> * Can you go through the README and see if all the setup instructions (and other info) are up to date? I imagine some of it -- e.g. the section on `eject` -- will need some changes.
> 
> Otherwise, woohoo and looking forward to playing with this / being unblocked from forward progress!

I suppose it is ready for review! Just moved it from draft to...not draft. Reviewing readme instructions now--definitely will need to drop the cra specific stuff!

.env.example
```diff
@@ -13,11 +13,11 @@
 # https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest
 # In Keycloak, this is realm-local; do not include the trailing
 # `/.well-known/openid-configuration` path
-REACT_APP_OIDC_AUTHORITY=https://auth.example.com/realms/example
+VITE_REACT_APP_OIDC_AUTHORITY=https://auth.example.com/realms/example
```
slifty, line 16: Realizing that we should remove `REACT_APP` from all of these env vars.

So instead of `VITE_REACT_APP_FOO` it would just be `VITE_FOO`.

package.json
```diff
@@ -19,47 +20,52 @@
...
-		"build:app": "react-scripts build",
+		"build:app": "tsc && vite build",
 		"build:storybook": "storybook build -o build/storybook",
 		"storybook": "storybook dev -p 6006",
-		"test": "react-scripts test",
-		"lint:prettier": "prettier . --check",
-		"lint:eslint": "eslint ./src --ext .ts,.tsx,.js --max-warnings=0",
 		"lint": "npm run lint:eslint && npm run lint:prettier",
+		"lint:prettier": "prettier . --check",
+		"lint:eslint": "eslint . --ext ts,tsx,js --report-unused-disable-directives --max-warnings 0",
```
slifty, line 62: This is now linting more than `/src` -- I know why we're doing this, but it's a change that could have accidental inclusions as a result so wanted to just point it out / make sure that running lint after build doesn't result in lint issues.

package.json
```diff
@@ -70,6 +76,11 @@
 		"parserOptions": {
 			"project": "./tsconfig.json"
 		},
+		"ignorePatterns": [
+			"vite.config.js",
+			"build",
+			"dist"
```
slifty, line 82: Ah here's how we prevent linting from going beyond what we want.

As an aside -- I'm not a big fan of using package.json for eslint config (as opposed to an `.eslintrc.json`).  That's out of scope for this PR!

src/components/AppHeader.tsx
```diff
@@ -1,4 +1,3 @@
-import React from 'react';
```
slifty, line 1: Gotta love to see these taken out.  Feels so light and clean!

package.json
slifty: Oh -- I did want to point out the addition of `--report-unused-disable-directives`

What is the reasoning for that?

slifty (dismissed):
Wooooooo!


I think the only real change request is to change the env var names.

Also, the git way to [give shared credit](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors) is to add the following line to the end of your commit:

```
Co-authored-by: Daniel Schultz <slifty@gmail.com>"
```

One meta thought: there's a lot in this single commit.  I realize it's probably hard to break it into smaller ones without having some of them in a broken state which we tend to avoid.  I wonder if the tradeoff would be worth it in the context of a major refactor like this.  I think it is what it is / it's still a good idea to avoid having any commits in main that are "broken" / don't pass CI or compile.

package.json
hminsky2002: > Oh -- I did want to point out the addition of `--report-unused-disable-directives`
> 
> What is the reasoning for that?

According to the eslint docs, it's for reporting any disables that wouldn't have thrown errors anyway! https://eslint.org/docs/latest/use/command-line-interface#--report-unused-disable-directives

hminsky2002:
@slifty @jasonaowen I have done the thing! The branch should be ready to go for merging, with all issues addressed (save `--report-unused-disable-directives`), which I could go either way on, I believe it's just an include from the create Vite latest package. I have highlighted two areas where I had to make changes that I wasn't entirely sure about (removing a now-unused eslint disable and changing a reducer function) so be sure to check out those comments! But otherwise this should be up to date and is tested/verified working locally!

src/components/BulkUploadListItem.tsx
```diff
@@ -90,7 +89,7 @@ export const BulkUploadListItem = ({ upload }: BulkUploadListItemProps) => {
 						{datum}
 					</>
 				),
-			null,
+			'',
```
hminsky2002, line 92: This was changed from null to empty string because typescript was barfing at the assignment of a string to a null value. I am unsure why this wasn't causing an error before, but my understanding is that this should not make a difference as far as output is concerned? I.e if the string was reduced to null before it would be output as an empty string either way

src/pdc-api.ts
```diff
@@ -48,17 +48,6 @@ const usePdcApi = <T>(
 			.then((res) => res.json())
 			.then(setResponse)
 			.catch((e) => logError(e, path, params));
-
-		/* eslint-disable-next-line react-hooks/exhaustive-deps --
```
hminsky2002, line 52: These rules are no longer included in our linting configuration with the react-scripts ejected, so the linter asked me to remove them. I figured if we're not disabling the rule, we no longer need the explanatory comment!

src/components/BulkUploadListItem.tsx
jasonaowen: I don't think that's quite accurate; look at the `null` check in the reducer function:

https://github.com/PhilanthropyDataCommons/front-end/blob/35cb42479ba077cd0580dc402245490030c7c717/src/components/BulkUploadListItem.tsx#L80-L94

I this will cause the first item to have a bullet in front of it, where previously it didn't:

before
after

The prod instance doesn't have any bulk uploads, so this change can't be tested against that - you'll need to make sure your local instance has some and point it at that.

I think this can best be solved by removing the initial value entirely, as well as the trigraph operator; as [the MDN docs on `reduce`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/reduce) say,

> The first time that the callback is run there is no "return value of the previous calculation". If [no initial value is supplied,] the array element at index 0 is used as the initial value and iteration starts from the next element (index 1 instead of index 0).

This should also be extracted into a separate commit.

package.json
jasonaowen: I think that was necessary with create-react-app, so once this lands we can and should move it out into a separate config file!

package.json
jasonaowen, line 1: Somehow, the first commit (fef5f9b628397d45f19fd574ff3433299e2fb4fb) rewrote `package.json` and `package-lock.json` to be indented with spaces rather than tabs, and the second rewrote it back to tabs - as a result, it looks like every line in these files was changed. Can you please fix that, @hminsky2002?

jasonaowen (changes requested):
Very cool, @hminsky2002 and @slifty! I'll be excited to land this.

tsconfig.json
```diff
@@ -1,22 +1,30 @@
...
-		"forceConsistentCasingInFileNames": true,
+		"target": "ES2022",
+		"useDefineForClassFields": true,
+		"lib": ["ES2022", "DOM", "DOM.Iterable"],
+		"module": "ESNext",
+		"skipLibCheck": true,
+
+		/* Bundler mode */
+		"moduleResolution": "bundler",
+		"allowImportingTsExtensions": true,
```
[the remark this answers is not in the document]
hminsky2002: This is a layover from the default vite typescript config file when bootstrapping a vite react typescript app. When @slifty and I first went through this process we wanted to follow their setup as closely as possible, but since we know we want ESLint specifically to disallow this I think it's fine to remove it!

tsconfig.node.json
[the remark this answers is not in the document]
hminsky2002: This file also came along with the vite bootstrapped app config, I will look further into why this might be necessary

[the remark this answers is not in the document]
hminsky2002: https://stackoverflow.com/a/72030801

Comment here suggests that this is a necessary configuration file for running the application in the node environment as opposed to the browser? Other comments suggest it's a bit of a hack included by Vite to get around some compile-time type checking.

.node-version
```diff
@@ -1 +1 @@
-18
+20
```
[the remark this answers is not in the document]
hminsky2002: Vite does indeed support Node 18. https://vitejs.dev/guide/migration.html#node-js-support

Agreed that anything we can do to minimize the changes in this commit is a good thing. I think we keep this as node 18 and make a subsequent bump version issue!

tsconfig.node.json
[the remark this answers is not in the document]
jasonaowen: The code in the Stack Overflow question has this line in `tsconfig.json`:

```
    "references": [{ "path": "./tsconfig.node.json" }]
```

but I don't see that in ours.

package.json
```diff
@@ -1,232 +1,143 @@
...
+		"eslint": "^v8.57.0",
+		"eslint-config-airbnb": "^19.0.4",
+		"eslint-config-airbnb-typescript": "^18.0.0",
+		"eslint-config-prettier": "^9.1.0",
+		"eslint-plugin-react-hooks": "^4.6.2",
+		"eslint-plugin-react-refresh": "^0.4.7",
+		"jsdom": "^24.0.0",
+		"prettier": "^3.3.3",
+		"storybook": "^8.1.3",
+		"typescript": "^5.5.4",
```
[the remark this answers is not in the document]
hminsky2002: No there's no requirement to upgrade here, I think since part of the reason we wanted to migrate was to upgrade typescript (as it was blocked by CRA) I included it here, but it can be a separate version bump!

src/components/BulkUploaderFilePicker.tsx
```diff
@@ -15,7 +15,7 @@ import './BulkUploaderFilePicker.css';
 import { Spinner } from './Spinner';
 
 interface FilePickerMessageProps {
-	text: React.ReactNode;
+	text?: React.ReactNode;
```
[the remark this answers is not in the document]
hminsky2002: I believe I was getting a linting error with it before but no longer seem to be getting it when I make it mandatory again, good catch thank you!

package.json
jasonaowen: @hminsky2002 has changed the `lint:eslint` command to reference `./src` instead of `.`, and there should be no new errors from the unused disable directives, so it's okay to add.

tsconfig.node.json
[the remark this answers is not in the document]
hminsky2002: After discussing this issue, we have decided that the best way forward is to model the tsconfig referencing system off of the latest typescript react create vite app template. This was informed by the vite issue ['
tsconfig.node.json should not be included in tsconfig.json references'](https://github.com/vitejs/vite/issues/11396).

package.json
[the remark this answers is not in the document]
slifty: Just to double clarify: the real answer to the question Jason asked is "yes!"

We could avoid updating typescript, but that would basically mean migrating to an older version of vite, and then migrating to a newer version of vite separately -- not worth the increase in effort, and we can just think of the TypeScript bump as part of the vite migration path.

src/components/BulkUploadListItem.tsx
hminsky2002: This is closed by #804, which should be merged in before this PR

src/components/ProposalListTable.tsx
```diff
@@ -13,7 +12,7 @@ enum ProposalDetailDestinations {
 interface ProposalListTableRowProps {
 	columns: string[];
 	proposal: FrontEndProposal;
-	rowClickDestination: ProposalDetailDestinations;
+	rowClickDestination?: ProposalDetailDestinations;
```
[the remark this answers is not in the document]
hminsky2002: This is closed by #806 which should be merged before this PR

tsconfig.node.json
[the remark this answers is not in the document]
hminsky2002: Scratch that: After further discussion with @slifty , we have decided that we would rather not use typescript project references in this repository, as we don't fully understand their weight nor do we explicitly need any optimization they provide. Though the base vite create react typescript package does include a tsconfig.node.json file, we've decided that we will stick with a single tsconfig.json file, and merely update it to work within a vite environment

hminsky2002:
@jasonaowen @slifty Ok I believe all requested changes have now been handled and this will be ready to merge AFTER #804 and #806 are merged and it is rebased to main!

hminsky2002:
@jasonaowen the branch is rebased on #804 and #806 and ready to be merged (after final review, of course :) )

jasonaowen (approved):
Great work, @hminsky2002!

I was able to run this locally, log in, upload bulk data, browse proposals, and browse organizations; I was also able to poke around in Storybook.

Three small changes below, but I don't think they need further review - please fix and merge at your leisure, @hminsky2002!

hminsky2002:
@jasonaowen WOOOO! Just ammended those changes and re-pushed, thank you for the final lookover. As soon as CI/CD passes I will be merging this badboy--happy thursday!

PR #804, opened 2024-08-23
Fix BulkUploadListItem metadata reducer

### LINKED ISSUE #803: Remove initial value for BulkUploadListItem reducer function
The reducer function in the BulkUploadListItem component defaults to the null value: 

```typescript
const metadataElements = metadata
		.filter((datum) => datum !== null)
		.reduce(
			(accumulator, datum) =>
				accumulator === null ? (
					datum
				) : (
					<>
						{accumulator}
						{METADATA_DELIMITER}
						{datum}
					</>
				),
			null,
		);
```
During the work on #748, we discovered that the value defaulting to null throws a type error on update of our typescript eslint plugin. 
As pointed out by @jasonaowen in https://github.com/PhilanthropyDataCommons/front-end/pull/748#discussion_r1710319764, the best way to handle this is to remove the initial value entirely, as well as the trigraph operator

hminsky2002:
This commit drops the initial value from the metadataElements reducer function, along with the conditional statement. As per the MDN docs[1], "if [no initial value is supplied,] the array element at index 0 is used as the initial value." Thus we don't need to specify an initial value, as we default to the first value, which, if it exists, is determined to be not null by the filter function.

Closes #803

### DIFF
--- a/src/components/BulkUploadListItem.tsx
+++ b/src/components/BulkUploadListItem.tsx
@@ -79,19 +79,13 @@ export const BulkUploadListItem = ({ upload }: BulkUploadListItemProps) => {
 
 	const metadataElements = metadata
 		.filter((datum) => datum !== null)
-		.reduce(
-			(accumulator, datum) =>
-				accumulator === null ? (
-					datum
-				) : (
-					<>
-						{accumulator}
-						{METADATA_DELIMITER}
-						{datum}
-					</>
-				),
-			null,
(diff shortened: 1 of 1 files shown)


jasonaowen (approved):
Thanks for pulling this out, @hminsky2002! I was able to run it locally and see the bulk uploads with the separators as expected. Yay for simpler, more correct code!

PR #837, opened 2024-09-21
Include Storybook in deploys

jasonaowen:
When we migrated to Vite in PR #748, we changed the output directory from `build` to `dist`. We (belatedly) updated our Netlify configuration to match this change, but in doing so left Storybook behind.

Update the output directory of Storybook to match, so that it is included in deploys.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
 		"start": "vite",
 		"build": "npm run build:app && npm run build:storybook",
 		"build:app": "tsc && vite build",
-		"build:storybook": "storybook build -o build/storybook",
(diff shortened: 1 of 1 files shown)


jasonaowen:
It is included in the deploy preview: https://deploy-preview-837--philanthropy-data-commons-viewer.netlify.app/storybook/, which you can also reach by following the deploy preview link in the [above comment](https://github.com/PhilanthropyDataCommons/front-end/pull/837#issuecomment-2364781501) and clicking Developers > Storybook.

slifty (approved):
Great catch and fix thanks @jasonaowen

PR #852, opened 2024-10-03
Deploy to DigitalOcean App Platform

jasonaowen:
In order to deploy to DigitalOcean App Platform, we need to make one adjustment to the code, and add some documentation.

Since this is currently deployed to Netlify, it seems to be about time to add a "Deploy to Netlify" button, and then a "Deploy to DigitalOcean" button next to it. (I wish they were the same size, but not sure how to do that in markdown.)

Issue PhilanthropyDataCommons/deploy#118 Migrate to DigitalOcean Apps

### DIFF
--- a/src/components/AppNavbar.tsx
+++ b/src/components/AppNavbar.tsx
@@ -136,7 +136,7 @@ const AppNavbar = () => {
 								</DropdownMenuLinkDescription>
 							</DropdownMenuLink>
 							{import.meta.env.VITE_SHOW_STORYBOOK === 'true' && (
-								<DropdownMenuLink to="/storybook" reloadDocument>
+								<DropdownMenuLink to="/storybook/" reloadDocument>
 									Storybook
 									<DropdownMenuLinkDescription>
 										View our UI component library. (Only relevant to PDC front
--- a/.do/deploy.template.yaml
+++ b/.do/deploy.template.yaml
@@ -1,0 +1,28 @@
+spec:
+  ingress:
+    rules:
+      - component:
+          name: front-end
+        match:
+          path:
+            prefix: /
+  name: pdc-front-end
+  static_sites:
+    - catchall_document: index.html
+      environment_slug: node-js
+      envs:
+        - key: VITE_API_URL
+          scope: BUILD_TIME
+          value: https://api.example.org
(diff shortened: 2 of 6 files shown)


jasonaowen:
I tested deploying to DigitalOcean on the [`digitalocean-app-test` branch](https://github.com/PhilanthropyDataCommons/front-end/tree/], and successfully created a new app from the template. I also tried to test the deploy to Netlify button, but it required more GitHub permissions... how did I do it in the first place? Anyway, the deploy preview succeeded - and, in particular, picked up the redirect rules moved into the `netlify.toml` file.

jasonaowen:
There is a new `pdc-front-end-test` app at https://pdc-front-end-test-i7pvl.ondigitalocean.app/ , and I configured Keycloak to allow it. We didn't previously have a test instance of the front-end, so this is new. I am planning to bundle the request for a new DNS record with the request for prod DNS updates.

src/components/AppNavbar.tsx
```diff
@@ -136,7 +136,7 @@ const AppNavbar = () => {
 								</DropdownMenuLinkDescription>
 							</DropdownMenuLink>
 							{import.meta.env.VITE_SHOW_STORYBOOK === 'true' && (
-								<DropdownMenuLink to="/storybook" reloadDocument>
+								<DropdownMenuLink to="/storybook/" reloadDocument>
```
hminsky2002, line 139: Was this meant to be included in this PR?

slifty: @hminsky2002 when reviewing I recommend going "commit by commit" -- you can see the rationale for this change in the commit message that included it: https://github.com/PhilanthropyDataCommons/front-end/pull/852/commits/a04c1b34e02995e7ef107879b353f29b103f7b69

jasonaowen:
> Looks great thanks! Similar to the service repo I didn't quite know _how_ the templates are used from the code base / git history as it stands. (e.g. even something as simple as a sentence in the commit along the lines of "these templates are used by digital ocean / netlify at the point the user clicks the Deploy button")

Good call; I was [too deep](https://xkcd.com/2501/) in it. Rebased and reworded!

PR #881, opened 2024-10-30
Upgrade pdc sdk

### LINKED ISSUE #836: Upgrade `@pdc/sdk`
We need to update the front-end to use the latest SDK produced by the back-end. Since it contains breaking changes, we'll need to update our client code to match.

See also PR #832

hminsky2002:
This PR upgrades the sdk to the latest version, and refactors the codebase to reflect two major updates: the change in name from the entity 'organization' to 'changemaker', and the requirement for a source id to be referenced when creating a new bulkupload. Notably, the current approach to providing a source is limited to a constant number, wherein we fetch the first 1000 sources in the database and filter only on that. See [service issue 1295](https://github.com/PhilanthropyDataCommons/service/issues/1295). 

Closes #836


bickelj:
@hminsky2002 I added the above deploy preview address to the valid redirect URIs for the client associated with the front end, so now login should work there. And when looking at the preview, I see the Changemakers and the Proposals. Nice.

src/components/AppNavbar.tsx
```diff
@@ -30,15 +30,15 @@ const AppNavbar = () => {
 		<nav className="App-navbar">
 			<ul>
 				<li>
-					<NavLink to="/organizations" className="App-navbar__item">
+					<NavLink to="/changemakerss" className="App-navbar__item">
```
bickelj, line 33: Typo here and confirmed it breaks the link by looking at deploy preview.

src/pages/AddData.tsx
```diff
@@ -41,10 +49,20 @@ const BulkUploader = ({ onBulkUpload }: BulkUploaderProps) => {
...
+		}
+
+		if (systemSource?.id) {
+			await registerBulkUpload({
+				fileName: file.name,
+				sourceKey: presignedPost.fields.key,
+				sourceId: systemSource?.id,
+			});
+		} else {
+			throw new Error('No System Source Available');
```
bickelj, line 64: Good call throwing an `Error` here. I wonder how it surfaces and if the message could be more explicit, like `No System Source with short code = yada found at [url used]`.

src/utils/sources.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Iterates through a list of Sources to find the Source with the specified
+ * data provider shortcode
+ * Returns a Source with the specified data provider shortcode,
+ * or undefined if no such source exists
+ * @param {Source[]} entries an array of Sources
+ * @param {string} systemSourceShortCode the shortcode for the system source
+ * @return {Source | undefined} the system source, or undefined if it does not exist
+ */
+const getSystemSource = (entries: Source[], systemSourceShortcode: string) =>
```
bickelj, line 12: Unfortunate that this needs to be done but a clean and concise implementation given that it needs to be done!

bickelj (approved):
I was able to see the app in action at the deploy preview. I did not try to run it locally, however. The changes all appear expected to me.

Since the app is broken until this gets merged, I am smashing "Approve" even though I know near-zero React. If you'd like to wait for others feedback, I cannot disagree with that decision either.

My simple way of thinking is that by merging this PR, if something breaks it won't be significantly more broken than it is now, and we have an opportunity to unbreak it with a merge.

package.json
```diff
@@ -6,7 +6,7 @@
 		"@axa-fr/react-oidc": "^7.22.32",
 		"@fontsource/source-sans-pro": "^5.1.0",
 		"@heroicons/react": "^2.1.3",
-		"@pdc/sdk": "^0.11.1",
+		"@pdc/sdk": "^0.16.1",
```
jasonaowen, line 9: Right now, this upgrade happens in the first commit, and therefore the first commit does not build.

Instead, let's gradually bump the SDK version along with the changes necessary for that specific version of the SDK. I believe, in order, they are: handle timestamp strings, include a source id with bulk uploads, and rename organizations to changemakers.

src/pdc-api.ts
```diff
@@ -189,20 +190,32 @@ const useProviderData = (externalId: string) =>
...
+
+const SOURCES_DEFAULT_PAGE = '1';
+const SOURCES_DEFAULT_COUNT = '1000';
+
+const useSources = (page: string, count: string) =>
+	usePdcApi<SourceBundle>(
+		'/sources',
 		new URLSearchParams({
 			_page: page,
 			_count: count,
```
jasonaowen, line 221: This works, but the API docs do not list the `_page` and `_count` parameters. New issue opened: https://github.com/PhilanthropyDataCommons/service/issues/1313

src/pages/AddData.tsx
```diff
@@ -31,6 +38,7 @@ interface BulkUploaderProps {
 	onBulkUpload: () => void;
 }
 const BulkUploader = ({ onBulkUpload }: BulkUploaderProps) => {
+	const [sources] = useSources(SOURCES_DEFAULT_PAGE, SOURCES_DEFAULT_COUNT);
```
jasonaowen, line 41: The problem that @hminsky2002 called out with this is that if the system source does not show up in this call, we error down below.

This is exasperated by two factors: using a synthetic key so that we need to do this at all, and that the `/sources` endpoint returns the most recent sources (sorted by id descending) rather than the earliest sources, as the system source should always be the first source.

The change that introduced the system source, https://github.com/PhilanthropyDataCommons/service/pull/1116, created the table and inserted the system source row in the same migration, and the id is `GENERATED ALWAYS AS IDENTITY`, so the system source ID should always be `1` and cannot be `UPDATE`d. Maybe it's worth hardcoding that here? cc @slifty 

```psql
pdc=*> update sources set id = 2;
ERROR:  column "id" can only be updated to DEFAULT
DETAIL:  Column "id" is an identity column defined as GENERATED ALWAYS.
```

jasonaowen (changes requested):
Looking good! Thanks for working on this, @hminsky2002.

I believe this covers what we discussed during my office hour.

package.json
hminsky2002: @jasonaowen Egads! I've hit an unfortunate issue here -- the SourceBundle, which is needed for the source handling, was missed in the openapi spec and I had to add it back in (https://github.com/PhilanthropyDataCommons/service/pull/1281) at a later date. As such, an older version of the pdc cannot be used for the sourceId commit without also refactoring many organizations to changemakers. Thoughts on either merging the two commits? That seems messy but then again, the changes are separate enough that they might still be discnerable

slifty: My .02 which may not be agreed on here is that it is appropriate to have one commit that bumps to the latest version along with all required changes to make that bump.

I know this is not ideal, but also I think this is the price of allowing versions to bump so much!

Alternatively Harry you could decide to take Jason's approach and bump one version at a time.  I'm just not sure that's worth the extra dev time really.

src/pdc-api.ts
slifty: I'll fix this up thanks for the catch @jasonaowen

src/pages/AddData.tsx
slifty: @jasonaowen that's fine; in the long run IMHO it would be appropriate to provide a documented API mechanism to load the system source rather than relying on magic values.

jasonaowen: @slifty that could be good, although it's not clear to me how often it'd be used or if it'd be used by other partners or only for the type of admin work that our team is still forecast to do. If it's just us, a magic number is fine, and `1` is the second least-magical number! If we could encode that in the SDK somehow, or at least document it, that would be good.

For now, @hminsky2002, I suggest a new `useSystemSource` hook that fetches `/sources/1`, so that you don't have to worry about search or pagination or anything like that.

hminsky2002:
@jasonaowen @slifty In a delightful turn of events, since we are now only retrieving a single Source object for the systemSource, I was able to at least split this PR into source handling and changemaker refactor commits! Also, I added a small change to accommodate the recent refactor of the bulk-uploads route! Should be ready for final review/merging!

src/pages/AddData.tsx
```diff
@@ -31,6 +32,7 @@ interface BulkUploaderProps {
 	onBulkUpload: () => void;
 }
 const BulkUploader = ({ onBulkUpload }: BulkUploaderProps) => {
+	const [systemSource] = useSystemSource();
 	const createPresignedPost = usePresignedPostCallback();
 	const registerBulkUpload = useRegisterBulkUploadCallback();
 	const handleUpload = async (file: File) => {
```
jasonaowen, line 38: I believe this should be using `useCallback`, because right now it appears to be recreating this callback on every render of the `<BulkUploader>` component.

However, this is not a new problem, so it doesn't need to be fixed in this PR.

jasonaowen (approved):
This looks great! Thanks for working through all that.

I was able to test it by successfully uploading a bulk CSV, and poking around in the changemakers tab.

jasonaowen:
Just to add some breadcrumbs for later, this also handles the front-end portion of https://github.com/PhilanthropyDataCommons/service/issues/1117 .

PR #914, opened 2024-12-03
Add source label to proposal version

### LINKED ISSUE #913: Display source on proposal versions
With the addition of sources, we would like to display the source for a given proposal version for users to see. Ideally something similar to the way in which we currently display the organization for a given proposal.

hminsky2002:
This commit adds the rendering for a given proposal version's Source, following the same pattern as we currently use for applicant and tax ID (the Panel Tag component).

Closes #913


jasonaowen (approved):
This looks great - thanks, @hminsky2002!

PR #964, opened 2025-02-05
Update dependabot interval

### LINKED ISSUE #963: Move dependabot updates to a monthly interval
Since we are not actively working on this repository, we've decided it makes sense to extend the dependabot update cycle from weekly to monthly.

hminsky2002:
Closes #963


slifty (approved):
Sounds good to me while front-end is in low-maintenance mode.

PR #990, opened 2025-04-16
Upgrade to Node 22

### LINKED ISSUE #989: Upgrade to Node 22
Node 22 is active; per our policy we should be using it.

slifty:
This PR upgrades our engine version to Node 22 (active).

I'm not sure if we have to make changes to the server or if it will detect the proper version from these files.  Let's find out!

Resolves #989


hminsky2002 (approved):
Hurray! Seems not to cause any issues when running locally.

Issue #988, opened 2025-04-16
Remove non-admin functionality


slifty:
Since we built out this repository the PDC has expanded, and there is now a new "Exchange App" which will be covering a large swath of user stories.  This is great for the ecosystem, and that software is going to be open source (once done I'm sure we'll link to it somewhere).

What that means for this repository is that the focus can be narrowed to JUST being `superuser` stories.  This means it will hone in on things like:

1. Base field management
2. Server settings
3. Integration management
4. User and permission management

(note that it's possible that user permission management will *also* exist some day on the exchange app, but we expect it to still have a place here from the superuser lens)

---

All of the above is to say: we can remove a fair amount of functionality that isn't related to the above.

- [x] Bulk uploads
- [x] Proposal lists
- [x] Organization (changemaker) lists

We should also reflect on some of the peripheral / informational functionality that we hold:

- [ ] Information on the landing page
- [ ] Developer links
- [ ] The About Page

slifty:
Now that the functionality is removed, we are going to call this "done" -- there will be further changes to the content on the referenced pages but we will adjust those as we make progress.

PR #991, opened 2025-04-16
Remove Changemaker functionality

slifty:
This PR removes interfaces related to changemakers, as we are no longer supporting non-admin use cases for the PDC front-end.

Related to #988


hminsky2002 (approved):
@slifty I'm shocked there wasn't more to review here, but after testing locally and scouring for anything changemaker related I came up blank. Front-end still functions and importantly auth is not affected. Looks good, will review the subsequent removal PRs

PR #992, opened 2025-04-16
Remove Proposal functionality

slifty:
This PR removes functionality related to viewing proposal data; these use cases are now covered by the Exchange App.

Related to #988


src/pages/Landing.tsx
```diff
@@ -29,27 +28,19 @@ const Landing = () => {
...
-							block
-							onClick={() => {
-								login('/').catch(logger.error);
-							}}
-						>
-							<UserIcon className="icon" />
-							Log in
-						</Button>
-					)}
+					(!isAuthenticated ? (
```
hminsky2002, line 31: This block needs to be wrapped in a curly bracket, it's being read as html text right now

slifty: Great catch thank you (and fixed)

hminsky2002 (approved):
Zam, nice work @slifty !

PR #993, opened 2025-04-16
Remove BulkUpload functionality

slifty:
This PR removes functionality related to Bulk Uploads, since these use cases are now the responsibility of the Exchange App.

Related to #988


src/pdc-api.ts
```diff
@@ -64,90 +55,6 @@ const usePdcApi = <T>(
 	return [response, fetchData];
 };
 
-const usePdcCallbackApi = <T>(
```
hminsky2002, line 67: I suppose if we're not using this anymore we should get rid of it, but it seems to me like we could encounter a need for something similar again, so it's a shame to delete it. I guess we'll always have this commit to reference

hminsky2002 (approved):
Looks great, it feels deeply saddening to open the pdc front end application now that it is stripped of so much. The unstoppable march of progress :(

src/pdc-api.ts
slifty: Yeah :(

It's always a good idea to remove unused code; worst case of removing is you add it back, worst case of leaving is it sits there for the rest of time silently breaking as the code around it changes.

PR #1013, opened 2025-05-20
Migrate front end to vue

### LINKED ISSUE #1009: Migrate front-end to Vue
@slifty and I discussed in #996 the pros and cons of migrating our front-end framework, and settled on moving to Vue as it is significantly more lightweight than React, and seems better suited to the needs of the front-end as an admin interface. We had also discussed migrating to svelte, but all things being equal chose Vue because I have experience working on it (namely from another OTS project, https://github.com/OpenTechStrategies/arches-app-exp).

hminsky2002:
The great and powerful vue migration. This PR refactors our entire front-end application from react to use Vue3Js.

Closes #1009


slifty:
@hminsky2002 this is still a draft PR right?

src/router.ts
```diff
@@ -1,8 +1,12 @@
 import { createWebHistory, createRouter } from 'vue-router';
 import LandingView from '@/views/LandingView.vue';
+import NotFoundView from '@/views/NotFoundView.vue';
 import type { Component } from 'vue';
 
-const routes = [{ path: '/', component: LandingView as Component }];
+const routes = [
+	{ path: '/', component: LandingView as Component },
```
slifty, line 7: The `as Component` lines raised an eyebrow for me since that is a typecast -- we don't know it's a component, it could be anything.

Do you get a type error if you omit `as Component`?  When I make the change locally it seems fine.

.env.example
```diff
@@ -13,10 +13,12 @@
 # https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationRequest
 # In Keycloak, this is realm-local; do not include the trailing
 # `/.well-known/openid-configuration` path
-VITE_OIDC_AUTHORITY=https://auth.example.com/realms/example
+VITE_KEYCLOAK_AUTHORITY=https://auth.example.com/realms/example
 
 # The client ID is assigned by the IdP
-VITE_OIDC_CLIENT_ID=example-front-end
+VITE_KEYCLOAK_CLIENT_ID=example-front-end
+
```
slifty, line 20: (Random line to note that this commit message also has the wrong issue title 😅)

env.d.ts
```diff
@@ -10,4 +10,3 @@ interface ImportMetaEnv {
 interface ImportMeta {
 	readonly env: ImportMetaEnv;
 }
-
```
slifty: Not a big deal but you added this whitespace in the previous commit and then removed it in this one.  May be worth just not adding it in the first place ;)

package.json
```diff
@@ -4,6 +4,7 @@
 	"private": true,
 	"dependencies": {
 		"@axa-fr/react-oidc": "^7.25.9",
+		"@dsb-norge/vue-keycloak-js": "^3.0.2",
```
slifty, line 7: Should this have been added in the previous commit?  I believe it's used there.

src/views/LandingView.vue
```diff
@@ -1,10 +1,81 @@
...
+}
+
+.launch-buttons {
+	display: flex;
+	gap: var(--fixed-spacing--1x);
+	flex-direction: column;
+
+	&.authenticated {
+		flex-direction: row;
+	}
```
slifty, line 82: Full disclosure: given the number of components migrated in this commit I did not compare each one side by side, but scanned the code to see if anything jumped out 🙈

src/components/EmailLink.vue
```diff
@@ -0,0 +1,49 @@
...
+	if (subject) {
+		paramObj.subject = subject;
+	}
+	if (body) {
+		paramObj.body = body;
+	}
+	// The following line encodes both the key and value of the object using `encodeURIComponent`
+	// and joins it into a query param string. It's like `URLSearchParams` except spaces are
+	// encoded as `%20` instead of `+`, which is necessary for email clients to decode them.
+	// TODO: Extract this out to a utils file so we can document and test it.
```
slifty, line 38: We don't have a lint rule to ban the use of `TODO` so I will have to emulate the rule!

If we wanna do this later then we should make an issue for it; otherwise just remove the note and let someone decide to do it later, OR even better: just do it now.

That said, I was asking the robots "hey does this really not exist?" and it flagged a bug in the implementation (this implementation is not able to support arrays apparently)

And then it was like `qs is a library that just does this` ✨  and so I'm gonna throw out the idea of adding https://www.npmjs.com/package/qs instead and using it for this.  With 95m weekly downloads we're in good company.

src/components/BaseFields.vue
```diff
@@ -0,0 +1,38 @@
+<script setup lang="ts">
```
slifty, line 1: My main concern with this commit overall is...

If it isn't used now, it may well not be used later (especially considering we're approaching base field admin from a clean slate); until that potentially-never-later these components all represent dead code.

I'd rather us just blow the files away and pull them back up in the future if they're needed.  What do you think?

slifty (commented):
I didn't finish the review but I think there's enough to act on in this!

Two main meta-comments: we shouldn't migrate dead code!  (This also includes utilities that we aren't using).

We can add them back if they become useful: we have the code history, but they're just silent tech debt as long as they exist in the repository unused.

Also in general the commit message's issue title should be consistent / match the real issue title!

package.json
```diff
@@ -67,13 +75,18 @@
 			"plugin:@typescript-eslint/recommended",
 			"plugin:@typescript-eslint/recommended-requiring-type-checking",
 			"plugin:@typescript-eslint/strict",
-			"airbnb",
-			"airbnb/hooks",
-			"airbnb-typescript",
+			"plugin:vue/vue3-recommended",
```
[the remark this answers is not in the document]
hminsky2002: you and me both friend, you and me both

src/components/BaseFields.vue
hminsky2002: Commits containing the rebuilt components are now sourced in the 'unused-vue-components' branch

slifty (approved):
Wait... this isn't Svelte.

jk jk: Yay!

There is a LOT here so disclaimer that we may find some kinks over time but I think that's the nature of this kind of migration.

Side note

Issue #1060, opened 2025-07-25
Implement Bulk Uploader


hminsky2002:
Figma Design hosted here: https://www.figma.com/design/9QpV41Ibm6gFXFj8wmwfc0/PDC-Front-End

The bulkuploader is a pdc front-end application designed to support the uploading of bulk data to the pdc service in a user-friendly UI.

slifty:
This issue has been replaced by its various sub-tasks (the issue itself is an epic / milestone)

PR #1081, opened 2025-07-31
Create Bulk-Uploader application

### LINKED ISSUE #1069: BulkUploader: Create Bulkuploader package
On successful resolution of #1048, the first step to implementing #1060 is to create a new bulkUploader package in the repository. This can be minimally fulfilled by copying the code in our existing admin-interface package, but updating the package name and all references to it in the root package.json and CI/CD files.

Related to #1060 
Depends on #1048

hminsky2002:
This commit creates the bulk-uploader application, which is mostly a copy of the existing admin-interface app, with components we know for certain we're not using stripped away. Note that a bug was found with the admin-interface package where the prettier format script did not include the write flag, so that was added in both applications. We did well to configure the monorepo to support multiple applications, so the only specific reference we need to update is in dependabot. Otherwise the root scripts and workspaces in the root package.json cover the new application. 

Furthermore, we've elected to introduce shared components and utilities package, noting that there is much shared overlap between the bulk uploader and the base field admin interface.

Also, a very minor change included in this commit is the correcting the errounesoly capitalized 'Monorepo' in the README.md.

Related to #1060 
Closes #1069


hminsky2002:
> Since this is a straightforward copy of the admin interface, there isn't a ton here for me to comment on (I could I suppose, but it'd just involve commenting on code that's already been vetted for the admin interface), but my real suggestion is to factor the components, storybooks and maybe some of the utilities from both UIs into shared packages. Perhaps starting with the components and storybooks.
> 
> Now, should that happen in this PR, or some subsequent PR? Well, this is the moment of separation for the code base, right? These components are being copied, and to put them back together into a shared package after this might involve resolving changes that were made to one side but not the other. This is also when we have the most leverage to make that kind of call, later there will be other demands. What happens if we don't do that now? The subsequent history of the repo will become path dependent on this one forking path. I leave you with [this diagram](https://s3.amazonaws.com/standardsmanualimages/gallery/original_005-2500.jpg).

@chriszs I think you're right that a shared components package would be the right approach, especially considering that the proposed base-field admin interface is already designed with some of the same components as the bulkuploader (namely the side panel). I think the way to go is to make the first commit in this pr a creation of the shared components package, and the second commit the implementation of the base application using those shared components. That diagram is awesome for reference

hminsky2002:
@chriszs pushing up my current progress on making a shared library package -- I think I'm a little too thick in the weeds on this right now, would appreciate some feedback/guidance if you've done a shared ui component package before! 

Broad strokes questions are: 

1) for shared UI components, do we need to build them before importing them into our applications, or are we just treating a package folder like an additional set of files to be imported and built with the application

2) does it make sense to group utility/api functionality with the ui components in one hefty package or split them up? I've considered both angles and on the one hand it's preferable to have all shared functionality between applications in one place, but on the other hand it may be better to distinguish between functions and components. Then again, it's all vue so maybe that's enough of a reason to group everything together. 

Again, there are still some building/runtime issues with this code, but I think you should take a look at it before I proceed down a bad path.

PR #1092, opened 2025-08-11
1063 implement side panel

### LINKED ISSUE #1063: Implement bulk uploader side panel

hminsky2002:
This PR contains the implementation of a side panel across both our applications. The side panel is built to spec as defined in the figma (see #1063)

Closes #1063


hminsky2002:
> Looks mostly good! Two comments.
> 
> This doesn't have any sort of responsive behavior on small screens, which results in this display on mobile: Image
> 
> Not sure if mobile devices disregarded on purpose, but I wonder if it would make sense to clean that up a bit, so if someone lands there on one, it's less of a mess.

I think historically we haven't designed this for mobile/we're not being expected to design a mobile interface (our figma is desktop only). The previous iteration of this application had some forced layout that made it unusable on mobile as well, so the precedent is there. Certainly I think it would be preferable to keep it mobile friendly, but we're gonna run into this again when we implement the table component (which is not a trivial conversion for mobile, i think): 
Screenshot 2025-08-11 at 11 38 52 AM

I can double check with @kfogel and @jmergy but I think it makes the most sense right now to continue to disregard the mobile layout until requested otherwise!

PR #1108, opened 2025-08-15
Add granular CI commands

### LINKED ISSUE #1107: Improve linting CI
We now have a single `lint` command which runs lint and prettier, but it's nice to be able to have these as separate steps since we can see exactly what's going right or wrong.

We also don't run `tsc` in our linting (which means builds can fail even though linting passes in the event that there's a type issue).

slifty:
This PR improves our workspace commands and creates more granular CI steps in line with our `service` repository.

It also adds `tsc` checks and `actionlint` to the lint steps.

Assuming this looks good, I'll update the required actions to remove `run_lint` and replace it with those more granular lint steps

Resolves #1107


hminsky2002 (approved):
Also great, and welcome addition to our work flows thanks @slifty

PR #1224, opened 2025-10-27
Enable auto-merge for Dependabot PRs

### LINKED ISSUE #1223: Enable dependabot auto-merge
The title of this issue might sound like a terrile idea but hear me out.

Last week I met with the Dependabot PM to go over our workflow (which happens to be a fairly common workflow for OTS and clients of OTS).  Essentially:

1. We require CI to pass
2. We require approval for merge
3. { other rules but those two are the key for this conversation }

I explained that right now we merge when approving using a `@dependabot merge` command, and that the upcoming deprecation will mean we have to approve and then merge manually / extra button clicks.  I also mentioned the opsec issue of merging causing a risk that we merge a dependabot spoofer.

---

Ultimately the suggestion she made is a good one, which is to use a [GitHub Action to automatically enable auto-merge](https://docs.github.com/en/enterprise-cloud@latest/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#enabling-automerge-on-a-pull-request) *for Dependabot PRs*.  This means when Dependabot opens a PR the "auto-merge" setting will be enabled by default / we don't have to click merge or auto merge.

Ultimately this is very similar to our current flow, just now we no longer need to type a merge command (or hit a merge button).  We can simply approve, and then so long as CI has passed / passes the merge will occur.

slifty:
This PR adds a workflow to enable auto-merge for Dependabots, which should save us lowly human reviewers a click for each Dependabot PR (these add up I swear).

Resolves #1223

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,0 +1,16 @@
+name: Auto-merge
+on: pull_request
+
+permissions:
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: github.event.pull_request.user.login == 'dependabot[bot]'
+    steps:
+      - name: Enable auto-merge for Dependabot PRs
+        run: gh pr merge --auto --merge "$PR_URL"
+        env:
+          PR_URL: ${{github.event.pull_request.html_url}}
+          GH_TOKEN: ${{secrets.GITHUB_TOKEN}}


chriszs:
Might need to specify the `contents: write` permission to merge: https://github.com/cli/cli/discussions/6379

hminsky2002 (approved):

PR #1229, opened 2025-10-28
1228 render file download url

### LINKED ISSUE #1228: Render presignedPost download url for bulkuploads
Now that we have download links for bulkuploads, we can render them in the individual bulkupload view! We can finally remove that tbd field.

hminsky2002:
Closes #1228


chriszs:
Seems okay! I tried to make your endpoint URL logic cleaner and couldn't.

PR #1235, opened 2025-10-28
1232 resolve routing redirection

### LINKED ISSUE #1232: Resolve routing redirection
Originally noted by @kfogel, but a larger issue that I have confirmed across the bulk-uploader and admin-interface applications. Page reloads redirect to the home page, instead of the page and the route that the user is currently on. This is likely a problem with our vue router config for both applications!

hminsky2002:
Closes #1232

### DIFF
--- a/packages/utilities/src/keycloak.ts
+++ b/packages/utilities/src/keycloak.ts
@@ -30,7 +30,7 @@ const getKeycloakOptions = (
 		},
 		init: {
 			onLoad: 'check-sso',
-			redirectUri: `${window.location.origin}/`,
+			redirectUri: `${window.location.origin}${window.location.pathname}${window.location.search}`,
 			scope: 'openid',
 			checkLoginIframe: false,
 		},


packages/utilities/src/keycloak.ts
```diff
@@ -30,11 +30,11 @@ const getOptions = (
 		},
 		init: {
 			onLoad: 'check-sso',
-			redirectUri: `${window.location.origin}/`,
+			redirectUri: `${window.location.origin}${window.location.pathname}${window.location.search}`,
```
chriszs: Could also be constructed as a `URL`, e.g. `new URL(window.location.pathname + window.location.search, window.location.origin).href`, but then what's the distinction from `window.location.href`?

hminsky2002: @chriszs i had tried using `window.location.href`, but got an error wherein on login, the user would first be directed to an empty redirect page, and would have to reload to get the actual page content. I'm not sure why that was, mind you. It would probably be worth investigating a little further, but I think this works for now!

PR #1275, opened 2025-11-18
Ignore major Storybook updates

slifty:
This PR stops Dependabot from bumping storybook (and related dependencies) to a version that only supports ESM


hminsky2002 (approved):
Yep, this will calm our dependabot down, thanks @slifty !

PR #1297, opened 2025-12-15
1068 implement table sorting [2/??]: Adjust column helper typing, add <EditIconLink/>

chriszs:
This is a PR that ladders on @hminsky2002's #1253 to make a couple suggestions. Here's how I got here:
1. I tried to regain the typing characteristics of TanStack Table's built in [column helpers](https://tanstack.com/table/latest/docs/guide/column-defs#column-helpers) by borrowing more heavily from their design, reverting to a curried factory function called `createColumnHelpers` that returns several helpers instead of the plain function approach. The goal here was for `.getValue()` in the `cell` rendering method of the column definitions to return a specific type drawn from the relevant column to obviate the need to cast the value, and that worked.
2. ~~I experimented with using [Vue's JSX support](https://vuejs.org/guide/extras/render-function.html#render-functions-jsx) in the column definitions to obviate the need to use the slightly opaque `h` function style for rendering.~~ (Update: Pull this out to solve an ESLint issue.)
3. In the process, I broke the way the link icon column helper worked, but concluded a better way to encapsulate that might be in a component, so I made a component for it instead.

@hminsky2002 See what you think. I don't promise it's all perfect, but I think I achieved what I wanted to here.

### DIFF
--- a/apps/admin-interface/src/views/BaseFieldsView.vue
+++ b/apps/admin-interface/src/views/BaseFieldsView.vue
@@ -1,4 +1,4 @@
-<script setup lang="ts">
+<script setup lang="tsx">
 import {
 	PanelComponent,
 	PanelBody,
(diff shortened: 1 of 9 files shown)


hminsky2002 (approved):
@chriszs this is a much, much better approach to invoking the typescript/component-tizable features of tanstack into the codebase, thank you for running free with this! I tested locally and everything works except for the linting config, and boy oh BOY is it more readable. I really did not like using the hyperprocessor hook in vue also, I am so pleased to not be using it anymore.

PR #1318, opened 2026-02-03
Improving column resizing

### LINKED ISSUE #1309: Improve experience of column resizing
**Notes on what could be improved after qa testing column resizing:**
- Columns resize very slowly. This was the most noticeable thing by far. It took a lot of effort to move a border and resize a column, and the selected border would move at a slower speed than the mouse.
- Just by looking at the table, it wasn’t immediately obvious that there was a way to resize the columns or how to do it, unless your mouse stumbled onto the borders and turned them blue.
- Columns don’t stay resized. If you leave the main list page to edit a field, to make a new field, or to go to the home page, they revert to original size.
- The check mark column is a lot wider than I would have expected, maybe because the check marks are left-aligned instead of right-aligned.

hminsky2002:
This PR iterates on our existing column resizing and responds to feedback in #1309 to make the column resizing more natural and responsive! Specifically, it allows for the entire data table to extend past the default screen size, instead of conforming all resizing to still fit within the data table max width. 

Closes #1309

### DIFF
--- a/packages/components/src/components/Table/DataTable.vue
+++ b/packages/components/src/components/Table/DataTable.vue
@@ -7,7 +7,6 @@ import {
 	type ColumnDef,
 	type SortingState,
 	type ColumnFiltersState,
-	type ColumnResizeMode,
 	type ColumnSizingInfoState,
 	FlexRender,
 } from '@tanstack/vue-table';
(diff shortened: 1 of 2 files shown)


chriszs (commented):
This iteration looks okay to me in my local testing. I notice you abandoned the CSS vars approach, which may make sense, as I interpret that as mainly a workaround to avoid React re-renders. Unclear to me whether the inline calls you're doing to set style props directly is the most performant thing, but if it works, it works. Only flaw I can see is that resize dragging errantly selects table header text, and it feels a bit stranger to have the edit icon change position than the contents of some of the other columns. Feels like that should be more strongly anchored.

PR #1321, opened 2026-02-13
Add WordPress publishing workflow

yshng:
See https://github.com/PhilanthropyDataCommons/service/pull/2246

Addresses https://github.com/PhilanthropyDataCommons/meta/issues/137

Add workflow to publish .md files on push to WordPress site. Docs will be categorized based on file path so if we one day have more docs here, a directory structure will help that. We can also [add categories in front matter](https://github.com/OpenTechStrategies/wp-post-gfm/blob/da9252cf0c020868a9581a44db1ae8f7973637ed/src/index.js#L357), as below at the top of any Markdown file.

```
---
categories: cat1, cat2, data
---
```

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,0 +1,30 @@
+name: Publish to WordPress
+on:
+  push:
+    branches:
+      - main
+    paths:
+      - 'docs/**/*.md'
+  workflow_dispatch:
+    inputs:
+      force_publish:
+        description: 'Force republish all posts'
+        default: 'true'
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+        with:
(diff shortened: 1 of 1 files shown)


docs/SENSITIVITY_CLASSIFICATIONS.md
```diff
@@ -1,22 +1,44 @@
 # Data Sensitivity Classifications for Base Fields
 
-See below descriptions for the different sensitivity classifications for base fields in the pdc service.
+See below descriptions for the different sensitivity classifications
```
slifty: Ideally these whitespace changes wouldn't be in the same commit as the WP workflow -- Is the format change required for this workflow to work / intended?  If not, maybe we could even remove it from this PR entirely!

yshng: You're right, there is no reason for it. And now I know how to split commits with Magit!

PR #1369, opened 2026-04-15
More consistency in Bulk Upload instructional text

kfogel:
Various capitalization and punctuation consistency fixes. This is follow-up work for issue #1335.

PR #1370, opened 2026-04-15
1335 mention file path case sensitivity

kfogel:
This builds on commit f75f3e4a70b in PR #1369, so merge that first then this.


slifty (approved):
Approved -- though prettier needs to be run to fix the lint error in CI

hminsky2002 (approved):
Yep all looks reasonable, +1 to @slifty on rebasing with prettier. @kfogel that would just involve running `npm install` + `npm run lint` locally, or I could amend the commit here. Whatever's more convenient!

slifty:
correction you will need to run `npm run format` not `npm run lint` (lint checks, format changes)

PR #1374, opened 2026-04-16
Fix `format`

### LINKED ISSUE #1373: Fix the `format` command
Lint and format commands don't target the same files -- this means lint can pass but format would render lint errors.

I don't know which one has it right, but I do know we should have them both cover the same files!

slifty:
Our format eslint was targeting more files than our lint eslint; this fixes that and also expands the lint target to include storybook for both.

Resolves #1373


hminsky2002 (approved):
Much appreciated QoL improvement, thanks @slifty!

PR #1378, opened 2026-04-27
Add edit view and component

### LINKED ISSUE #1364: Add Edit View for Permissions Interface
Related to #1348 

This view will encompass editing existing permissions, via the `PUT` permissions routes on the service side.

hminsky2002:
This commit extends the PermissionGrantForm component to be used for the permission edit view. The main functional piece it introduces is the row to form state function, which is used for filling in the state of the permission at edit view load time.

Closes #1364

### DIFF
--- a/apps/user-tools/src/permissionsHelpers.ts
+++ b/apps/user-tools/src/permissionsHelpers.ts
@@ -157,6 +157,26 @@ export const getScopesForContextEntityType = (
 ): readonly string[] =>
 	isContextEntityType(value) ? CONTEXT_ENTITY_SCOPES[value] : [];
 
+export const rowToFormState = (
+	row: PermissionGrantRow,
+): PermissionGrantFormState => {
+	const entityKey = isContextEntityType(row.contextEntityType)
+		? row[CONTEXT_ENTITY_KEYS[row.contextEntityType]]
+		: undefined;
+	return {
+		granteeType: row.granteeType ?? null,
+		granteeId:
+			row.granteeUserKeycloakUserId ?? row.granteeKeycloakOrganizationId ?? '',
+		contextEntityType: row.contextEntityType,
+		entityKey:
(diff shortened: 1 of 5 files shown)


apps/user-tools/src/permissionsHelpers.ts
```diff
@@ -157,6 +157,26 @@ export const getScopesForContextEntityType = (
...
+		const { [CONTEXT_ENTITY_KEYS[row.contextEntityType]]: value } = row;
+		if (typeof value === 'string') entityKey = value;
+		else if (typeof value === 'number') entityKey = value.toString();
+	}
+	return {
+		granteeType: row.granteeType ?? null,
+		granteeId:
+			row.granteeUserKeycloakUserId ?? row.granteeKeycloakOrganizationId ?? '',
+		contextEntityType: row.contextEntityType,
+		entityKey,
```
chriszs: Another approach:
```suggestion
	const entityKey = isContextEntityType(row.contextEntityType)
		? row[CONTEXT_ENTITY_KEYS[row.contextEntityType]]
		: undefined;
	return {
		granteeType: row.granteeType ?? null,
		granteeId:
			row.granteeUserKeycloakUserId ?? row.granteeKeycloakOrganizationId ?? '',
		contextEntityType: row.contextEntityType,
		entityKey:
			typeof entityKey === 'string' || typeof entityKey === 'number'
				? String(entityKey)
				: '',
```

I started poking at this mainly because `String()` seemed simpler than the multiple if statements for type conversion and wound up with this, which is more succinct though reliant on some ternaries.

chriszs (approved):
Generally seems fine. One optional suggestion.

apps/user-tools/src/permissionsHelpers.ts
hminsky2002: I like it! Certainly more succinct.

PR #1388, opened 2026-05-05
Disable bulk upload creation when there are no application forms present

### LINKED ISSUE #1366: [Bug]: Bulk Upload - Empty Application Form Drop Down
### Component or URL

/add-data

### Steps to reproduce

1. Log into https://app.demo.philanthropydatacommons.org
2. Click Bulk Uploads
3. Click New bulk upload

### Expected behavior

There should be a list of Application Forms

### Actual behavior

Within the form, the first field 'Choose Application Form' shows an empty drop down when clicked:

Image

### Additional context & screenshots

I am a brand new user with a couple of accounts (admin, test, etc...).  It's possible I've used the wrong one...?  I'm not really sure, but wanted to call that out.

hminsky2002:
This commit explicitly disables the bulk-upload create view when no application forms are found. This was an issue that was brought to our attention by a user who had not been granted view permissions on any application forms, which effectively disabled the use of the view anyway. Now we document that case and add an info block when there are no forms present!

Closes #1366 [Bug]: Bulk Upload - Empty Application Form Drop Down

### DIFF
--- a/apps/user-tools/src/components/bulk-upload/BulkUploader.vue
+++ b/apps/user-tools/src/components/bulk-upload/BulkUploader.vue
@@ -22,7 +22,7 @@ import {
 } from '@pdc/components';
 import { getLogger } from '@pdc/utilities';
 import { ArrowDownTrayIcon } from '@heroicons/vue/24/outline';
-import { ref } from 'vue';
+import { computed, ref } from 'vue';
 import { downloadApplicationFormCsv } from '../../pdc-api';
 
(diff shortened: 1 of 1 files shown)


hminsky2002:
@slifty requesting your review explicitly on the wording of the info block here!

If we have any explicit docs on creating application forms it would probably be good to link them here too, but I don't believe we have any in the service, at least. Perhaps we could whip one up!

slifty (approved):

PR #1400, opened 2026-06-09
Add human readable labels for related permission entities

### LINKED ISSUE #1379: Use human-readable fields for displaying entity label in `/permissions`
[Our current rendering of the related entity (entity label) for a permission grant can be improved upon ](https://app.philanthropydatacommons.org/permissions)
Image

We should try to render the entity using it's human-readable label to the best of our abilities. This may involve some more work on the service side in how we return permissions!

hminsky2002:
This commit adds a front-end-only constructor for resolving human readable labels on related permissions entities. It does a pre-fetch of all related entities and then uses the resolver functions to get out the human readable label and render it. A lot of this code should be able to be cut down when we can get oneOf/anyOf types working in our sdk.

Closes #1379

### DIFF
--- a/apps/admin-interface/package.json
+++ b/apps/admin-interface/package.json
@@ -6,7 +6,7 @@
 		"@dsb-norge/vue-keycloak-js": "^3.0.7",
 		"@fontsource/source-sans-pro": "^5.2.5",
 		"@heroicons/vue": "^2.2.0",
-		"@pdc/sdk": "^0.36.1",
+		"@pdc/sdk": "^0.37.1",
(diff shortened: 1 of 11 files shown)


bickelj (approved):
In lieu of actual pagination, I would use maximum signed 32-bit integer to get the entities. It would also be nice to get the shallow versions of these entities instead of rich versions, but that only is possible with changemakers when sending unauthenticated GET requests (and IDK if that will last).

bickelj:
My review didn't really look at the code but I verified that when using the code it did what we want: human readable labels, also looked at the network tab of the browser tools to notice the one-backend-call-per-entity.

PR #1417, opened 2026-07-03
Block application form download on application form, instead of bulkupload

### LINKED ISSUE #1416: Bulk-upload: Re-enable application form download
Currently, the application form download feature is disabled as it is gated on the user providing a file, which is backwards! It should be gated on the relevant application form having loaded in. This should be a simple, one-line fix.

hminsky2002:
Closes #1416


bickelj (approved):
I confirmed in the demo environment that I could not get this link to work when selecting application forms in the "new bulk upload" interface and then in the preview environment the link does work. Nice!

PR #1422, opened 2026-07-28
Update pdc_changemaker_id info block

### LINKED ISSUE #1419: Update `pdc_changemaker_id` informational block
See corresponding thread in zulip: [#PDC > Bulk Uploads @ 💬](https://chat.opentechstrategies.com/#narrow/channel/75-PDC/topic/Bulk.20Uploads/near/265445)

The current `pdc_changemaker_id` informational block has not been updated to reflect the latest state of the service, namely that we require that column to be prepended by `control:`, as the first instance of a new paradigm where we have data and control columns. A simple fix for now is to just update the text block to include control.

hminsky2002:
Closes #1419


bickelj (approved):
LGTM (from the preview env):

image

I also see that the only change is this wording.

